PR #21, opened 2015-02-21
Reproduce JENKINS-27042

olivergondza:
[JENKINS-27042](https://issues.jenkins-ci.org/browse/JENKINS-27042)

### DIFF
--- a/src/test/java/plugins/CompressArtifactsPluginTest.java
+++ b/src/test/java/plugins/CompressArtifactsPluginTest.java
@@ -25,11 +25,13 @@ package plugins;
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.greaterThanOrEqualTo;
 import static org.hamcrest.Matchers.not;
 import static org.hamcrest.collection.IsCollectionWithSize.hasSize;
 
 import org.jenkinsci.test.acceptance.Matcher;
 import org.jenkinsci.test.acceptance.junit.AbstractJUnitTest;
+import org.jenkinsci.test.acceptance.junit.Bug;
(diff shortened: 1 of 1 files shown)


src/test/java/plugins/CompressArtifactsPluginTest.java
```diff
@@ -75,6 +77,33 @@ public void access_uncompressed_artifact_after_plugin_was_installed() {
...
+    @Test @Bug("JENKINS-27042")
+    @WithPlugins("compress-artifacts")
+    public void archiveLargerThan4GInTotal() throws Exception {
+        configureArtifactCompression();
+
+        FreeStyleJob job = jenkins.jobs.create();
+        job.configure();
+        job.addPublisher(ArtifactArchiver.class).includes("*");
+        job.addBuildStep(ShellBuildStep.class).command( // Generate archive larger than 4G
+                "wget $JENKINS_URL/jnlpJars/jenkins-cli.jar -O stuff.jar; for i in {0..7000}; do cp -l stuff.jar stuff.${i}.jar; done"
```
rsandell: Instead of relying on wget being installed you could do something like the other archive tests: https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/test/java/core/ArtifactsTest.java#L45

olivergondza:
@rsandell Unfortunately, content of /dev/zero have very good compression ratio for generating archive larger than 4G. I was thinking about /dev/random, though the test might be flaky because of the unpredictable ratio. That is why I use a file that was zipped already. But, I guess I can copy slave.jar from slaves remote FS.

olivergondza:
None of the usual jars seems to be on predictable FS location in case the build is run on slave vs on master. We rely on wget/curl on several places, if this really causes some problems we can come up with reusable `curl ... || wget ...` snippet to be used in shell scripts.

PR #24, opened 2015-05-06
[JENKINS-27669] Demonstrate ClassCircularityError before fix

jglick:
Requires https://github.com/jenkinsci/support-core-plugin/pull/31 to be merged and released before this test can be expected to pass (without `LOCAL_SNAPSHOTS=true`).

@reviewbybees

### DIFF
--- a/src/test/java/plugins/SupportCorePluginTest.java
+++ b/src/test/java/plugins/SupportCorePluginTest.java
@@ -1,0 +1,45 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2015 Jesse Glick.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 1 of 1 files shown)


src/test/java/plugins/SupportCorePluginTest.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Collections;
+import java.util.logging.Level;
+import org.jenkinsci.test.acceptance.junit.AbstractJUnitTest;
+import org.jenkinsci.test.acceptance.junit.WithPlugins;
+import org.jenkinsci.test.acceptance.po.JenkinsLogger;
+import static org.junit.Assert.*;
+import org.junit.Test;
+import org.jvnet.hudson.test.Issue;
+
+@WithPlugins("support-core@2.22")
```
olivergondza: This should be `@WithPlugins("support-core")` IMO. As I understand it, it says it is pointless to test on any release older than 2.22, but it is not the case. The bug was there and the test should fail when run with any older version.

It would be OK if the test cover a feature introduced in 2.22.

rsandell:
:+1:

src/test/java/plugins/SupportCorePluginTest.java
jglick: My intention was to have the test defend against future regressions (which would be very easy to introduce, given the fragility of the fix), so testing against older plugin releases without the fix was not even a consideration, but it is no problem to allow that (at least now that 2.22 has been released and so should be present on the update center).

PR #26, opened 2015-06-10
Xvnc tests

jglick:
- [X] switched to a Docker fixture for better control and reproducibility
- [X] tested Workflow integration from https://github.com/jenkinsci/xvnc-plugin/pull/5

@reviewbybees and @olivergondza

### DIFF
--- a/src/test/java/plugins/XvncPluginTest.java
+++ b/src/test/java/plugins/XvncPluginTest.java
@@ -1,7 +1,7 @@
 package plugins;
 
+import com.google.inject.Inject;
 import org.jenkinsci.test.acceptance.junit.AbstractJUnitTest;
-import org.jenkinsci.test.acceptance.junit.Native;
 import org.jenkinsci.test.acceptance.junit.WithPlugins;
 import org.jenkinsci.test.acceptance.plugins.xvnc.XvncGlobalJobConfig;
 import org.jenkinsci.test.acceptance.plugins.xvnc.XvncJobConfig;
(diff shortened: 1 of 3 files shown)


src/test/java/plugins/XvncPluginTest.java
```diff
@@ -11,21 +11,36 @@
...
+import static org.hamcrest.Matchers.containsString;
+import org.jenkinsci.test.acceptance.docker.DockerContainerHolder;
+import org.jenkinsci.test.acceptance.docker.fixtures.XvncSlaveContainer;
 import static org.jenkinsci.test.acceptance.plugins.xvnc.XvncJobConfig.*;
+import org.jenkinsci.test.acceptance.po.Slave;
+import org.jenkinsci.test.acceptance.po.WorkflowJob;
+import org.jvnet.hudson.test.Issue;
 
 @WithPlugins("xvnc")
 public class XvncPluginTest extends AbstractJUnitTest {
```
olivergondza: This should have `@WithDocker` so it will be skipped if docker is not available.

olivergondza:
:+1: on improving the test. Please make sure to put the test case back to https://wiki.jenkins-ci.org/display/JENKINS/LTS+RC+Testing until we can run it on jenkins.ci.cloudbees.com.

jglick:
> Please make sure to put the test case back to https://wiki.jenkins-ci.org/display/JENKINS/LTS+RC+Testing until we can run it on jenkins.ci.cloudbees.com.

Sorry, I did not follow what you are saying here.

olivergondza:
@jglick, As more and more test uses docker and those tests are not run on any CI for now, I need to run them locally during the LTS RC period. Noting so we will not forget:  https://wiki.jenkins-ci.org/pages/diffpages.action?pageId=79921289&originalId=80183775

jglick:
> As more and more test uses docker and those tests are not run on any CI for now

Ah. We should just get a Docker-capable slave set up for this purpose I guess. Another thing I want to do is switch the acceptance test jobs to be Workflows so that they can survive server restarts, which should be possible once jenkins.ci is on 1.609.x, which should be soon, because…then we can use the `Xvnc` wrapper!

(CC @recampbell)

PR #27, opened 2015-07-05
[JENKINS-28550] Test reviewed to work fine with the new proposal of side-panel and main-panel

recena:
This [PR](https://github.com/jenkinsci/jenkins/pull/1717) have modified DOM tree in the main layout. These changes affect to `side-panel` and `main-panel`. From now on, `div#main-panel-content` won't exist.


daniel-beck:
@olivergondza Any possible problems with this change merged the same time as https://github.com/jenkinsci/jenkins/pull/1717, e.g. Jenkins LTS ATH runs?

olivergondza:
If I am not mistaken, it will work with both new and old version after this fix. In that case, :+1: .

recena:
@olivergondza Thanks so much for your quick feedback

daniel-beck:
@olivergondza Great, thanks!

PR #29, opened 2015-08-14
[JENKINS-25977] Reflect analysis-core configuration page modifications

amuniz:
Downstream of https://github.com/jenkinsci/analysis-core-plugin/pull/31

The configuration page for any Analysis Core derived plugin has now a different HTML selector for everything under the old section ```canComputeNew```. This intermediate selector level has been removed as ```inline="true"``` [has been added in the jelly file](https://github.com/jenkinsci/analysis-core-plugin/pull/31/files#r35576335).

@reviewbybees


jglick:
:bug: just use for example

```java
control("failedNewAll", "canComputeNew/failedNewAll")
```

which ought to work before or after the change. (Try it both ways, of course!)

amuniz:
After using the two arguments ```control``` as pointed by @jglick this PR could be merged without conflicts.

jglick:
:bee:

PR #30, opened 2015-08-18
[JENKINS-29976] Create an attribute to set Dockerfile in @DockerFixture 

lordofthejars:
Create an attribute to set Dockerfile in @DockerFixture. This PR also fixes one small issue [JENKINS-29984](https://issues.jenkins-ci.org/browse/JENKINS-29984) but since it was found while writing a test for [JENKINS-29976](https://issues.jenkins-ci.org/browse/JENKINS-29976) it was difficult to split at that stage to two different commits because the bug was at the same class that I touched for implementing this feature. Since it is an small bug fix (support for inner classes for DockerFixture), I decided to not overcomplicate the development by removing/copyback the code.

I paste here the description of the issue:

```
Create attribute to set Dockerfile location in @DockerFixture. By default Dockerfile fixture should be placed in the resource directory and at the same package as DockerContainer subtype.

A new optional attribute called dockerfile can be added at @DockerFixture, so if it is not present then default behaviour is the expected, but if the attribute is present, Dockerfile location is get from there. This allows save Dockerfile in places where they can be reused.

An example of the annotation might be:

@DockerFixture(id="jira", ports="2909", dockerfile="a/b")
public class JiraContainer ....

Classpath resolution will be based on class where annotation is defined.
```

@reviewbybees

### DIFF
--- a/docs/FIXTURES.md
+++ b/docs/FIXTURES.md
@@ -75,12 +75,8 @@ publicclass TestContainer extends DockerContainer {
 }
 ```
 
-In this case the docker file must be located at `org/ame/fixture/test` instead of the default location.
+In this case the docker file must be located at classpath and in `org/ame/fixture/test` instead of the default location.
 
-dockerfile attribute can contain three kind of values:
-* File based: file:///tmp/fixture/test which resolves Dockerfile from file location.
-* Classpath based: `classpath://org/acme/fixture/test which resolves Dockerfile from classpath location.
-* No schema provided which by default uses the classpath ones.
 
 
 
--- a/docs/FIXTURES.md
+++ b/docs/FIXTURES.md
@@ -63,4 +63,24 @@ The `@DockerFixture` bindIp allows you to bind the accessible ports of the docke
 A fixture also needs to define `Dockerfile` in the resources directory. If a fixture class is
 `org/acme/FooContainer.java`, then the docker file must be located at `org/acme/FooContainer/Dockerfile`.
 
+In case of inner classes, dollar sign ( $ ) is replaced with a slash ( / ).
+
+You can also set custom location of `Dockerfile` by using `dockerfile` attribute from`@DockerFixture` annotation.
+A simple example might be:
+
+```java
+@DockerFixture(id="test", ports="8080", dockerfile="org/acme/fixture/test")
+publicclass TestContainer extends DockerContainer {
+//...
+}
+```
+
+In this case the docker file must be located at `org/ame/fixture/test` instead of the default location.
+
+dockerfile attribute can contain three kind of values:
+* File based: file:///tmp/fixture/test which resolves Dockerfile from file location.
+* Classpath based: `classpath://org/acme/fixture/test which resolves Dockerfile from classpath location.
+* No schema provided which by default uses the classpath ones.
+
+
 
--- a/docs/FIXTURES.md
+++ b/docs/FIXTURES.md
@@ -65,11 +65,11 @@ A fixture also needs to define `Dockerfile` in the resources directory. If a fix
 
 In case of inner classes, dollar sign ( $ ) is replaced with a slash ( / ).
 
-You can also set custom location of `Dockerfile` by using `dockerfile` attribute from`@DockerFixture` annotation.
+You can also set custom location of `Dockerfile` by using `dockerfileFolder` attribute from`@DockerFixture` annotation.
 A simple example might be:
 
 ```java
-@DockerFixture(id="test", ports="8080", dockerfile="org/acme/fixture/test")
+@DockerFixture(id="test", ports="8080", dockerfileFolder="org/acme/fixture/test")
 publicclass TestContainer extends DockerContainer {
 //...
 }
--- a/docs/FIXTURES.md
+++ b/docs/FIXTURES.md
@@ -70,7 +70,7 @@ A simple example might be:
 
 ```java
 @DockerFixture(id="test", ports="8080", dockerfileFolder="org/acme/fixture/test")
-publicclass TestContainer extends DockerContainer {
+public class TestContainer extends DockerContainer {
 //...
 }
 ```
--- a/src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
@@ -27,9 +27,13 @@ import java.util.jar.JarFile;
  * Use this subsystem by injecting this class into your test.
  *
  * @author Kohsuke Kawaguchi
+ * @author asotobueno
  */
 @Singleton
 public class Docker {
+
+    private static final String FILE_SCHEMA = "file://";
+    private static final String CLASSPATH_SCHEMA = "classpath://";
 
     /**
      * Command to invoke docker.
@@ -111,44 +115,7 @@ public class Docker {
             dir.delete();
             dir.mkdirs();
             try {
-                File jar = null;
-                try {
-                    jar = Which.jarFile(fixture);
-                } catch (IllegalArgumentException e) {
-                    // fall through
-                }
-
-                if (jar!=null && jar.isFile()) {
-                    // files are packaged into a war. extract them
-                    String prefix = fixture.getName().replace('.', '/') + "/";
-                    try (JarFile j = new JarFile(jar)) {
-                        Enumeration<JarEntry> e = j.entries();
-                        while (e.hasMoreElements()) {
-                            JarEntry je = e.nextElement();
-                            if (je.getName().startsWith(prefix)) {
-                                File dst = new File(dir, je.getName().substring(prefix.length()));
-                                if (je.isDirectory()) {
-                                    dst.mkdirs();
-                                } else {
-                                    try (InputStream in = j.getInputStream(je)) {
-                                        FileUtils.copyInputStreamToFile(in, dst);
-                                    }
-                                }
-                            }
-                        }
-                    }
-                } else {
-                    // Dockerfile is not packaged into a jar file, so copy locally
-                    URL resourceDir = classLoader.getResource(fixture.getName().replace('.', '/'));
-                    File dockerFileDir;
-                    try {
-                        dockerFileDir = new File(resourceDir.toURI());
-                    } catch (URISyntaxException e) {
-                        dockerFileDir = new File(resourceDir.getPath());
-                    }
-                    FileUtils.copyDirectory(dockerFileDir, dir);
-                }
-
+                copyDockerfileDirectory(fixture, f, dir);
                 return build("jenkins/" + f.id(), dir);
             } finally {
                 FileUtils.deleteDirectory(dir);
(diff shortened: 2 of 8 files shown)


src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
```diff
@@ -158,6 +125,107 @@ public DockerImage build(Class<? extends DockerContainer> fixture) throws IOExce
         }
     }
 
+    //package scope for testing purposes. Ideally we should encapsulate Docker interactions so they can be mocked
+    // and call public method.
+    void copyDockerfileDirectory(Class<? extends DockerContainer> fixture, DockerFixture f, File dir)
+            throws IOException {
+        String dockerfilePath = resolveDockerfileLocation(fixture, f);
+
+        if(dockerfilePath.startsWith(FILE_SCHEMA)) {
```
rsandell: ```FileUtils.copyURLToFile(URL source, File destination)``` could be simpler to use, and you would get support for http e.al. at the same time.

lordofthejars: I thought the same but then I open the door that someone could use any schema like `http` or `ftp` since Java provides out of the box this support. So I decided to put some protection on this.
But of course I could implement the `if`as now but then call `copyURLToFile` as you mention.

jglick:
I see no purpose for the file schema. Just get rid of the protocol and treat the value as a resource name. Simplifies the code and nothing of value is lost.

lordofthejars:
I was thinking on more global purpose where users are not limited only in
classpath location and also opening to other protocols, but let s simplify
the code and if it is required in future, we can add it.
El dt., 18 d’ag., 2015 a les 18.22 Jesse Glick <notifications@github.com>
va escriure:

src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
```diff
@@ -158,6 +122,80 @@ public DockerImage build(Class<? extends DockerContainer> fixture) throws IOExce
...
+        }
+        return prefix;
+    }
+
+    private void copyDockerfileDirectoryFromClasspath(Class<? extends DockerContainer> fixture, String dockerfileLocation, File dir) throws IOException {
+        File jar = null;
+        try {
+            jar = Which.jarFile(fixture);
+        } catch (IllegalArgumentException e) {
+            // fall through
```
oleg-nenashev: Some logging would be useful

src/test/java/org/jenkinsci/test/acceptance/docker/DockerTest.java
```diff
@@ -0,0 +1,138 @@
...
+import java.lang.reflect.Field;
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.util.Map;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.Is.is;
+
+/**
+ * @author asotobueno
```
oleg-nenashev: NIT: If you add the Javadoc header, it's useful to add some info. E.g. <code>Tests for {@link DockerFixture}</code>. Probably the class could be also renamed, because actually we test the fixtures.

oleg-nenashev:
Looks good to me in general, only minor comments

src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
lordofthejars: yes I completely agree but this was originally there, and since it is an improvement I didn't want to touch this as side effect. I have seen this structure in more places so we would need to create an issue for fixing some of them. Also in this case is not bad at all (maybe we could log as info) since if/else that are below this line takes care of everything. But as I said I agree with you that we need to log these lines. 

Do you think it has sense to open another issue with these improvements?

src/test/java/org/jenkinsci/test/acceptance/docker/DockerTest.java
lordofthejars: Absolutely.

src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
lordofthejars: What logger framework you usually use? I tend to use JUL.

oleg-nenashev: I don't think it's really required for the internal test framework. Since you're decoupling the code here, you can just add such messaging as a "refactoring". I would just keep it as a separate commit just to highlight the change for developers if they are interested in.

lordofthejars: Yes in separate commit/PR make sense to not mix things. To log it you mean using Sysout directly?

oleg-nenashev: probably a default java.logging.Logger would be the best option, but I'm not sure which log listeners are attached at this point.

lordofthejars: yes I tend to use java.logging.Logger (JUL) in this situations to not add a new framework

olivergondza:
Is there any use-case for reusing dockerfile where you can not result whole fixture?

lordofthejars:
Not now since it copies the whole directory. For what I understand you mean copying only Dockerfile and not the directory and the contents where it is stored isn't it? If it is the case, it is not implemented but we could open a new issue to do it.

olivergondza:
I guess I do not understand what are you trying to achieve. My point was, why sharing dockerfiles between fixtures if you can reuse those fixtures for multiple tests?

lordofthejars:
Because by default a Dockerfile must be in the same package and class name of the DockerFixture. What I am trying to achieve is breaking this hard dependency and allow users to put fixtures where they think it is better for the test. For example I have two test classes (not test methods) that wants to share same Dockerfile, then you can do it. But what it is really important this change is because it is a prior step to https://issues.jenkins-ci.org/browse/JENKINS-30009 and be able to make DockerFixture classes self-contained.

olivergondza:
Well, I think of Dockerfile as an implementation detail of its fixture (which seems to be useful for JENKINS-30009). This change exposes Dockerfiles for reuse and encourages other fixtures to depend on your dockerfile content which seem to be the exact opposite of what you want.

Why can you not create *one* fixture and use it in *both* your tests?

lordofthejars:
For what I see, in previous implementation you, before this PR, directory is automatically resolved. So if you have `public class ATest` and `public class Btest` then the directories where the fixtures are loaded are `<package>/ATest` and `<package>/BTest` so it is impossible to reuse them.

olivergondza:
As I said `SshdContainer` is shared between `PublishOverSSHPluginTest` and `ScpPluginTest` so there is a way to reuse. In what regard it does not work for you?

As we failed to understand each other until this point, do you mind implementing your fixtures with duplicity and refactor the fixture sharing after that to demonstrate what the problem actually is?

lordofthejars:
Sure you can do it in this way, and you can still do it in this way, but what I am implementing here is a way to allow developer to add fixtures in different directory and not depending only in defaults. The old behaviour is still valid and working so I simply enrich a bit more the project with this feature which adds more freedom to developers. It is not a problem itself with the current implementation it is simply an enhancement to the project.

olivergondza:
I see. I am -0 on adding more ways to do the same thing.

lordofthejars:
It is not the same thing, for example if I want to have all fixtures in src/test/resources/fixtures, then it is impossible. This is something that all test frameworks that works with fixtures offers. I don't see any problem.

src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
```diff
@@ -158,6 +122,80 @@ public DockerImage build(Class<? extends DockerContainer> fixture) throws IOExce
         }
     }
 
+    //package scope for testing purposes. Ideally we should encapsulate Docker interactions so they can be mocked
+    // and call public method.
+    void copyDockerfileDirectory(Class<? extends DockerContainer> fixture, DockerFixture f, File dir)
```
tfennelly: Seems like `@Nonnull` etc might be appropriate?

src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
```diff
@@ -158,6 +122,80 @@ public DockerImage build(Class<? extends DockerContainer> fixture) throws IOExce
...
+    }
+
+    private String resolveDockerfileLocation(Class<? extends DockerContainer> fixture, DockerFixture f) {
+        String prefix = null;
+        if(isSpecificDockerfileLocationSet(f)) {
+            prefix = f.dockerfile().replace('.', '/').replace('$', '/');
+        } else {
+            prefix = fixture.getName().replace('.', '/').replace('$', '/');
+        }
+        return prefix;
```
tfennelly: Might be clearer to do the replace in one place?

E.g.

```java
String prefix = null;
if(isSpecificDockerfileLocationSet(f)) {
    prefix = f.dockerfile();
} else {
    prefix = fixture.getName();
}
return prefix.replace('.', '/').replace('$', '/');
```

src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
```diff
@@ -158,6 +122,80 @@ public DockerImage build(Class<? extends DockerContainer> fixture) throws IOExce
...
+        return f.dockerfile() != null && !"".equals(f.dockerfile().trim());
+    }
+
+    private void copyDockerfileDirectoryFromPackaged(File jar, String fixtureLocation, File outputDirectory) throws IOException {
+        try (JarFile j = new JarFile(jar)) {
+            Enumeration<JarEntry> e = j.entries();
+            while (e.hasMoreElements()) {
+                JarEntry je = e.nextElement();
+                if (je.getName().startsWith(fixtureLocation)) {
+                    File dst = new File(outputDirectory, je.getName().substring(fixtureLocation.length()));
```
tfennelly: possible :bug:  ... wouldn't you need to check the length here before doing `substring` ?

tfennelly: Sorry ... ignore ... I see you did a `startsWith` check before it.

tfennelly:
:bee:

src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
lordofthejars: True, I am going to refactor :)

src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
lordofthejars: Well I am not aware of the politics you follow regarding this, in package scope you also use them?

docs/FIXTURES.md
```diff
@@ -63,4 +63,20 @@ The `@DockerFixture` bindIp allows you to bind the accessible ports of the docke
...
 `org/acme/FooContainer.java`, then the docker file must be located at `org/acme/FooContainer/Dockerfile`.
 
+In case of inner classes, dollar sign ( $ ) is replaced with a slash ( / ).
+
+You can also set custom location of `Dockerfile` by using `dockerfile` attribute from`@DockerFixture` annotation.
+A simple example might be:
+
+```java
+@DockerFixture(id="test", ports="8080", dockerfile="org/acme/fixture/test")
+publicclass TestContainer extends DockerContainer {
```
jglick: missing space

src/main/java/org/jenkinsci/test/acceptance/docker/DockerFixture.java
```diff
@@ -38,4 +39,23 @@
...
+     *  By default Dockerfile fixture should be placed in the resource directory
+     *  and at the same package as DockerContainer subtype.
+     * </p>
+     * <p>
+     *  If this attribute is present, Dockerfile file specified in the attribute
+     *  is used as Dockerfile fixture.
+     * </p>
+     * <p>
+     *  The location can be expressed in terms of file location (file:///a/b/Dockerfile)
+     *  or classpath location (classpath://a/b/Dockerfile).
```
jglick: I thought you removed this feature?

src/main/java/org/jenkinsci/test/acceptance/docker/DockerFixture.java
```diff
@@ -38,4 +39,23 @@
...
+    /**
+     * Path of Dockerfile file.
+     *
+     * <p>
+     *  By default Dockerfile fixture should be placed in the resource directory
+     *  and at the same package as DockerContainer subtype.
+     * </p>
+     * <p>
+     *  If this attribute is present, Dockerfile file specified in the attribute
+     *  is used as Dockerfile fixture.
```
jglick: According to the documentation and the tests, this attribute specifies the (resource) path to the _folder_ containing `Dockerfile`, not the `Dockerfile` itself. So this Javadoc is misleading, as is the name of the attribute.

src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
jglick: Generally we like to add nullness annotations everywhere we can.

src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
```diff
@@ -158,6 +122,80 @@ public DockerImage build(Class<? extends DockerContainer> fixture) throws IOExce
...
+            dockerfileLocation += "/";
+            copyDockerfileDirectoryFromPackaged(jar, dockerfileLocation, dir);
+        } else {
+            // Dockerfile is not packaged into a jar file, so copy locally
+            copyDockerfileDirectoryFromLocal(dockerfileLocation, dir);
+        }
+    }
+
+    private boolean isSpecificDockerfileLocationSet(DockerFixture f) {
+        return f.dockerfile() != null && !"".equals(f.dockerfile().trim());
```
jglick: Annotation values cannot be null,

```
@DockerFixture(…, dockerfile=" ")
```

seems like an unlikely user mistake, and comparing to `""` is deprecated. This method could just be inlined as

```java
!f.dockerfile().isEmpty()
```

jglick:
The added tests are good. As with @olivergondza I see no real motivation for the change and minor added API complexity—the conventional location worked fine—but it seems harmless enough.

src/main/java/org/jenkinsci/test/acceptance/docker/DockerFixture.java
lordofthejars: Ups I updated doc but not javadoc.

src/main/java/org/jenkinsci/test/acceptance/docker/DockerFixture.java
lordofthejars: Maybe `dockerfileFolder` may be better? I am don't know exactly how to name it

src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
lordofthejars: Yes I know but I have find in  my career doing such "awesome" things using reflection. But well let's assume that in this case since it is for testing we won't get this things.

lordofthejars:
Ok, then the maintainer can accept or close the PR, but this conversion goes to nowhere. As I don't see any complication on the API, in fact I have refactored a lot of things that originally were not refactored and makes difficult to follow, and simply I have added new features to the API.

src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
lordofthejars: Ok I have seen that in this project it is using java8 annotations for non nullable.

src/main/java/org/jenkinsci/test/acceptance/docker/DockerFixture.java
lordofthejars: Since I agree with you that `dockerfile` doesn't describe very well the purpose I have changed to `dockerFolder`

src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
```diff
@@ -27,6 +27,7 @@
  * Use this subsystem by injecting this class into your test.
  *
  * @author Kohsuke Kawaguchi
+ * @author asotobueno
```
jglick: BTW you can also feel free (IMO) to just delete any `@author` tags. Anyone who cares can `git blame` for themselves.

jglick:
:bee:

PR #31, opened 2015-08-18
Use workdir

KostyaSha:
not verified locally (atm)

### DIFF
--- a/src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/XvncSlaveContainer/Dockerfile
+++ b/src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/XvncSlaveContainer/Dockerfile
@@ -8,7 +8,8 @@ RUN mkdir /tmp/.X11-unix && chmod 1777 /tmp/.X11-unix/
 ENV XAUTHORITY /home/jenkins/.Xauthority
 
 USER jenkins
-RUN mkdir /home/jenkins/.vnc && (echo jenkins; echo jenkins) | vncpasswd /home/jenkins/.vnc/passwd
+WORKDIR /home/jenkins/
+RUN mkdir .vnc && (echo jenkins; echo jenkins) | vncpasswd .vnc/passwd
 # Default content includes x-window-manager, which is not installed, plus other stuff we do not need (vncconfig, x-terminal-emulator, etc.):
-RUN touch /home/jenkins/.vnc/xstartup && chmod a+x /home/jenkins/.vnc/xstartup
+RUN touch .vnc/xstartup && chmod a+x .vnc/xstartup
 USER root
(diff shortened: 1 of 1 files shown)


recena:
@KostyaSha Looks fine. Did not you try it?

lordofthejars:
Looks it is going to work without much problems. @KostyaSha can you open a JIRA issue so we can trace all the changes that goes in the release?

Thank you very much for your improvement and collaboration.

KostyaSha:
What release??

lordofthejars:
When we make the release of acceptance-test-harness project we would like to add a reference of things that has been changed. For this reason we would like to have everything managed in public JIRA. There is a component called acceptance-test-harness for this purpose.

KostyaSha:
Jenkins project is not managing releases in jira. I see no reasons for having releases for storage of tests.

lordofthejars:
You can see an example here just in case you want to open an issue and referencing here: https://issues.jenkins-ci.org/browse/JENKINS-29976

jglick:
Seems like a straightforward internal code cleanup with no effect on tests. I see no reason to bother with a JIRA issue for it. Also most of the work here is done by @olivergondza and he generally does not use JIRA for it, so if you wanted to create a changelog you would need to scan the commit list anyway (or get agreement from him to enforce consistent use of JIRA going forward).

CloudBees does consume `acceptance-test-harness` releases, and perhaps others do as well. I have never felt a strong need for an upstream changelog as such; things get deprecated and sometimes we clean up those usages. Of course when an incompatible change is made, that is painful, but I would want the change to have been done compatibly to begin with, if at all possible. If we do start up a changelog, keep it focused on changes to framework APIs and usage patterns.

olivergondza:
@jglick, it seems that we agree that tracking changes in JIRA would be unnecessary for ATH, as there is handful of clients and most if not all of them are contributors.

lordofthejars:
Ok then in my opinion we are treating test (frameworks) as second-class citizen. I think both should be developed in the same way, but well if the idea is to not open JIRA issues, we can avoid it for sure.

KostyaSha:
ATH was never a jenkins production project in comparison to CB that enforces their internal rules to Jenkins project?

KostyaSha:
@recena tried with current state as part of centos based image, workdir and user must be "unset" either they will become as default workdir/user for final image. So it should work without any errors.

https://docs.docker.com/articles/dockerfile_best-practices/#workdir

lordofthejars:
@KostyaSha what I am saying is that for what I understand Jenkins open source project uses its open JIRA register issues, IMHO we should follow the same strategy for the test framework.

KostyaSha:
@lordofthejars Jenkins has no resources for managing everything in JIRA and not all projects managing them, even more, part of people (and me) managing issues in GH, because it allows save time and work comfortably instead of bad-maintained, not-updated JIRA in Jenkins project.

PR #34, opened 2015-08-20
[JENKINS-26942] Update acceptance test to use stashes

jglick:
On hold until https://github.com/jenkinsci/workflow-plugin/pull/186 is merged and released.

@reviewbybees


olivergondza:
Btw, it is possible to require updated version of plugin only for the one test that really needs it (and keep the rest testable with older version).

oleg-nenashev:
Cannot be merge till the workflow-aggregator release. I would also prefer per-method declarations, because such implementation increases the test coverage.

jglick:
Looked at `WithPlugins.RuleImpl.apply` and it was unclear whether a plugin version in a method annotation would properly *override* that in a class annotation, so am instead trying to just replace the class annotation.

oleg-nenashev:
:bee:

stephenc:
:bee: from the :sunglasses: on :horse_racing:

PR #39, opened 2015-09-21
adds screen recorder capabilities.

lordofthejars:
@reviewbybees

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ There is a bit of a delay since we bring up Jenkins for every single test, with
 it's own sandboxed workspace.
 
 All executed tests are screen recorded by default, but only videos of failing tests are persited to `target` directory.
-By default, vide file is named with the fully qualified test class name, minus sign (-) and the test method name.
+By default, video file is named with the fully qualified test class name, minus sign (-) and the test method name.
 
 If you want to persist all videos, the ones that succeeded too, you can set `RECORDER_SAVE_ALL` Java system property to true.
 
--- a/README.md
+++ b/README.md
@@ -21,6 +21,11 @@ Set `BROWSER=chrome` and install http://code.google.com/p/chromedriver/downloads
 There is a bit of a delay since we bring up Jenkins for every single test, with
 it's own sandboxed workspace.
 
+All executed tests are screen recorded by default, but only videos of failing tests are persisted to `target` directory.
+By default, vide file is named with the fully qualified test class name, minus sign (-) and the test method name.
+
+If you want to persist all videos, the ones that succeeded too, you can set `RECORDER_SAVE_ALL` Java system property to true.
+
 ## Further Reading
 
 ### Running tests
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Possible values are:
 * failuresOnly
 * always
 
-Environment variable takes precedence over Java system property.
+Java system property takes precedence over environment variable.
 
 ## Further Reading
 
--- a/README.md
+++ b/README.md
@@ -21,10 +21,12 @@ Set `BROWSER=chrome` and install http://code.google.com/p/chromedriver/downloads
 There is a bit of a delay since we bring up Jenkins for every single test, with
 it's own sandboxed workspace.
 
-All executed tests are screen recorded by default, but only videos of failing tests are persisted to `target` directory.
+All executed tests are screen recorded by default, but only videos of failing tests are persited to `target` directory.
 By default, vide file is named with the fully qualified test class name, minus sign (-) and the test method name.
 
 If you want to persist all videos, the ones that succeeded too, you can set `RECORDER_SAVE_ALL` Java system property to true.
+
+If you want to completely disable recording, you can set `RECORDER_DISABLED` Java system property to true.
 
 ## Further Reading
 
--- a/README.md
+++ b/README.md
@@ -24,9 +24,14 @@ it's own sandboxed workspace.
 All executed tests are screen recorded by default, but only videos of failing tests are persited to `target` directory.
 By default, video file is named with the fully qualified test class name, minus sign (-) and the test method name.
 
-If you want to persist all videos, the ones that succeeded too, you can set `RECORDER_SAVE_ALL` Java system property to true.
+You can configure what is persisted by using an environment variable or a Java system property called `RECORDER`.
+Possible values are:
 
-If you want to completely disable recording, you can set `RECORDER_DISABLED` Java system property to true.
+* off
+* failuresOnly
+* always
+
+Environment variable takes precedence over Java system property.
 
 ## Further Reading
 
--- a/src/main/java/org/jenkinsci/test/acceptance/recorder/JUnitScreenRecorder.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/recorder/JUnitScreenRecorder.java
@@ -1,40 +1,19 @@
-/*
- * The MIT License
- *
- * Copyright (c) 2014 Red Hat, Inc.
- *
- * Permission is hereby granted, free of charge, to any person obtaining a copy
- * of this software and associated documentation files (the "Software"), to deal
- * in the Software without restriction, including without limitation the rights
- * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
- * copies of the Software, and to permit persons to whom the Software is
- * furnished to do so, subject to the following conditions:
- *
- * The above copyright notice and this permission notice shall be included in
- * all copies or substantial portions of the Software.
- *
- * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
- * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
- * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
- * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
- * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
- * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
- * THE SOFTWARE.
- */
 package org.jenkinsci.test.acceptance.recorder;
 
 import org.junit.runner.Description;
 import org.monte.media.Format;
 import org.monte.media.Registry;
 import org.monte.screenrecorder.ScreenRecorder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.awt.*;
 import java.io.File;
 import java.io.IOException;
-import java.text.SimpleDateFormat;
-import java.util.Date;
 
 public class JUnitScreenRecorder extends ScreenRecorder {
+
+    private static final Logger logger = LoggerFactory.getLogger(JUnitScreenRecorder.class);
 
     private Description description;
     private Format format;
@@ -49,9 +28,13 @@ public class JUnitScreenRecorder extends ScreenRecorder {
 
     protected File createMovieFile(Format fileFormat) throws IOException {
         if(!this.movieFolder.exists()) {
-            this.movieFolder.mkdirs();
-        } else if(!this.movieFolder.isDirectory()) {
-            throw new IOException("\"" + this.movieFolder + "\" is not a directory.");
+            if (!this.movieFolder.mkdirs()) {
+                logger.warn("Directory {} could not be created. mkdirs operation returned false.", this.movieFolder);
+            }
+        } else {
+            if(!this.movieFolder.isDirectory()) {
+                logger.warn("{} is not a directory.", this.movieFolder);
+            }
         }
 
         final File f = generateOutput(fileFormat);
--- a/src/main/java/org/jenkinsci/test/acceptance/recorder/JUnitScreenRecorder.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/recorder/JUnitScreenRecorder.java
@@ -1,0 +1,72 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2014 Red Hat, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.jenkinsci.test.acceptance.recorder;
+
+import org.junit.runner.Description;
+import org.monte.media.Format;
+import org.monte.media.Registry;
+import org.monte.screenrecorder.ScreenRecorder;
+
+import java.awt.*;
+import java.io.File;
+import java.io.IOException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+public class JUnitScreenRecorder extends ScreenRecorder {
+
+    private Description description;
+    private Format format;
+
+    public JUnitScreenRecorder(GraphicsConfiguration cfg, Rectangle captureArea, Format fileFormat,
+                               Format screenFormat, Format mouseFormat, Format audioFormat,
+                               File movieFolder, Description description) throws IOException, AWTException {
+        super(cfg, captureArea, fileFormat, screenFormat, mouseFormat, audioFormat, movieFolder);
+        this.format = fileFormat;
+        this.description = description;
+    }
+
+    protected File createMovieFile(Format fileFormat) throws IOException {
+        if(!this.movieFolder.exists()) {
+            this.movieFolder.mkdirs();
+        } else if(!this.movieFolder.isDirectory()) {
+            throw new IOException("\"" + this.movieFolder + "\" is not a directory.");
+        }
+
+        final File f = generateOutput(fileFormat);
+        return f;
+    }
+
+    public void removeMovieFile() {
+        final File output = generateOutput(this.format);
+        if(output.exists()) {
+            output.delete();
+        }
+    }
+
+    private File generateOutput(final Format fileFormat) {
+        final String filename = this.description.getClassName() + "-" + this.description.getMethodName();
+        return new File(this.movieFolder, filename + "." + Registry.getInstance().getExtension(fileFormat));
+    }
+}
--- a/src/main/java/org/jenkinsci/test/acceptance/recorder/JUnitScreenRecorder.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/recorder/JUnitScreenRecorder.java
@@ -7,10 +7,15 @@ import org.monte.screenrecorder.ScreenRecorder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.awt.*;
+import java.awt.GraphicsConfiguration;
+import java.awt.Rectangle;
+import java.awt.AWTException;
 import java.io.File;
 import java.io.IOException;
 
+/**
+ * Custom Screen Recorder that allow to rename video output file.
+ */
 public class JUnitScreenRecorder extends ScreenRecorder {
 
     private static final Logger logger = LoggerFactory.getLogger(JUnitScreenRecorder.class);
@@ -27,12 +32,12 @@ public class JUnitScreenRecorder extends ScreenRecorder {
     }
 
     protected File createMovieFile(Format fileFormat) throws IOException {
-        if(!this.movieFolder.exists()) {
+        if (!this.movieFolder.exists()) {
             if (!this.movieFolder.mkdirs()) {
                 logger.warn("Directory {} could not be created. mkdirs operation returned false.", this.movieFolder);
             }
         } else {
-            if(!this.movieFolder.isDirectory()) {
+            if (!this.movieFolder.isDirectory()) {
                 logger.warn("{} is not a directory.", this.movieFolder);
             }
         }
@@ -43,7 +48,7 @@ public class JUnitScreenRecorder extends ScreenRecorder {
 
     public void removeMovieFile() {
         final File output = generateOutput(this.format);
-        if(output.exists()) {
+        if (output.exists()) {
             output.delete();
         }
     }
--- a/src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
@@ -54,7 +54,9 @@ public class TestRecorderRule extends TestWatcher {
 
     @Override
     protected void starting(Description description) {
-        startRecording(description);
+        if(!isRecorderDisabled()) {
+            startRecording(description);
+        }
     }
 
     private void startRecording(Description des) {
@@ -103,6 +105,10 @@ public class TestRecorderRule extends TestWatcher {
     @Override
     protected void finished(Description description) {
         stopRecording();
+    }
+
+    private boolean isRecorderDisabled() {
+        return Boolean.getBoolean("RECORDER_DISABLED");
     }
 
     private boolean saveAllExecutions() {
--- a/src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
@@ -1,5 +1,6 @@
 package org.jenkinsci.test.acceptance.recorder;
 
+import org.jenkinsci.test.acceptance.utils.SystemEnvironmentVariables;
 import org.junit.rules.TestWatcher;
 import org.junit.runner.Description;
 import org.monte.media.Format;
@@ -30,13 +31,23 @@ public class TestRecorderRule extends TestWatcher {
     private static final int FRAME_RATE_PER_SEC = 60;
     private static final int BIT_DEPTH = 16;
     private static final float QUALITY_RATIO = 0.97f;
-    public static final String TARGET = "target";
+
+    private static final String TARGET = "target";
+    static final String OFF = "off";
+
+    static final String FAILURES = "failuresOnly";
+    static final String ALWAYS = "always";
+
+    private static final String DEFAULT_MODE = FAILURES;
+
+    static String RECORDER_OPTION = SystemEnvironmentVariables
+            .getEnvironmentOrPropertyVariable("RECORDER", DEFAULT_MODE).trim();
 
     private JUnitScreenRecorder screenRecorder;
 
     @Override
     protected void starting(Description description) {
-        if (!isRecorderDisabled()) {
+        if (isRecorderEnabled()) {
             startRecording(description);
         }
     }
@@ -91,12 +102,12 @@ public class TestRecorderRule extends TestWatcher {
         stopRecordingWithFinalWaiting();
     }
 
-    private boolean isRecorderDisabled() {
-        return Boolean.getBoolean("RECORDER_DISABLED");
+    private boolean isRecorderEnabled() {
+        return !OFF.equals(RECORDER_OPTION);
     }
 
     private boolean saveAllExecutions() {
-        return Boolean.getBoolean("RECORDER_SAVE_ALL");
+        return ALWAYS.equals(RECORDER_OPTION);
     }
 
     private void stopRecording() {
--- a/src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
@@ -1,26 +1,3 @@
-/*
- * The MIT License
- *
- * Copyright (c) 2014 Red Hat, Inc.
- *
- * Permission is hereby granted, free of charge, to any person obtaining a copy
- * of this software and associated documentation files (the "Software"), to deal
- * in the Software without restriction, including without limitation the rights
- * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
- * copies of the Software, and to permit persons to whom the Software is
- * furnished to do so, subject to the following conditions:
- *
- * The above copyright notice and this permission notice shall be included in
- * all copies or substantial portions of the Software.
- *
- * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
- * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
- * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
- * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
- * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
- * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
- * THE SOFTWARE.
- */
 package org.jenkinsci.test.acceptance.recorder;
 
 import org.junit.rules.TestRule;
@@ -32,6 +9,8 @@ import org.monte.media.FormatKeys;
 import org.monte.media.VideoFormatKeys;
 import org.monte.media.math.Rational;
 import org.monte.screenrecorder.ScreenRecorder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.awt.*;
 import java.io.File;
@@ -44,11 +23,13 @@ import java.util.concurrent.TimeUnit;
  */
 public class TestRecorderRule extends TestWatcher {
 
+    private static final Logger logger = LoggerFactory.getLogger(TestRecorderRule.class);
 
     private static final int MAX_RECORDING_TIME_SECS = 120000;
     private static final int FRAME_RATE_PER_SEC = 60;
     private static final int BIT_DEPTH = 16;
     private static final float QUALITY_RATIO = 0.97f;
+    public static final String TARGET = "target";
 
     private JUnitScreenRecorder screenRecorder;
 
@@ -65,7 +46,7 @@ public class TestRecorderRule extends TestWatcher {
                 .getDefaultScreenDevice()
                 .getDefaultConfiguration();
 
-        File movieFolder = new File("target");
+        File movieFolder = new File(TARGET);
 
         String mimeType = FormatKeys.MIME_QUICKTIME;
         String videoFormatName = VideoFormatKeys.ENCODING_QUICKTIME_ANIMATION;
@@ -86,17 +67,19 @@ public class TestRecorderRule extends TestWatcher {
                             outputFormatForScreenCapture, null, null, movieFolder, des);
             this.screenRecorder.start();
         } catch (IOException e) {
-            throw new IllegalArgumentException("Exception starting test recording.", e);
+            logger.warn("Exception starting test recording {}", e);
         } catch (AWTException e) {
-            throw new IllegalArgumentException("Exception starting test recording.", e);
+            logger.warn("Exception starting test recording {}", e);
         }
     }
 
     @Override
     protected void succeeded(Description description) {
         if (this.screenRecorder != null) {
-            stopRecording();
-            if(!saveAllExecutions()) {
+            if (saveAllExecutions()) {
+                stopRecordingWithFinalWaiting();
+            } else {
+                stopRecording();
                 this.screenRecorder.removeMovieFile();
             }
         }
@@ -104,7 +87,7 @@ public class TestRecorderRule extends TestWatcher {
 
     @Override
     protected void finished(Description description) {
-        stopRecording();
+        stopRecordingWithFinalWaiting();
     }
 
     private boolean isRecorderDisabled() {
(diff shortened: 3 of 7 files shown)


src/main/java/org/jenkinsci/test/acceptance/recorder/JUnitScreenRecorder.java
```diff
@@ -0,0 +1,72 @@
...
+                               Format screenFormat, Format mouseFormat, Format audioFormat,
+                               File movieFolder, Description description) throws IOException, AWTException {
+        super(cfg, captureArea, fileFormat, screenFormat, mouseFormat, audioFormat, movieFolder);
+        this.format = fileFormat;
+        this.description = description;
+    }
+
+    protected File createMovieFile(Format fileFormat) throws IOException {
+        if(!this.movieFolder.exists()) {
+            this.movieFolder.mkdirs();
```
oleg-nenashev: No check of <code>mkdirs()</code> result

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
```diff
@@ -0,0 +1,148 @@
...
+                compressorName, outputDimension,
+                bitDepth, quality, screenRate);
+
+        try {
+            this.screenRecorder = new JUnitScreenRecorder
+                    (gc, gc.getBounds(), getFileFormat(mimeType),
+                            outputFormatForScreenCapture, null, null, movieFolder, des);
+            this.screenRecorder.start();
+        } catch (IOException e) {
+            throw new IllegalArgumentException("Exception starting test recording.", e);
```
oleg-nenashev: It would be better to fairly propagate the exception, but seems it's not possible

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    private boolean saveAllExecutions() {
+        return Boolean.getBoolean("RECORDER_SAVE_ALL");
+    }
+
+    private void stopRecording() {
+        if (this.screenRecorder != null && this.screenRecorder.getState() == ScreenRecorder.State.RECORDING) {
+            try {
+                TimeUnit.SECONDS.sleep(1);
```
jtnord: sleep(1)?  why?
if it is needed for the recorder to sync or something then please add an inline comment.

src/main/java/org/jenkinsci/test/acceptance/recorder/JUnitScreenRecorder.java
lordofthejars: True I am going to check.

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
lordofthejars: Yes the ScreenRecorder API is a bit simple and you need to do some weird things to set the output filename, like extending a class.

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
lordofthejars: Yes I am going to put an inline because this is a bit hacky thing to make that last second of the recording is also recorded. Probably with 500ms would be enough but well I prefer putting 1 second.

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
```diff
@@ -0,0 +1,148 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2014 Red Hat, Inc.
```
jtnord: really?

src/main/java/org/jenkinsci/test/acceptance/recorder/JUnitScreenRecorder.java
```diff
@@ -0,0 +1,72 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2014 Red Hat, Inc.
```
jtnord: really?

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
lordofthejars: The funny thing is that I copied this from one ATH class, so it is somewhere too.

oleg-nenashev: Most probably it is a part implemented by @olivergondza . Personally I would vote for "Jenkins developers" for a highly shared code, but each company has its own policy.

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
oleg-nenashev: BTW any explicit delay makes the tests unstable on particular instances (debugging, slow build machines, etc.)

jtnord: kind of with @oleg on this this would add 506s to the cpu time taken to run the tests. (or over 1 minute in wall time).
when recording at 60fps you should only need a max of 1/60 of a second to make sure the last relevant capture has been made - which then gets the overall time added down to a managable ballpark of 1 second.

lordofthejars: I am going to try to adjust this time, but last time I tried using 1/60 didn't work. I suspect that the library isn't really capturing at 60 fps. I am going to try with 1/10 or so.

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
```diff
@@ -0,0 +1,148 @@
...
+            stopRecording();
+            if(!saveAllExecutions()) {
+                this.screenRecorder.removeMovieFile();
+            }
+        }
+    }
+
+    @Override
+    protected void finished(Description description) {
+        stopRecording();
```
jtnord: for future enhancements ([JUnitAttachmentsPlugin How to attach files](https://wiki.jenkins-ci.org/display/JENKINS/JUnit+Attachments+Plugin#JUnitAttachmentsPlugin-Howtoattachfiles))

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
olivergondza: How about logging a warning here? I do not think it is worth failing the test. Also, `IllegalArgumentException` does not seem right.

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
```diff
@@ -0,0 +1,148 @@
...
+        return Boolean.getBoolean("RECORDER_SAVE_ALL");
+    }
+
+    private void stopRecording() {
+        if (this.screenRecorder != null && this.screenRecorder.getState() == ScreenRecorder.State.RECORDING) {
+            try {
+                TimeUnit.SECONDS.sleep(1);
+                screenRecorder.stop();
+            } catch (IOException e) {
+                throw new IllegalArgumentException("Exception stopping test recording.", e);
```
olivergondza: Ditto.

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
lordofthejars: yes it is a good point, let's release without this enhancement and when we need it, we open an issue and I implement it.

olivergondza:
:+1:  on the approach.

src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
```diff
@@ -0,0 +1,106 @@
...
+    @Test
+    public void shouldRecordFailingTestExecutionByDefault() {
+
+        TestRecorderRule testRecorderRule = new TestRecorderRule();
+        Description shouldNotRecordSuccessTestExecutionByDefault = Description.createTestDescription("org.jenkinsci.test.acceptance.recorder.TestRecorderRuleTest", "shouldRecordFailingTestExecutionByDefault");
+        testRecorderRule.starting(shouldNotRecordSuccessTestExecutionByDefault);
+
+        System.out.println("Good Bye World");
+        //succeeded is not called since a failure is simulated
+        //testRecorderRule.succeeded(shouldNotRecordSuccessTestExecutionByDefault);
```
oleg-nenashev: NIT: Commented code without TODO

src/main/java/org/jenkinsci/test/acceptance/recorder/JUnitScreenRecorder.java
```diff
@@ -0,0 +1,55 @@
...
+        this.description = description;
+    }
+
+    protected File createMovieFile(Format fileFormat) throws IOException {
+        if(!this.movieFolder.exists()) {
+            if (!this.movieFolder.mkdirs()) {
+                logger.warn("Directory {} could not be created. mkdirs operation returned false.", this.movieFolder);
+            }
+        } else {
+            if(!this.movieFolder.isDirectory()) {
```
recena: @lordofthejars `if (`

src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
```diff
@@ -0,0 +1,106 @@
...
+        assertThat(outputFile.exists(), is(true));
+
+        //Clean the field
+        outputFile.delete();
+    }
+
+    @Test
+    public void shouldRecordSuccessTestExecutionWhenSaveAll() {
+        String oldValue = System.getProperty("RECORDER_SAVE_ALL");
+        System.setProperty("RECORDER_SAVE_ALL", "true");
```
oleg-nenashev: This approach may be unstable in the case of multithreaded testing. IIRC it is possible to make JUnit to run such tests in a single-thread mode only

src/main/java/org/jenkinsci/test/acceptance/recorder/JUnitScreenRecorder.java
```diff
@@ -0,0 +1,55 @@
...
+            }
+        }
+
+        final File f = generateOutput(fileFormat);
+        return f;
+    }
+
+    public void removeMovieFile() {
+        final File output = generateOutput(this.format);
+        if(output.exists()) {
```
recena: @lordofthejars Ditto

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
```diff
@@ -0,0 +1,146 @@
...
+    private static final int FRAME_RATE_PER_SEC = 60;
+    private static final int BIT_DEPTH = 16;
+    private static final float QUALITY_RATIO = 0.97f;
+    public static final String TARGET = "target";
+
+    private JUnitScreenRecorder screenRecorder;
+
+    @Override
+    protected void starting(Description description) {
+        if(!isRecorderDisabled()) {
```
recena: @lordofthejars Ditto

src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
```diff
@@ -0,0 +1,106 @@
...
+        System.out.println("Hello World");
+
+        testRecorderRule.succeeded(shouldNotRecordSuccessTestExecutionByDefault);
+        testRecorderRule.finished(shouldNotRecordSuccessTestExecutionByDefault);
+
+        File outputFile = new File("target", "org.jenkinsci.test.acceptance.recorder.TestRecorderRuleTest-shouldRecordSuccessTestExecutionWhenSaveAll.mov");
+        assertThat(outputFile.exists(), is(true));
+
+        //Clean the field
+        if(oldValue != null) {
```
recena: @lordofthejars Ditto

src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
lordofthejars: where is the TODO?

src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
```diff
@@ -0,0 +1,106 @@
...
+        System.out.println("Hello World");
+
+        //testRecorderRule.succeeded(shouldNotRecordSuccessTestExecutionByDefault);
+        testRecorderRule.finished(shouldNotRecordSuccessTestExecutionByDefault);
+
+        File outputFile = new File("target", "org.jenkinsci.test.acceptance.recorder.TestRecorderRuleTest-shouldNotRecordWhenRecorderIsDisabled.mov");
+        assertThat(outputFile.exists(), is(false));
+
+        //Clean the field
+        if(oldValue != null) {
```
recena: @lordofthejars Ditto

src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
lordofthejars: Well in case of multithreading you can only set this property once for all. This is just a unit test to test that this feature works but I don't expect users do this. This test is not an acceptance test but a test that test the test harness feature.

src/main/java/org/jenkinsci/test/acceptance/recorder/JUnitScreenRecorder.java
```diff
@@ -0,0 +1,55 @@
...
+    public JUnitScreenRecorder(GraphicsConfiguration cfg, Rectangle captureArea, Format fileFormat,
+                               Format screenFormat, Format mouseFormat, Format audioFormat,
+                               File movieFolder, Description description) throws IOException, AWTException {
+        super(cfg, captureArea, fileFormat, screenFormat, mouseFormat, audioFormat, movieFolder);
+        this.format = fileFormat;
+        this.description = description;
+    }
+
+    protected File createMovieFile(Format fileFormat) throws IOException {
+        if(!this.movieFolder.exists()) {
```
recena: @lordofthejars Ditto

src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
```diff
@@ -0,0 +1,106 @@
...
+        testRecorderRule.succeeded(shouldNotRecordSuccessTestExecutionByDefault);
+        testRecorderRule.finished(shouldNotRecordSuccessTestExecutionByDefault);
+
+        File outputFile = new File("target", "org.jenkinsci.test.acceptance.recorder.TestRecorderRuleTest-shouldNotRecordSuccessTestExecutionByDefault.mov");
+        assertThat(outputFile.exists(), is(false));
+
+        //Clean the field
+        outputFile.delete();
+    }
+    @Test
```
recena: @lordofthejars New line

recena:
@lordofthejars Maybe a silly question but, where is screen resolution configured?

lordofthejars:
Line 44 of TestRecorderRule.

README.md
```diff
@@ -21,6 +21,13 @@ Set `BROWSER=chrome` and install http://code.google.com/p/chromedriver/downloads
 There is a bit of a delay since we bring up Jenkins for every single test, with
 it's own sandboxed workspace.
 
+All executed tests are screen recorded by default, but only videos of failing tests are persited to `target` directory.
+By default, video file is named with the fully qualified test class name, minus sign (-) and the test method name.
```
jtnord: if the file is named differently then we can [automatically attach it in Jenkins](https://wiki.jenkins-ci.org/display/JENKINS/JUnit+Attachments+Plugin#JUnitAttachmentsPlugin-Howtoattachfiles).

src/main/java/org/jenkinsci/test/acceptance/recorder/JUnitScreenRecorder.java
```diff
@@ -0,0 +1,55 @@
...
+import org.monte.media.Registry;
+import org.monte.screenrecorder.ScreenRecorder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.awt.*;
+import java.io.File;
+import java.io.IOException;
+
+public class JUnitScreenRecorder extends ScreenRecorder {
```
jtnord: don't forget to add some documentation :-)

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
```diff
@@ -0,0 +1,146 @@
...
+import org.junit.runners.model.Statement;
+import org.monte.media.Format;
+import org.monte.media.FormatKeys;
+import org.monte.media.VideoFormatKeys;
+import org.monte.media.math.Rational;
+import org.monte.screenrecorder.ScreenRecorder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.awt.*;
```
amuniz: Please, avoid to use `*` imports (trying to search where is `GraphicsConfiguration` coming from, but I can not find it).

lordofthejars: Need to find where IntelliJ has this option configured.

src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
olivergondza: Perhaps I miss something, but when I tried this it recorded my current content of the screen and not the firefox window. If this happens when running in parallel (and that is what we rely on during CI build), I guess the test runs will fight over the screen making at least `N-1/N` of the recording useless or even confusing as it will record some other test.

I wanted to suggest to make this feature turned off by default. Perhaps it should run only when `forkCount == 1`.

README.md
lordofthejars: Yes, already commented in previous commit. We are going to do it for next release after today's demo.

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * JUnit Rule that before executing a test it starts a recording current screen
+ * and after the test is executed, it stops recording.
+ */
+public class TestRecorderRule extends TestWatcher {
+
+    private static final Logger logger = LoggerFactory.getLogger(TestRecorderRule.class);
+
+    private static final int MAX_RECORDING_TIME_SECS = 120000;
```
amuniz: 33 hours of recording?

recena:
@lordofthejars I saw this line but I know if it receives the screen resolution or the web browser window size.

src/main/java/org/jenkinsci/test/acceptance/recorder/JUnitScreenRecorder.java
recena: +1 to add some javadoc

src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+public class TestRecorderRuleTest {
+
+    @Test
+    public void shouldNotRecordSuccessTestExecutionByDefault() {
+        TestRecorderRule testRecorderRule = new TestRecorderRule();
+        Description shouldNotRecordSuccessTestExecutionByDefault = Description.createTestDescription("org.jenkinsci.test.acceptance.recorder.TestRecorderRuleTest", "shouldNotRecordSuccessTestExecutionByDefault");
+        testRecorderRule.starting(shouldNotRecordSuccessTestExecutionByDefault);
+
+        System.out.println("Hello World");
```
amuniz: Is this required?

src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
lordofthejars: Yes this is the normal behaviour . It records the screen since this is the only thing we can do now. Maybe with more effort on the project we would be able to detect the browser window, but I think it could be complicated to do it.

Moreover this project basically takes an screenshot every period of time and finally compose a video, not sure if you can record a minimized window.

I am not sure if turning off by default recording feature could be a good approach for adoption or not.

lordofthejars:
It is the screen resolution, the project simply takes screenshots every configured period and finally compose a video.

src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
amuniz: Simply remove commented out code, otherwise add a TODO explaining why you want to keep this code here.

src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
jtnord: @olivergondza FYI we are working on porting the ATH test run to workflow (we have this done and running internally but not yet available on the OSS CI server).  This means that the forkcount will always be 1 on the CI server in future as we run the tests on different slaves in parallel (which reduces the overall time taken to run these tests considerably!)

src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
lordofthejars: It is a test, I have added a `sout` because it is easier to follow the video that it is doing something (recording the screen). Ig not you see a plain screen that does nothing :)

src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
lordofthejars: But this is not a TODO, it is simply a comment on why we don't add this line. SO anyone who read the test in 5 months, don't get the temptation to read all tests and say, "look in all tests there is a succeeded call maybe this test is wrong let's add it".

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
lordofthejars: It is the default value, we can reduce but well it is safer to use a big value :)

recena:
@lordofthejars ah ok. Thank you.

src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
olivergondza: @jtnord, looking forward to see that implemented!

amuniz:
:bee: As far as I can see (didn't try by myself).

oleg-nenashev:
Why don't we run a CI for this library? It would allow to ensure the implementation works correctly on the PR builder. 

:ant: for merging before such kind of testing, but it's not a blocker

lordofthejars:
@oleg-nenashev How we can write a test for testing what you mention?

oleg-nenashev:
The best way would be to setup a  new PR Builder somewhere on https://jenkins.ci.cloudbees.com ...

recena:
@lordofthejars I agree with @oleg-nenashev

lordofthejars:
@oleg-nenashev keep in mind that tests of this project ATH, requires a lot of things to be run like Docker, Maven, .... So I am not really sure how do you want to test this.

oleg-nenashev:
ATH is supposed to be run on that instance: https://jenkins.ci.cloudbees.com/job/core/job/acceptance-test-harness/ , so I don't expect any infra issues

BTW a reduced test suite for the PR builder would be useful, because ATH is too long (CC @rsandell )

olivergondza:
@oleg-nenashev, I agree the PR builder would be great. There is a way to run a smoke test category or we can add *metatest* (testing the harness, not Jenkins and plugins) category. Though, It is desirable tu run all the tests IMO.

lordofthejars:
Ok as we talked in chat,  what @oleg-nenashev mention doesn't affect this PR but the whole project. So the idea is to get the required level to merge the change, and then open another issue to fix this problem globally.

jtnord:
@oleg-nenashev /cc @olivergondza we have the ATH run (all tests) down to <1hr from >6hr.

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
```diff
@@ -0,0 +1,147 @@
...
+        }
+    }
+
+    @Override
+    protected void finished(Description description) {
+        stopRecordingWithFinalWaiting();
+    }
+
+    private boolean isRecorderDisabled() {
+        return Boolean.getBoolean("RECORDER_DISABLED");
```
jglick: Properties with a negative sense are confusing. Should be `RECORDER_ENABLED` (or `ENABLE_RECORDER`), with the default if unset being true.

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    private boolean saveAllExecutions() {
+        return Boolean.getBoolean("RECORDER_SAVE_ALL");
```
jglick: For properties like these, prefer `System.getenv`. Then you need to use [this trick](https://github.com/jenkinsci/acceptance-test-harness/blob/935342b6b6da2bfe2bc257e5fbccbfc8a03771a5/pom.xml#L604-L623) to let people customize the value in a Maven profile.

src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
jglick: > Perhaps it should run only when `forkCount == 1`.

Makes sense, if that information is available to the test.

README.md
```diff
@@ -21,6 +21,13 @@ Set `BROWSER=chrome` and install http://code.google.com/p/chromedriver/downloads
...
+
+If you want to persist all videos, the ones that succeeded too, you can set `RECORDER_SAVE_ALL` Java system property to true.
+
+If you want to completely disable recording, you can set `RECORDER_DISABLED` Java system property to true.
```
jglick: Might be clearer to have one property, `RECORDER`, which could be set to

· `off`
· `failuresOnly`
· `always`

src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
jglick: Actually I sort of feel this should just be off by default. You want it turned on for CI servers, but not for interactive runs.

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
jtnord: @jglick why.  stick to system properties with and configure them with [systemPropertyVariables](http://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#)  These are much easier to set from IDEs as you can configure the JVM arguments on a per test basis - but configuring ssytem environment variables is much harder.

lordofthejars: We talked this with @jtnord and we agreed that it works better with Java System Property rather than an environment variable to avoid having to do all these things. Also if you are using with Eclipse M2e doesn't work pretty well and you need to start using `resources` tags and so on. But you mean only this property or both properties (SAVE_ALL and ENABLE)

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
lordofthejars: I have no real preference about this, but maybe it is true that positive is better, I am going to change it but then I need to do some tricky things since by default is enabled and I will need to check for null and so on in case of property is not set.

src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
lordofthejars: You prefer to have the recording feature disabled by default. But I think that since it is for ATH it will always interesting having this information. If we disabled by default, then it will be difficult to remember to enable it. Also it can happens that just when these random fails happens you had the recording disabled because you didn't remember to set the property.

README.md
lordofthejars: Good point. And default `failuresOnly` to avoid people forget about enabling it.

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
jglick: > stick to system properties with and configure them with systemPropertyVariables

You are welcome to prototype that. Last I checked it did not work with the Maven profile system already in use in this plugin. Also all the other flags/options for ATH are done as environment variables; change them all or none.

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
jglick: Right, it would be a null check.

jtnord: [`Boolean.getBoolean(SystemPropertyName)`](http://docs.oracle.com/javase/7/docs/api/java/lang/Boolean.html#getBoolean(java.lang.String)) never returns `null` - it returns a `boolean` primative

jglick: Right, that is point: if switching to positive sense, you need a slightly more verbose idiom:

```java
String v = System.getProperty("RECORDER_ENABLED");
return v == null ? true : Boolean.parseBoolean(v);
```

jtnord: righto gothca - but no `null` checks are needed and the following is easier to read (YMMV)
```java
String v = System.getProperty("RECORDER_ENABLED", /* enabled by default */ "true" );
return Boolean.parseBoolean(v);
```

lordofthejars: I am going to use the @jglick approach to use only one variable with three possible options.

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
lordofthejars: Yes it is true that we should use always the same strategy, if not it is going to be a problem to understand what is configured in which way. To avoid confusions I am going to implement it in both ways.

src/main/java/org/jenkinsci/test/acceptance/utils/SystemEnvironmentVariables.java
```diff
@@ -0,0 +1,30 @@
...
+ * it returns the configured value.
+ * If not then it checks if it is defined as java system property, and if not it returns the default value.
+ */
+public class SystemEnvironmentVariables {
+
+    public static final String getPropertyVariable(String property, String def) {
+        return System.getProperty(property, def);
+    }
+
+    public static final String getEnvironmentVariable(String property) {
```
jtnord: seems a little strange to have a function that does nothing that call an identical function.

src/main/java/org/jenkinsci/test/acceptance/utils/SystemEnvironmentVariables.java
```diff
@@ -0,0 +1,30 @@
...
+        return System.getenv(property);
+    }
+
+    public static final String getEnvironmentOrPropertyVariable(String property, String def) {
```
jtnord: I would argue that this is the wrong way around.
environment variables are global (more broad) so if a more specific way is used (system property) then it should use that.

The code is correct though.

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
```diff
@@ -0,0 +1,158 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(TestRecorderRule.class);
+
+    private static final int MAX_RECORDING_TIME_SECS = 120000;
+    private static final int FRAME_RATE_PER_SEC = 60;
+    private static final int BIT_DEPTH = 16;
+    private static final float QUALITY_RATIO = 0.97f;
+
+    private static final String TARGET = "target";
+    static final String OFF = "off";
```
jtnord: (just a comment - no need to change) personally I prefer not exposing the inner workings and then making the test use reflection to set the field - as it maintains the "this is for internal use only" contract. Others do have different opinions YMMV :-)

src/main/java/org/jenkinsci/test/acceptance/utils/SystemEnvironmentVariables.java
lordofthejars: mmm well I have seen both approaches in different projects,I can change to what you purpose since probably this is how you have done in other Jenkins components so we can be aligned as well.

jtnord: Other jenkins stuff does one or the other and does not offer a fall back AFAIK - I'm just thinking of the most specific (closest) first principal

jtnord:
there are some bits that could be tidied up ("Hello World") but :bee: on the approach.

lordofthejars:
This was also commented in other note. But as I said to make the video that is recording a bit dynamic, I chosen to print something to the console. In this way if someone in the future want to inspect how the videos looks like, they will see something that is moving on the screen.

lordofthejars:
Ok then the administrator is ready to merge the change :)

README.md
```diff
@@ -21,6 +21,18 @@ Set `BROWSER=chrome` and install http://code.google.com/p/chromedriver/downloads
 There is a bit of a delay since we bring up Jenkins for every single test, with
 it's own sandboxed workspace.
 
+All executed tests are screen recorded by default, but only videos of failing tests are persited to `target` directory.
+By default, video file is named with the fully qualified test class name, minus sign (-) and the test method name.
+
+You can configure what is persisted by using an environment variable or a Java system property called `RECORDER`.
```
jglick: Missing `pom.xml` section to specify this from a profile.

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
jglick: Eek, certainly prefer direct (package-level) access to a field over reflection.

src/main/java/org/jenkinsci/test/acceptance/utils/SystemEnvironmentVariables.java
```diff
@@ -0,0 +1,28 @@
+package org.jenkinsci.test.acceptance.utils;
+
+/**
+ * Gets a system environment variable. First it checks if the variable is an environment variable and if it is the case,
+ * it returns the configured value.
+ * If not then it checks if it is defined as java system property, and if not it returns the default value.
```
jglick: This is backwards.

src/main/java/org/jenkinsci/test/acceptance/utils/SystemEnvironmentVariables.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+public class SystemEnvironmentVariables {
+
+    public static final String getPropertyVariableOrEnvironment(String property, String def) {
```
jglick: If you are going to introduce configuration by system property, where all other configuration flags in this project currently use environment variables only, then to maintain consistency you should update all of the existing flags to use this call.

src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
oleg-nenashev: OK, agreed

src/main/java/org/jenkinsci/test/acceptance/utils/SystemEnvironmentVariables.java
olivergondza: FTR, I do not think there is any consistency at all. Envvars was inherited from the ruby times so some contributors kept using those, while a lot of new features started using properties.

Assuming the pom bestiality is needed just to support the old envvar approach, I see no reason not to move to properties (with some BC mechanism for existing envvars). I do not think this PR should address that but it would be nice if it used the mechanism we agree to use in the long run.

src/main/java/org/jenkinsci/test/acceptance/utils/SystemEnvironmentVariables.java
lordofthejars: This is why I don't like javadoc :stuck_out_tongue_winking_eye: I am going to fix it.

src/main/java/org/jenkinsci/test/acceptance/utils/SystemEnvironmentVariables.java
lordofthejars: As @olivergondza said I think this is not something in this PR but global strategy.

README.md
lordofthejars: Well this is something related to Maven not this PR and also in the rest of the documentation of ATH mention that you can use environment variables but does not provide an example of how to do it in Maven since we assume that we can use Maven or not and we know how to use the tool. But of course if we change here we will need to change everywhere.

rsandell:
:bee:

jtnord:
@reviewbybees done

README.md
jglick: > this is something related to Maven not this PR

Not at all. You are introducing a new property which it is desirable to define from a Maven profile.

jglick:
@olivergondza

> Assuming the pom bestiality is needed just to support the old envvar approach, I see no reason not to move to properties

The POM “bestiality” :-) is _able_ to support environment variables. Last I checked it was not able to support properties, but that no longer seems to be true. So for example if in `~/.m2/settings.xml` I have

```xml
<profile>
    <id>jenkins-acceptance-test</id>
    <properties>
        <JENKINS_WAR>/…/jenkinsci/jenkins/war/target/jenkins.war</JENKINS_WAR>
        <LOCAL_SNAPSHOTS>true</LOCAL_SNAPSHOTS>
    </properties>
</profile>
```

then I can use `-Pjenkins-acceptance-test` (or an IDE equivalent) to run tests with my preferred local settings. This works because in the POM,

```xml
<plugin>
  <artifactId>maven-surefire-plugin</artifactId>
  <configuration>
    <environmentVariables>
      <JENKINS_WAR>${JENKINS_WAR}</JENKINS_WAR>
    </environmentVariables>
  </configuration>
</plugin>
```

etc.  Probably this could be replaced with `<systemPropertyVariables>`. What is _not_ possible (AFAICT) is to automatically set properties without being enumerated; Ant makes this easy using some variants of `<property>` but it seems harder in Maven, or at least is not natively supported by `maven-surefire-plugin`. Perhaps there is some trick to do it.

BTW I was trying to test with `JdkTest` and `JdkTest.ORACLE_LOGIN`/`JdkTest.ORACLE_PASSWORD` variables, but `autoinstallJdk` is calling `System.getenv("ORACLE_LOGIN")` contrary to the documentation of `@TestActivation`. I think it needs `System.getProperty("JdkTest.ORACLE_LOGIN")`?

olivergondza:
@jglick to me it seems we can make the problem disappear (and remove the large and forever incomplete translation section from `pom.xml`) if we deprecate configuration via envvars (accepting the values from both sources for BC perhaps) and start relying on properties only. Since the envvar names matches the property names, it should be possible without braking any private profiles such as yours.

PR #40, opened 2015-09-27
[JENKINS-30390] Workflow job using Analysis Collector

amuniz:
Inspired by the hidden problem found in [JENKINS-30390](https://issues.jenkins-ci.org/browse/JENKINS-30390).

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/WorkflowJob.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/WorkflowJob.java
@@ -26,6 +26,7 @@ package org.jenkinsci.test.acceptance.po;
 
 import com.google.inject.Injector;
 import java.net.URL;
+import org.jenkinsci.test.acceptance.junit.Resource;
 
 @Describable("org.jenkinsci.plugins.workflow.job.WorkflowJob")
 public class WorkflowJob extends Job {
@@ -36,5 +37,10 @@ public class WorkflowJob extends Job {
 
     public final Control script = control("/definition/script");
     public final Control sandbox = control("/definition/sandbox");
+
+    public String copyResourceStep(String filePath) {
+        final Resource res = resource(filePath);
+        return String.format("sh '''%s'''%n", copyResourceShell(res, res.getName()));
+    }
 
 }
--- a/src/test/java/plugins/AnalysisCollectorPluginTest.java
+++ b/src/test/java/plugins/AnalysisCollectorPluginTest.java
@@ -13,7 +13,6 @@ import org.jenkinsci.test.acceptance.plugins.analysis_core.AnalysisConfigurator;
 import org.jenkinsci.test.acceptance.plugins.checkstyle.CheckStyleFreestyleSettings;
 import org.jenkinsci.test.acceptance.plugins.dashboard_view.DashboardView;
 import org.jenkinsci.test.acceptance.plugins.findbugs.FindBugsFreestyleSettings;
-import org.jenkinsci.test.acceptance.plugins.maven.MavenInstallation;
 import org.jenkinsci.test.acceptance.plugins.pmd.PmdFreestyleSettings;
 import org.jenkinsci.test.acceptance.plugins.tasks.TasksFreestyleSettings;
 import org.jenkinsci.test.acceptance.plugins.warnings.WarningsBuildSettings;
@@ -331,18 +330,14 @@ public class AnalysisCollectorPluginTest extends AbstractAnalysisTest<AnalysisCo
     }
 
     @Test
-    @WithPlugins({"git@2.4.0", "workflow-aggregator@1.10"})
+    @WithPlugins("workflow-aggregator")
     public void should_compute_annotations_on_workflow() {
-        MavenInstallation.installMaven(jenkins, "M3", "3.3.3");
         WorkflowJob job = jenkins.jobs.create(WorkflowJob.class);
-        // TODO: Job.copyResource(String) does not work for WorkflowJob, since it does not owns
-        // a workspace but the build does (actually it could own 0..N workspaces).
         job.script.set(
             "node {\n" +
-            "  git 'https://github.com/amuniz/maven-helloworld.git'\n" +
-            "  def mvnHome = tool 'M3'\n" +
-            "  sh \"${mvnHome}/bin/mvn -B -Dmaven.test.failure.ignore verify\"\n" +
-            "  step([$class: 'FindBugsPublisher', pattern: '**/findbugsXml.xml'])\n" +
+               job.copyResourceStep(ANALYSIS_COLLECTOR_PLUGIN_RESOURCES + "/checkstyle-result.xml") +
+               job.copyResourceStep(ANALYSIS_COLLECTOR_PLUGIN_RESOURCES +"/findbugs.xml") +
+            "  step([$class: 'FindBugsPublisher', pattern: '**/findbugs.xml'])\n" +
             "  step([$class: 'CheckStylePublisher'])\n" +
             "  step([$class: 'AnalysisPublisher'])\n" +
             "}");
@@ -356,8 +351,8 @@ public class AnalysisCollectorPluginTest extends AbstractAnalysisTest<AnalysisCo
         AnalysisCollectorAction action = new AnalysisCollectorAction(build);
         action.open();
 
-        assertThat(action.getNumberOfWarnings(), is(17)); // 1 from FB and 16 from CS
-        assertThat(action.getNumberOfNewWarnings(), is(17));
+        assertThat(action.getNumberOfWarnings(), is(FINDBUGS_ALL + CHECKSTYLE_ALL));
+        assertThat(action.getNumberOfNewWarnings(), is(FINDBUGS_ALL + CHECKSTYLE_ALL));
     }
 
     private AnalysisCollectorAction deselectPluginAndBuild(AnalysisPlugin plugin, Job job) {
--- a/src/test/java/plugins/AnalysisCollectorPluginTest.java
+++ b/src/test/java/plugins/AnalysisCollectorPluginTest.java
@@ -13,6 +13,7 @@ import org.jenkinsci.test.acceptance.plugins.analysis_core.AnalysisConfigurator;
 import org.jenkinsci.test.acceptance.plugins.checkstyle.CheckStyleFreestyleSettings;
 import org.jenkinsci.test.acceptance.plugins.dashboard_view.DashboardView;
 import org.jenkinsci.test.acceptance.plugins.findbugs.FindBugsFreestyleSettings;
+import org.jenkinsci.test.acceptance.plugins.maven.MavenInstallation;
 import org.jenkinsci.test.acceptance.plugins.pmd.PmdFreestyleSettings;
 import org.jenkinsci.test.acceptance.plugins.tasks.TasksFreestyleSettings;
 import org.jenkinsci.test.acceptance.plugins.warnings.WarningsBuildSettings;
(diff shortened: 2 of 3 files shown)


src/test/java/plugins/AnalysisCollectorPluginTest.java
```diff
@@ -329,6 +330,36 @@ public void should_aggregate_warnings_in_dashboard_portlet() {
         assertThat(dashboard, not(hasWarningsFor(job, TASKS, TASKS_ALL)));
     }
 
+    @Test
+    @WithPlugins({"git@2.4.0", "workflow-aggregator@1.10"})
+    public void should_compute_annotations_on_workflow() {
+        MavenInstallation.installMaven(jenkins, "M3", "3.3.3");
+        WorkflowJob job = jenkins.jobs.create(WorkflowJob.class);
+        // TODO: Job.copyResource(String) does not work for WorkflowJob, since it does not owns
+        // a workspace but the build does (actually it could own 0..N workspaces).
```
amuniz: Probably a new API in `org.jenkinsci.test.acceptance.po.WorkflowJob` is required. Something like `String copyResource(String filePath)` that would return a `writeFile` step using the given resource content in the step `text` parameter.

recena:
:bee:

src/test/java/plugins/AnalysisCollectorPluginTest.java
```diff
@@ -329,6 +330,36 @@ public void should_aggregate_warnings_in_dashboard_portlet() {
...
+        job.script.set(
+            "node {\n" +
+            "  git 'https://github.com/amuniz/maven-helloworld.git'\n" +
```
jglick: Do not like the idea of an automated test relying on some mutable repository in some personal account.

src/test/java/plugins/AnalysisCollectorPluginTest.java
jglick: :bug: so override these methods in `WorkflowJob` to do just that. (Or better, implement [my suggestion](https://github.com/jenkinsci/acceptance-test-harness/blob/935342b6b6da2bfe2bc257e5fbccbfc8a03771a5/src/main/java/org/jenkinsci/test/acceptance/po/Job.java#L50).)

src/test/java/plugins/AnalysisCollectorPluginTest.java
amuniz: Agree. I was thinking about it (thus the TODO in line 338). I can implement the proposal (described in [this  comment](https://github.com/jenkinsci/acceptance-test-harness/pull/40/files#r40503584)) as part of this PR if there is general agreement on it.

It would avoid the use of a external repository.

BTW we should fix this in [other places](https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/test/java/plugins/WorkflowPluginTest.java#L72).

src/test/java/plugins/AnalysisCollectorPluginTest.java
amuniz: Overriding `copyResource ` method in `WorkflowJob` does not make sense, which workspace the content is going to be copied to?

src/main/java/org/jenkinsci/test/acceptance/po/WorkflowJob.java
```diff
@@ -37,4 +47,20 @@ public WorkflowJob(Injector injector, URL url, String name) {
...
+    public String copyResourceStep(String filePath) {
+        final Resource res = resource(filePath);
+        try (InputStream in = res.asInputStream()) {
+            ByteArrayOutputStream out = new ByteArrayOutputStream();
+
+            try (OutputStream gz = new GZIPOutputStream(out)) {
+                IOUtils.copy(in, gz);
+            }
+
+            return String.format("sh '''base64 --decode << ENDOFFILE | gunzip > %1$s \n%2$s\nENDOFFILE'''\n",
```
olivergondza: This seems to duplicate what `Job#copyResource` does.

amuniz: Yes, the shell command is copied (after removing some parts) but the method signature (and the logic inside it) is different, since `Job#copyResource` returns `void` and makes use of `addShellStep` which is not valid for a `WorkflowJob`.

src/main/java/org/jenkinsci/test/acceptance/po/WorkflowJob.java
```diff
@@ -37,4 +47,20 @@ public WorkflowJob(Injector injector, URL url, String name) {
     public final Control script = control("/definition/script");
     public final Control sandbox = control("/definition/sandbox");
 
+    public String copyResourceStep(String filePath) {
```
tfennelly: Does the `filePath` arg need a `@Nonnull` annotation?

src/main/java/org/jenkinsci/test/acceptance/po/WorkflowJob.java
```diff
@@ -37,4 +47,20 @@ public WorkflowJob(Injector injector, URL url, String name) {
...
+        final Resource res = resource(filePath);
+        try (InputStream in = res.asInputStream()) {
```
tfennelly: Is the `resource` method new? I don't see it. Can it return `null` and should there be a `null` check on the `if` statement?

I would add the `null` check anyway (irrespective of `resource`s return being marked `@Nonnull`). I know others wouldn't agree with that but I'd respectfully agree to disagree with them :smile:

src/main/java/org/jenkinsci/test/acceptance/po/WorkflowJob.java
tfennelly: You should be able to use `%n` for the newlines (to get platform specific newlines).

src/test/java/plugins/AnalysisCollectorPluginTest.java
```diff
@@ -329,6 +330,33 @@ public void should_aggregate_warnings_in_dashboard_portlet() {
...
+        build.shouldSucceed();
+
+        assertThat(build, hasAction("Static Analysis Warnings"));
+
+        AnalysisCollectorAction action = new AnalysisCollectorAction(build);
+        action.open();
+
+        assertThat(action.getNumberOfWarnings(), is(FINDBUGS_ALL + CHECKSTYLE_ALL));
+        assertThat(action.getNumberOfNewWarnings(), is(FINDBUGS_ALL + CHECKSTYLE_ALL));
+    }
```
tfennelly: NIT: I always like to see some comments in the test code explaining the different parts of it - what's being tested, expected and why etc. I always think test code is more than just about testing .... it's also a really useful way for people to understand how things work, but that requires a few comments to guide.

src/main/java/org/jenkinsci/test/acceptance/po/WorkflowJob.java
amuniz: It's not marked as `@NonNull` but [it never returns null](https://github.com/jenkinsci/acceptance-test-harness/blob/c8a96001366835afd3d8fa15e4e6ed0dfd611443/src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java#L431), but throws `AssertionError` if the resource does not exist.

src/main/java/org/jenkinsci/test/acceptance/po/WorkflowJob.java
olivergondza: @amuniz, sorry, my point was to keep the shell magic on one place. So the both methods will delegate to another one that will assemble the actual script.

src/test/java/plugins/AnalysisCollectorPluginTest.java
```diff
@@ -329,6 +330,33 @@ public void should_aggregate_warnings_in_dashboard_portlet() {
         assertThat(dashboard, not(hasWarningsFor(job, TASKS, TASKS_ALL)));
     }
 
+    @Test
+    @WithPlugins({"git@2.4.0", "workflow-aggregator@1.10"})
```
varmenise: would it make sense to add `@Issue` with a link to the issue you re trying to test?

amuniz: Well, it was an issue related to `analysis-test-plugin` dependency which was solved some time ago. Moreover this is a general test to check that the plugins works fine with Workflow so it is not trying to catch any concrete issue.

olivergondza: What is the reason to require the explicit plugin versions? Is git used at all?

amuniz: Oops, git was used in the previous approach (there was a git clone as a step). I have to delete it. And a concrete WF version is not required neither, but I think it's a good practice to specify the version, don't you think so?

olivergondza: >  I have to delete it. And a concrete WF version is not required neither, but I think it's a good practice to specify the version, don't you think so?

Quite the contrary, the exact versions should be configurable by whoever runs the test, not who write it. This way you declare it has to be run with newer version or not at all, reducing usefulness of your test.

amuniz: Ok, agree. It's fixed.

olivergondza:
LGTM

tfennelly:
:bee:

src/main/java/org/jenkinsci/test/acceptance/po/WorkflowJob.java
oleg-nenashev: Oracle coding style presumes that unmarked parameters are always Nonnull. BTW FindBugs does not work in such way by default, hence it won't discover the issue if it happens. It's NIT for the test suite, but nice to have

oleg-nenashev:
:bee: , some minor improvements could be done

amuniz:
@reviewbybees done

src/test/java/plugins/AnalysisCollectorPluginTest.java
```diff
@@ -329,6 +329,32 @@ public void should_aggregate_warnings_in_dashboard_portlet() {
         assertThat(dashboard, not(hasWarningsFor(job, TASKS, TASKS_ALL)));
     }
 
+    @Test
+    @WithPlugins("workflow-aggregator")
+    public void should_compute_annotations_on_workflow() {
+        WorkflowJob job = jenkins.jobs.create(WorkflowJob.class);
+        job.script.set(
+            "node {\n" +
+               job.copyResourceStep("/analysis_collector_plugin/checkstyle-result.xml") +
```
uhafner: Use ANALYSIS_COLLECTOR_PLUGIN_RESOURCES constant.

src/test/java/plugins/AnalysisCollectorPluginTest.java
```diff
@@ -329,6 +329,32 @@ public void should_aggregate_warnings_in_dashboard_portlet() {
...
+               job.copyResourceStep("/analysis_collector_plugin/findbugs.xml") +
```
uhafner: ANALYSIS_COLLECTOR_PLUGIN_RESOURCES

src/main/java/org/jenkinsci/test/acceptance/po/WorkflowJob.java
```diff
@@ -37,4 +38,9 @@ public WorkflowJob(Injector injector, URL url, String name) {
     public final Control script = control("/definition/script");
     public final Control sandbox = control("/definition/sandbox");
 
+    public String copyResourceStep(String filePath) {
+        final Resource res = resource(filePath);
+        return String.format("sh '''%s'''%n", copyResourceShell(res, res.getName()));
```
jglick: I think this is OK, just be careful, since `'''` does not escape everything—in particular `'''` itself (obviously), but also `\` (less obviously). So long as the return value of `copyResourceShell` never includes either of these things it should work.

src/test/java/plugins/AnalysisCollectorPluginTest.java
jglick: True, though in this case I think what is really being asserted is that the build does not explode in a fireball. :-)

jglick:
@olivergondza wrote

bq.  This way you declare it has to be run with newer version or not at all, reducing usefulness of your test.

But in fact the test will only run using certain relatively new versions of `analysis-collector` and matching Workflow plugins, right? So what would be the point of producing test failures if someone tries to run it with older versions?

amuniz:
@uhafner fixed.

PR #41, opened 2015-09-30
[JENKINS-22252] Reproduced problem in test

jglick:
[JENKINS-22252](https://issues.jenkins-ci.org/browse/JENKINS-22252)

@reviewbybees

### DIFF
--- a/src/test/java/plugins/MavenPluginTest.java
+++ b/src/test/java/plugins/MavenPluginTest.java
@@ -49,7 +49,9 @@ import com.google.inject.Inject;
 import static org.hamcrest.CoreMatchers.*;
 import static org.hamcrest.MatcherAssert.*;
 import static org.jenkinsci.test.acceptance.Matchers.*;
+import org.jenkinsci.test.acceptance.junit.WithPlugins;
 import static org.jenkinsci.test.acceptance.plugins.maven.MavenInstallation.*;
+import org.jenkinsci.test.acceptance.plugins.tasks.TasksMavenSettings;
 
 public class MavenPluginTest extends AbstractJUnitTest {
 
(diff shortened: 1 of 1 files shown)


src/test/java/plugins/MavenPluginTest.java
```diff
@@ -246,4 +248,20 @@ private void assertHasModule(MavenModuleSet job, String name) {
         find(by.xpath("//a[@href='%s/']", name)).click();
         assertThat(driver.getCurrentUrl(), equalTo(build.module(name).url.toExternalForm()));
     }
+
+    @Issue("JENKINS-22252")
+    @WithPlugins({"maven-plugin@2.12", "tasks"})
```
amuniz: Is the plugin version specified for some reason? If the plugin version is fixed here, then this test is not going to catch possible regressions in future versions.

amuniz: Ok. This is actually reproducing the error exactly in this version. Nevermind.

amuniz: Well, reading the code again... I think my initial question still makes sense :smile:

oleg-nenashev: I suppose this version specification is a temporary thing and @jglick is going to update it to 2.13 after the release. Before that this test will be failing.

uhafner: Shouldn't this be the initial version required to get the test running? I.e., 2.12-SNAPSHOT or 2.13? It is not required to change it when 2.13 has been released. So we can better see in which version the fix has been applied.

olivergondza: It should be the first version where necessary features was introduced. As it is a regression introduced in 2.12 IIUC, there should be no version specified (presuming it does not depend on any recent features).

As a rule of thumb, consider someone is running this against a plugin version with `JENKINS-22252` present. If we declare this should run with first *fixed* version we effectively **hide** the presence of the bug from the user instead of pointing his attention to it.

jglick: > there should be no version specified (presuming it does not depend on any recent features)

It does not depend on any recent features. But if you do not specify a version, then the test will *not* test the version on the update center, currently 2.12. Instead it will use the 2.7.1 or whatever bundled¹ with `jenkins.war`. This was the way I found to force it to install version 2.12 or newer so that the test is meaningful.

¹One of the many irritating consequences of plugin bundling—hope to see it die soon.

olivergondza: Makes sense then, we should remove once we get rid of bundled plugins.

stephenc:
:bee:

jglick:
Now passes with 2.12.1 on the update center.

PR #42, opened 2015-09-30
Diagnose failures in WorkflowPluginTest.linearFlow

jglick:
Consistently failing on CI as for example [here](https://jenkins.ci.cloudbees.com/job/core/job/acceptance-test-harness/1233/testReport/plugins/WorkflowPluginTest/linearFlow/) but I cannot reproduce the failure locally. Looks like the build is failing prior to `input`, but my mistaken code idiom prevents this build log from being shown in ATH test output.

@reviewbybees

PR #44, opened 2015-10-01
Diagnosed failure in WorkflowPluginTest.linearFlow

jglick:
With #42 merged, I can see that the problem is a [weird Maven error](http://stackoverflow.com/q/29983683/12916); apparently the CI job is defining `M2_HOME`, so we are forced to override it. Using `withEnv` is more the [recommended idiom](https://github.com/jenkinsci/workflow-plugin/blob/master/TUTORIAL.md#managing-the-environment) anyway.

@reviewbybees


stephenc:
:bee:

jglick:
Combined with #34, passing for me locally against Jenkins 1.625 and Workflow 1.10.

PR #45, opened 2015-10-15
[FIXED JENKINS-30971] Refresh page after restart failure

lordofthejars:
[JENKINS-30971](https://issues.jenkins-ci.org/browse/JENKINS-30971)

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Jenkins.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Jenkins.java
@@ -2,6 +2,8 @@ package org.jenkinsci.test.acceptance.po;
 
 import static org.hamcrest.Matchers.not;
 import static org.jenkinsci.test.acceptance.Matchers.hasContent;
+import static org.jenkinsci.test.acceptance.Matchers.hasElement;
+
 import hudson.util.VersionNumber;
 
 import java.io.IOException;
@@ -11,6 +13,7 @@ import java.util.logging.Level;
 
 import org.jenkinsci.test.acceptance.controller.JenkinsController;
 import com.google.inject.Injector;
+import org.openqa.selenium.By;
 import org.openqa.selenium.TimeoutException;
 
 /**
@@ -115,11 +118,13 @@ public class Jenkins extends Node {
 
         try {
             waitFor(driver, not(hasContent("Please wait")), JenkinsController.STARTUP_TIMEOUT);
+            throw new TimeoutException();
         }catch(TimeoutException e) {
             //Let's try to avoid false negatives or not auto refresh
             visit(driver.getCurrentUrl());
             //we wait 10 seconds for refresh things.
-            waitFor(driver, hasContent("New Item"), 10);
+            //breadcrumbBarAnchor is only present when the Jenkins is ready
+            waitFor(driver, hasElement(By.className("breadcrumbBarAnchor")), 10);
         }
     }
 
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Jenkins.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Jenkins.java
@@ -11,6 +11,7 @@ import java.util.logging.Level;
 
 import org.jenkinsci.test.acceptance.controller.JenkinsController;
 import com.google.inject.Injector;
+import org.openqa.selenium.TimeoutException;
 
 /**
  * Top-level object that acts as an entry point to various systems.
@@ -112,7 +113,14 @@ public class Jenkins extends Node {
         visit("restart");
         clickButton("Yes");
 
-        waitFor(driver, not(hasContent("Please wait")), JenkinsController.STARTUP_TIMEOUT);
+        try {
+            waitFor(driver, not(hasContent("Please wait")), JenkinsController.STARTUP_TIMEOUT);
+        }catch(TimeoutException e) {
+            //Let's try to avoid false negatives or not auto refresh
+            visit(driver.getCurrentUrl());
+            //we wait 10 seconds for refresh things.
+            waitFor(driver, hasContent("New Item"), 10);
+        }
     }
 
     public JenkinsLogger getLogger(String name) {
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Jenkins.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Jenkins.java
@@ -118,7 +118,6 @@ public class Jenkins extends Node {
 
         try {
             waitFor(driver, not(hasContent("Please wait")), JenkinsController.STARTUP_TIMEOUT);
-            throw new TimeoutException();
         }catch(TimeoutException e) {
             //Let's try to avoid false negatives or not auto refresh
             visit(driver.getCurrentUrl());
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/Jenkins.java
```diff
@@ -112,7 +113,14 @@ public void restart() {
...
         clickButton("Yes");
 
-        waitFor(driver, not(hasContent("Please wait")), JenkinsController.STARTUP_TIMEOUT);
+        try {
+            waitFor(driver, not(hasContent("Please wait")), JenkinsController.STARTUP_TIMEOUT);
+        }catch(TimeoutException e) {
+            //Let's try to avoid false negatives or not auto refresh
+            visit(driver.getCurrentUrl());
+            //we wait 10 seconds for refresh things.
+            waitFor(driver, hasContent("New Item"), 10);
```
jtnord: if you have a secured instance (not sure if ATHs have this on restart?) then you may not see the "New Item" part.

lordofthejars: Any part that will be there for sure in any situation?

olivergondza: We use `not(hasContent("Please wait"))` to check it is loaded. Why changing it?

lordofthejars: Yes the problem is as I explained in JIRA, a lot of test fails because a timeout error is thrown. Of course this sometimes happens and sometimes not, so since timeout is set to 360 seconds, probably the Jenkins is restarted but the screen has not been refreshed. So what I do in these cases is force a refresh.

Also notice that I am not changing anything as you pointed out, simply adding a manual refresh in case of error timeout error. So I am not modifying anything but trying to solve a problem. Of course maybe this is not the problem, but since we cannot reproduce since in local always work and in Jenkins sometimes yes sometimes no, the only thing to proceed to avoid this kind of errors taht comes to my mind is doing in this way.

jtnord: so "please wait" will not be present on reloading.

so if this this is just a refresh issue then you could just do the same test.  The if this is a refresh issue the main wait should be fixed so that it a loop (so the total wait time is the same for timeout but individually if  a refresh works will be less)

lordofthejars: Ok we are going to change the condition to "Jenkins" in breadcumb tag.

jtnord:
:bee: if this works then we need to clean it up a little.

src/main/java/org/jenkinsci/test/acceptance/po/Jenkins.java
```diff
@@ -112,7 +116,16 @@ public void restart() {
         visit("restart");
         clickButton("Yes");
 
-        waitFor(driver, not(hasContent("Please wait")), JenkinsController.STARTUP_TIMEOUT);
+        try {
+            waitFor(driver, not(hasContent("Please wait")), JenkinsController.STARTUP_TIMEOUT);
+            throw new TimeoutException();
+        }catch(TimeoutException e) {
```
jglick: :ant: Strange code idiom. Assuming you really meant it to do what it seems to be doing, I think it would be clearer written thus:

```java
try {
    waitFor(driver, not(hasContent("Please wait")), JenkinsController.STARTUP_TIMEOUT);
catch (TimeoutException e) {/* ignore */}
visit(driver.getCurrentUrl());
waitFor(driver, hasElement(By.className("breadcrumbBarAnchor")), 10);
```

jglick:
Would be better to use `JenkinsLogWatcher.waitTillReady` in the case of `LocalController` at least, so that we have consistent logic between first start and restart. Some controllers cannot use that technique, such as `ExistingJenkinsController`, so we may need to introduce a new API into `Controller` that encapsulates the “waiting for Jenkins to be up and running” functionality that is otherwise bound into `startNow`.

src/main/java/org/jenkinsci/test/acceptance/po/Jenkins.java
lordofthejars: I uploded the incorrect code, the one used for testing.

src/main/java/org/jenkinsci/test/acceptance/po/Jenkins.java
```diff
@@ -112,7 +116,15 @@ public void restart() {
...
         clickButton("Yes");
 
-        waitFor(driver, not(hasContent("Please wait")), JenkinsController.STARTUP_TIMEOUT);
+        try {
+            waitFor(driver, not(hasContent("Please wait")), JenkinsController.STARTUP_TIMEOUT);
+        }catch(TimeoutException e) {
+            //Let's try to avoid false negatives or not auto refresh
+            visit(driver.getCurrentUrl());
+            //we wait 10 seconds for refresh things.
+            //breadcrumbBarAnchor is only present when the Jenkins is ready
```
jglick: And can I assume that is present even when there is no anonymous read access and you have been redirected to the login screen?

jglick:
:bee: probably works.

src/main/java/org/jenkinsci/test/acceptance/po/Jenkins.java
lordofthejars: Well this is a case, but what this logic is only executed randomly when the waitFor logic fails. So we see that the rate of timeout errors decreases then let's see if it is definitive or not.

jtnord:
:bee: and if it works can be cleaned up

olivergondza:
How inspecting `/api/json` to detect it is ready? It should the most portable of options.

src/main/java/org/jenkinsci/test/acceptance/po/Jenkins.java
olivergondza: @lordofthejars, you have missed my point. We are trying to detect when "Please wait" turns into the real page, but after your fix there are 2 mechanisms to detect that: `not(hasContent("Please wait"))` and `hasContent("New Item")`. We should agree on one. Anyway, there seems to be a better alternative, see my PR comment bellow.

And as James pointed out there is a way to save some time in happy scenario using `waitFor().withTimeout(..., ...).until(...)`.

lordofthejars:
@reviewbybees done

lordofthejars:
@olivergondza this is a first approach, if it works and we see that the error does not appear anymore, then we are going to work on the final solution and also investigate why this happens.

olivergondza:
https://github.com/jenkinsci/acceptance-test-harness/pull/46 should help pointing out if it is a page reload problem or not (capturing the threaddump).

lordofthejars:
@olivergondza well failing tests should be recorded (I implemented this, so this would be another way to see it) Now I am going against my brach and see what's happening.

lordofthejars:
@olivergondza in my branch the timeout error in this line has not occurred anymore.

lordofthejars:
@olivergondza I didn't see that this issue was not closed yet and I have appended another PR. We can do some git tricky things but I thing that we are going to complicate everything a bit.

olivergondza:
I agree, none of the few timeout failures I observed have any interesting activity on server side. Merging amended with https://github.com/jenkinsci/acceptance-test-harness/commit/7b1644295e7172586a34dd274540fbdcda3320f1.

PR #47, opened 2015-10-19
[FIXED JENKINS-31031] incorrect matcher in GradlePluginTest.

lordofthejars:
GradlePluginTest class contains a regular expression that is gradle --quiet but under some circumstances the output does not contain gradle --quiet but _ "..../gradle\" --quiet hello put quotes because of white space in the path. For this reason a new regular expression should be provided to fix this possibility.

@reviewbybees

### DIFF
--- a/src/test/java/plugins/GradlePluginTest.java
+++ b/src/test/java/plugins/GradlePluginTest.java
@@ -49,8 +49,7 @@ public class GradlePluginTest extends AbstractJUnitTest {
 
         job.startBuild().shouldSucceed()
                 .shouldContainsConsoleOutput("Hello world!")
-                //reg expression gradle[\w\\/"]* --quiet
-                .shouldContainsConsoleOutput("gradle[\\w\\\\/\"]* --quiet")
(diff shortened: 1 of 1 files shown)


src/test/java/plugins/GradlePluginTest.java
```diff
@@ -49,7 +49,8 @@ public void run_gradle_scirpt() {
 
         job.startBuild().shouldSucceed()
                 .shouldContainsConsoleOutput("Hello world!")
-                .shouldContainsConsoleOutput("gradle --quiet")
+                //reg expression gradle[\w\\/"]* --quiet
+                .shouldContainsConsoleOutput("gradle[\\w\\\\/\"]* --quiet")
```
olivergondza: `gradle.*--quiet` seems to read better and be less fragile in case of other options will be placed before that.

olivergondza:
:+1:

recena:
:bee:

oleg-nenashev:
:bee:

lordofthejars:
@reviewbybees done

PR #49, opened 2015-10-20
[FIXED JENKINS-31053] by accepting the modified content modal dialog

lordofthejars:
Rarely  a test fails because next exception is thrown:

```
org.openqa.selenium.UnhandledAlertException: Unexpected modal dialog (text: This page is asking you to confirm that you want to leave - data you have entered may not be saved.): This page is asking you to confirm that you want to leave - data you have entered may not be saved.
```

It seems that depending on previous execution, the browser contains some old values and when the new test tries to create a new item, since browser is in a form with some data, the modal dialg is shown.
To avoid this situation, pushing confirm button might be a good solution to avoid this problem instead of making test fails.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
@@ -12,10 +12,12 @@ import org.hamcrest.StringDescription;
 import org.jenkinsci.test.acceptance.junit.Resource;
 import org.jenkinsci.test.acceptance.junit.Wait;
 import org.jenkinsci.test.acceptance.utils.ElasticTime;
+import org.openqa.selenium.Alert;
 import org.openqa.selenium.By;
 import org.openqa.selenium.JavascriptExecutor;
 import org.openqa.selenium.NoSuchElementException;
 import org.openqa.selenium.StaleElementReferenceException;
+import org.openqa.selenium.UnhandledAlertException;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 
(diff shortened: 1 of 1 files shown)


stephenc:
:bee:

olivergondza:
https://github.com/jenkinsci/jenkins/pull/1854 promised to improve the situation. Are you using Jenkins version with that fix?

I am undecided here but it effectively turns off the check. Thinking of this further it has very little value for automated tests, since when you forget to hit save, the test will likely fail reliably. Thoughts?

lordofthejars:
@olivergondza The error appeared with 1.625.1.2-SNAPSHOT

amuniz:
Yes, https://github.com/jenkinsci/jenkins/pull/1854 fixes that issue.
If something is wrong in core code, it must solved in core, not workarounded in ATH, IMO.

Moreover, what if there were an ATH checking that the alert is shown? It would fail after this change (when it should not!)

lordofthejars:
The problem is that we are running some tests against Jenkins 1.625 which it has not fixed this. I agree what you mention but the logic I have touched is the one called in the initial of a test so in theory it should not be this alert as expected.

amuniz:
> The problem is that we are running some tests against Jenkins 1.625 which it has not fixed this

https://github.com/jenkinsci/jenkins/pull/1854 is going to be backported in 1.625.2 LTS

> the logic I have touched is the one called in the initial of a test so in theory it should not be this alert as expected

I don't understand that. The `visit` method can be called always and everywhere.
IMO, this workaround is not needed.

lordofthejars:
If we don't do this then we have randomly some fails in this method regarding the dialog

amuniz:
After the core fix they are not random, the alert shows up when it has to do it, and tests has to deal with it (as they do with any other Jenkins behavior).

lordofthejars:
yes but the version we are testing against does not have this fix.

amuniz:
Do we patch ATH for every bug found in Jenkins? Or do you fix the bug? 
I don't see why this case is different.

lordofthejars:
Ok if it is backported then we can avoid merging this PR.

jglick:
:bug:, I think this should be closed and we should rely on the core fix.

olivergondza:
I do not observe any further failures in recent builds. @amuniz, thanks!

PR #50, opened 2015-10-20
[FIXED JENKINS-31065] Override RECORDER property

lordofthejars:
Running Recorder tests might fail if the Jenkins node where the test is run overrides the RECORDER property. To fix this an override of a possible variable should be done.

@reviewbybees

### DIFF
--- a/src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
+++ b/src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
@@ -125,7 +125,7 @@ public class TestRecorderRuleTest {
         final String envVar = System.getenv("RECORDER");
         final String recorder = System.getProperty("RECORDER");
 
-        if ((envVar != null && !"".equals(envVar)) || (recorder != null && !"".equals(recorder))) {
+        if ((envVar != null && !envVar.isEmpty()) || (recorder != null && !recorder.isEmpty())) {
             System.setProperty("RECORDER", TestRecorderRule.FAILURES);
             return recorder;
         }
--- a/src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
+++ b/src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
@@ -2,6 +2,7 @@ package org.jenkinsci.test.acceptance.recorder;
 
 import org.jenkinsci.test.acceptance.guice.TestName;
 import org.jenkinsci.test.acceptance.junit.FailureDiagnostics;
+import org.jenkinsci.test.acceptance.utils.SystemEnvironmentVariables;
 import org.junit.Test;
 import org.junit.runner.Description;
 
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
```diff
@@ -28,11 +32,15 @@ public void shouldNotRecordSuccessTestExecutionByDefault() {
 
         //Clean the field
         outputFile.delete();
+
+        restoreRecorderOption(oldMode);
```
kzantow: Why not just do @Before / @After ?

jglick: :ant: agreed that would be the better style. Avoids duplicating this stuff in each test. Also would reset the option even if the test fails, unlike the current patch which neglects to use a `finally` block.

src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
```diff
@@ -105,4 +120,23 @@ private TestRecorderRule rule(Description desc) {
     private File outputFile(Description desc) {
         return new File("target/diagnostics/" +desc + "/ui-recording.mov");
     }
+
+    private String updateRecorderToDefaultOption() {
+        final String envVar = System.getenv("RECORDER");
+        final String recorder = System.getProperty("RECORDER");
+
+        if ((envVar != null && !"".equals(envVar)) || (recorder != null && !"".equals(recorder))) {
```
jglick: `String.isEmpty`

src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
```diff
@@ -105,4 +120,23 @@ private TestRecorderRule rule(Description desc) {
     private File outputFile(Description desc) {
         return new File("target/diagnostics/" +desc + "/ui-recording.mov");
     }
+
+    private String updateRecorderToDefaultOption() {
+        final String envVar = System.getenv("RECORDER");
+        final String recorder = System.getProperty("RECORDER");
```
jglick: This is ugly, we should not be allowing _both_ environment variables and system properties IMO.

lordofthejars: Well this has been discussed when we developed this and some need one others the other one so I implemented both.

src/test/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRuleTest.java
lordofthejars: Yes I thought the same, but the problem is that I think that we might need to set different options in each test, not done yet because I want to see if this works when is executed in Jenkins, and if it works I will need to set different values in some tests. So although now seems that it is the same code, it will change in near future.

olivergondza: As currently written, it violates test isolation in case some of the tests fails and the temporary property value will leak to unrelated tests. Both `finally` and `@After` (configuring property from test, resetting old value stored in a field in `@After`) approach will give you the flexibility you are looking for.

tfennelly:
:bee: 

But agree with the others wrt wrapping `updateRecorderToDefaultOption` and `restoreRecorderOption ` in `try`/`finally` (or use `@Before` / `@After`).

tfennelly:
:bee:

olivergondza:
+1

lordofthejars:
@reviewbybees done

PR #51, opened 2015-10-21
[FIXED JENKINS-31081] Adding JUnit Attachment support to screen recorer

lordofthejars:
ScreenRecorder shall be JUnitAttachment aware by implementing https://wiki.jenkins-ci.org/display/JENKINS/JUnit+Attachments+Plugin#JUnitAttachmentsPlugin-ByprintingoutthefilenameinaformatthatJenkinswillunderstand so the videos could be got from any execution.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/junit/FailureDiagnostics.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/junit/FailureDiagnostics.java
@@ -22,6 +22,8 @@ import com.google.inject.Inject;
 @TestScope
 public class FailureDiagnostics extends TestWatcher {
 
+    private static String JUNIT_ATTACHMENT = "[[ATTACHMENT|%s]]";
+
     private final File dir;
 
     @Inject
@@ -84,6 +86,19 @@ public class FailureDiagnostics extends TestWatcher {
                     FileUtils.deleteDirectory(dir);
                 } catch (IOException e) {
                     // Nah
+                }
+            }
+        }
+    }
+
+    @Override
+    public void failed(Throwable e, Description description) {
+        if (dir.exists()) {
+            File[] files = dir.listFiles();
+            if (files != null) {
+                for (File file : files) {
+                    //https://wiki.jenkins-ci.org/display/JENKINS/JUnit+Attachments+Plugin#JUnitAttachmentsPlugin-ByprintingoutthefilenameinaformatthatJenkinswillunderstand
+                    System.out.println(String.format(JUNIT_ATTACHMENT, file.getAbsolutePath()));
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


olivergondza:
This should be done for all files attached by `FailureDiagnostics`.

lordofthejars:
@olivergondza for what I understand FilaureDiagnostics saves things in `diagnostics` directory, but then you need to have access to them and the problem is that you might not have access to a Jenkins node to see the recording.

lordofthejars:
@olivergondza if it is not the case how can I get that files from JUnit report?

olivergondza:
I am not sure I follow, all files stored by diagnostics (and the recorder) are stored on the system where the testsuite is running. In case you are using ATH from Jenkins, it is most likely the Jenkins slave. So it should be stored in the workspace so the *JUnit Attachments plugin* should be able to collect that.

You fix created marker log lines only for recording, I suggest to create it in diagnostics for all the files left behind, not only recordings.

As the files can be created and then removed (removing recording in case the test succeeded), how about using `FailureDiagnostics#failed` to print a line for each file found in the diagnostics folder? It will *just work* for all kinds of diagnostics reported.

lordofthejars:
Well the recorder is not saved in FailureDiagnostics but in its own class. The problem is that JUnit attachment plugin expects a file in surefire-reports/<nameofclass> or printing a line on console, so things on `diagnostic` directory cannot be attached to JUnit report. Also as I said the Recorder uses its own way of storing files, and only uses `FailureDiagnostics` for getting the file name.

olivergondza:
Details aside the files are going to end in `/target/diagnostics/<TEST_NAME>`. I do not feel strongly if we move it to the dir where the plugin expects that (there should not be many users surprised as it is a new unreleased thing) or we use the marker line approach.

My whole point was to take other diagnostics (threaddums, html pages) and attach them as well. and as all these delegate to diagnostics subsystem for persistence, I suggested it should generate the marker lines and not whoever contribute diagnostics informations.

lordofthejars:
so I think that if we change the directory of FailureDiagnostic to match the JUnitAttachment expectations it would be perfect and easy. WDYT?

olivergondza:
Yes, provided we can predict the xml report name reliably.

lordofthejars:
@olivergondza implemented the JUnit attachment plugin support for FailureDiagnostics as we commented in previous comment.

olivergondza:
Thanks, I have configured that on https://jenkins.ci.cloudbees.com/job/core/job/acceptance-test-harness/ and https://jenkins.ci.cloudbees.com/job/core/job/acceptance-test-harness-stable/.

olivergondza:
Hm,
```
Attachment /scratch/jenkins/workspace/core/acceptance-test-harness-stable/target/diagnostics/trigger_batch_task_on_other_job(plugins.BatchTaskPluginTest)/ui-recording.mov was referenced from the test 'plugins.BatchTaskPluginTest' but it doesn't exist. Skipping.
```

Some of the files are not attached at all. Can you have a look?

PR #53, opened 2015-11-02
[Fixed JENKINS-31357] TestRecorderRule to handle HeadlessException

tfennelly:
[JENKINS-31357](https://issues.jenkins-ci.org/browse/JENKINS-31357)

Happens on Mac.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
@@ -62,29 +62,31 @@ public class TestRecorderRule extends TestWatcher {
     }
 
     private void startRecording(Description des) {
-        GraphicsConfiguration gc = GraphicsEnvironment
-                .getLocalGraphicsEnvironment()
-                .getDefaultScreenDevice()
-                .getDefaultConfiguration();
-
-        String mimeType = FormatKeys.MIME_QUICKTIME;
-        String videoFormatName = VideoFormatKeys.ENCODING_QUICKTIME_ANIMATION;
-        String compressorName = VideoFormatKeys.COMPRESSOR_NAME_QUICKTIME_ANIMATION;
-        Dimension outputDimension = gc.getBounds().getSize();
-        int bitDepth = BIT_DEPTH;
-        float quality = QUALITY_RATIO;
-        int screenRate = FRAME_RATE_PER_SEC;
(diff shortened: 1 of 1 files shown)


tfennelly:
@reviewbybees

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
```diff
@@ -62,29 +62,31 @@ protected void starting(Description description) {
...
+                    compressorName, outputDimension,
+                    bitDepth, quality, screenRate);
+
             this.screenRecorder = new JUnitScreenRecorder
                     (gc, gc.getBounds(), getFileFormat(mimeType),
                             outputFormatForScreenCapture, null, null, diagnostics);
             this.screenRecorder.start();
+        } catch (UnsupportedOperationException e) {
+            // E.g. java.awt.HeadlessException
+            logger.warn("Exception starting test recording {}", e);
```
oleg-nenashev: Why don't you catch HeadlessException only?

jglick:
[More readable diff](https://github.com/jenkinsci/acceptance-test-harness/pull/53/files?w=1)

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
jglick: Or simply

```java
} catch (Exception e) {
    logger.warn("Exception starting test recording {}", e);
}
```

If you have some kind of visceral dislike of catching `Exception` (personally I have never understood what the problem is—we know that _any_ kind of exception ought to be caught here, and reported) you could use

```java
} catch (HeadlessException | IOException | AWTException e) {
    logger.warn("Exception starting test recording {}", e);
}
```

tfennelly: I did originally and can put it back.

tfennelly: > if you have some kind of visceral dislike of catching Exception

I assume that comment is for @oleg-nenashev?

@oleg-nenashev seems to prefer a more specific exception while @jglick prefers a more general. I'd probably prefer the more general, but don't think it's really worth debating so how about we just leave it as I have it?

rsandell:
:bee:

PR #55, opened 2015-11-06
[FIXED JENKINS-31431] Warning Axis Test refresh browser

lordofthejars:
WarningPluginTest fails when showing the axis of all warning in the details section. In summary section you can see that there is 12 warnings but in the details section only 11 are shown. If you execute this test locally it works, so probably it can be a problem of refresh.

https://issues.jenkins-ci.org/browse/JENKINS-31431

@reviewbybees

### DIFF
--- a/src/test/java/plugins/WarningsPluginTest.java
+++ b/src/test/java/plugins/WarningsPluginTest.java
@@ -112,7 +112,7 @@ public class WarningsPluginTest extends AbstractAnalysisTest<WarningsAction> {
             assertThat(driver, hasContent(title + ": " + warningsPerAxis.get(axis.name)));
         }
 
-        // sometimes the details is not refreshed yet (https://cloudbees.atlassian.net/browse/CJP-3668) 
+        // sometimes the details is not refreshed yet (https://issues.jenkins-ci.org/browse/JENKINS-31431) 
         // so let's add an sleep and a refresh 
 
         try {
--- a/src/test/java/plugins/WarningsPluginTest.java
+++ b/src/test/java/plugins/WarningsPluginTest.java
@@ -4,6 +4,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.SortedMap;
 import java.util.TreeMap;
+import java.util.concurrent.TimeUnit;
 
 import org.jenkinsci.test.acceptance.junit.SmokeTest;
 import org.jenkinsci.test.acceptance.junit.WithPlugins;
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
Why you need sleep call if you do the refresh?

lordofthejars:
Because since it is a flaky test I don't know if a refresh is enough or also an sleep could be required. For this reason I prefer to put both conditions and see if the test is fixed or not.

oleg-nenashev:
Ok, :bee: from me

src/test/java/plugins/WarningsPluginTest.java
```diff
@@ -111,6 +112,15 @@ public void configure(WarningsBuildSettings settings) {
             assertThat(driver, hasContent(title + ": " + warningsPerAxis.get(axis.name)));
         }
 
+        // sometimes the details is not refreshed yet (https://cloudbees.atlassian.net/browse/CJP-3668)
```
rsandell: An internal issue tracking number that has no usage in OSS.
Maybe replace it with the link to JENKINS-31431?

oleg-nenashev: Nice catch. it must be replaced, of course

lordofthejars:
@rsandell done, can you give a bee?

src/test/java/plugins/WarningsPluginTest.java
```diff
@@ -111,6 +112,15 @@ public void configure(WarningsBuildSettings settings) {
             assertThat(driver, hasContent(title + ": " + warningsPerAxis.get(axis.name)));
         }
 
+        // sometimes the details is not refreshed yet (https://issues.jenkins-ci.org/browse/JENKINS-31431) 
+        // so let's add an sleep and a refresh 
+
+        try {
+            TimeUnit.SECONDS.sleep(1);
```
olivergondza: `CapybaraPortingLayerImpl#sleep` simplifies this.

lordofthejars: Going to fix then, I didn't know it exists this function.

lordofthejars: @olivergondza Done :) thanks for the suggestion.

rsandell:
:bee:

uhafner:
Did the refresh actually solve the problem? I'm wondering if there is no race condition somewhere in my warnings plug-in that is the cause for the flaky test.

amuniz:
I'd say this patch is making the issue less frequent, but not fixing it. IMO all cases like this one should be managed as a wait-and-check loop with a time wait limit.

lordofthejars:
the problem is that it cannot be a wait and check if there is no refresh call. So it can be a wait and check but always execute a refresh. The problem is that I am not sure that this is the problem it is happening. So this PR was adding small piece of code, see if this was the cause of the fail, and if it is the case then spent time to develop the final code, since now I cannot say that the problem is in the refresh or not.

PR #56, opened 2015-11-10
Need to be able to specify the socket

rsandell:
I have tests that controls two jenkins masters at the same time, so there is a need to be able to specify what socket the controller should use.

@reviewbybees


oleg-nenashev:
:bee:

scoheb:
I'd be interested in seeing an example of such a test!

rsandell:
@reviewbybees done

rsandell:
@olivergondza would it be possible to to release a new version with this change?

PR #54, opened 2015-11-04
[JENKINS-28769] Tests for ACE editor in workflow

tfennelly:
Upstream of https://github.com/jenkinsci/workflow-plugin/pull/234

PR #57, opened 2015-11-12
[JENKINS-28769] Allow Workflow tests to run with or without ACE

jglick:
Hotfix incorporating parts of #54 but allowing the existing tests to run whether we are using 1.10.1 (plain text area) or 1.11-beta-3+ (ACE editor).

#54 should be adjusted to do this more cleanly somehow TBD, as well as add its tests specific to ACE using `@WithPlugins("workflow-aggregator@1.11")`.

@reviewbybees esp. @tfennelly

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/WorkflowJob.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/WorkflowJob.java
@@ -26,7 +26,12 @@ package org.jenkinsci.test.acceptance.po;
 
 import com.google.inject.Injector;
 import java.net.URL;
+import javax.annotation.Nonnull;
 import org.jenkinsci.test.acceptance.junit.Resource;
+import org.junit.Assert;
+import org.openqa.selenium.JavascriptExecutor;
+import org.openqa.selenium.NoSuchElementException;
+import org.openqa.selenium.WebDriver;
 
 @Describable("org.jenkinsci.plugins.workflow.job.WorkflowJob")
 public class WorkflowJob extends Job {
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/WorkflowJob.java
```diff
@@ -35,7 +40,60 @@ public WorkflowJob(Injector injector, URL url, String name) {
         super(injector, url, name);
     }
 
-    public final Control script = control("/definition/script");
+    public final Control script = new Control(this, "/definition/script") {
+        @Override
+        public void set(String text) {
+            try {
+                super.set(text);
```
tfennelly: So is this even needed now, considering it is always using the latest version of the plugin, which will have the ACE Editor?

tfennelly:
:bee: 

Thanks Jesse, and sorry for the trouble. I thought we circle back on these after the release via #54.

amuniz:
:bee:

src/main/java/org/jenkinsci/test/acceptance/po/WorkflowJob.java
amuniz: > it is always using the latest version of the plugin

Unless the plugin is bundled, I think.

jglick: Not if you run the test against 1.580.x or 1.596, in which case 1.8 will not even be available.

src/main/java/org/jenkinsci/test/acceptance/po/WorkflowJob.java
```diff
@@ -35,7 +40,60 @@ public WorkflowJob(Injector injector, URL url, String name) {
...
+            }
+            try {
+                Thread.sleep(200);
+            } catch (InterruptedException e) {
+                break;
+            }
+        }
+        Assert.fail("Timed out waiting on '" + cssSelector + "' to be rendered.");
+    }
+    private static boolean isRendered(@Nonnull String cssSelector, @Nonnull WebDriver driver) {
```
olivergondza: `CapybaraPortingLayerImpl#executeScript`

src/main/java/org/jenkinsci/test/acceptance/po/WorkflowJob.java
```diff
@@ -35,7 +40,60 @@ public WorkflowJob(Injector injector, URL url, String name) {
...
+            if (isRendered(cssSelector, driver)) {
+                return;
+            }
+            try {
+                Thread.sleep(200);
+            } catch (InterruptedException e) {
+                break;
+            }
+        }
+        Assert.fail("Timed out waiting on '" + cssSelector + "' to be rendered.");
```
olivergondza: `CapybaraPortingLayerImpl#waitFor`

jglick: Maybe more a comment for #54, from which I copied this code.

src/main/java/org/jenkinsci/test/acceptance/po/WorkflowJob.java
jglick: ditto

jglick:
@olivergondza Workflow 1.11 is now released so expect tests to begin failing immediately. I suggest this be merged now and a nicer, more general system be developed in #54.

PR #61, opened 2015-11-25
OptionalJobProperty updates

jglick:
Fixes some test failures caused by https://github.com/jenkinsci/jenkins/pull/1888 and https://github.com/jenkinsci/jenkins/pull/1889.

@reviewbybees & @olivergondza

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Job.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Job.java
@@ -278,7 +278,13 @@ public class Job extends TopLevelItem {
     public <T extends Parameter> T addParameter(Class<T> type) {
         ensureConfigPage();
 
-        check(find(by.xpath("//input[@name='parameterized']")));
+        WebElement checkbox;
+        try {
+            checkbox = find(by.path("/properties/hudson-model-ParametersDefinitionProperty/specified"));
+        } catch (org.openqa.selenium.NoSuchElementException x) { // 1.636-
+            checkbox = find(by.xpath("//input[@name='parameterized']"));
+        }
+        check(checkbox);
 
         control(by.xpath("//button[text()='Add Parameter']")).selectDropdownMenu(type);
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/Job.java
```diff
@@ -278,7 +278,13 @@ public Build getLastBuild() {
     public <T extends Parameter> T addParameter(Class<T> type) {
         ensureConfigPage();
 
-        check(find(by.xpath("//input[@name='parameterized']")));
```
olivergondza: If I am not mistaken it can be replaced by `control("/properties/hudson-model-ParametersDefinitionProperty/specified", "//input[@name='parameterized']").check()`

rsandell:
:bee:

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
jglick: Possibly. I just wanted to disturb the old code branch as little as possible. Do you consider that cleanup a requirement? I would like to get this _tested_ change in as soon as possible so tests will pass again; any further edits mean a delay.

olivergondza: > Do you consider that cleanup a requirement?

Not really, I just feel really sorry for bees that such changes can take days to get integrated and someone have to pick the ball and do the cleanup once we see it fail/succeed. Is there any problem setting up a PR builder, or just nobody did that?

jglick: I suppose we could have a PR builder for this repo that runs all tests against both LTS and trunk Jenkins code. @andresrc might volunteer…?

jglick:
Something is still wrong which causes `JobParameterSummaryPluginTest.show_summary` to fail; investigating.

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
andresrc: @jglick @olivergondza sure :)

(sorry for the delay, I somehow missed the mention notification)

andresrc: @jglick @olivergondza done. Thanks!

PR #65, opened 2015-12-01
[JENKINS-31772] Allow JenkinsAcceptanceTestRule to be used as a ClassRule.

andresrc:
[[JENKINS-31772]](https://issues.jenkins-ci.org/browse/JENKINS-31772)

Add support to using JenkinsAcceptanceTestRule as a ClassRule. PR scope:
* Change JenkinsAcceptanceTestRule to be a TestRule instead of MethodRule.
* Add support to use the rule as a ClassRule.
* Adapt the FolderPluginTest to use the ClassRule.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/junit/AbstractATHTest.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/junit/AbstractATHTest.java
@@ -1,0 +1,32 @@
+package org.jenkinsci.test.acceptance.junit;
+
+import org.jenkinsci.test.acceptance.po.CapybaraPortingLayerImpl;
+
+import java.io.IOException;
+import java.net.ServerSocket;
+
+/**
+ * Abstract base class factoring out common code for test classes using the {@link JenkinsAcceptanceTestRule}.
+ *
+ * @author Andres Rodriguez
+ */
+public abstract class AbstractATHTest extends CapybaraPortingLayerImpl {
+
+    protected AbstractATHTest() {
+        super(null);
+    }
+
+    /**
+     * @return finds an unused, available port on the test machine
+     */
+    public int findAvailablePort() {
+        // use port 65000 as fallback (but maybe there is something running)
+        int port = 65000;
+        try (ServerSocket s = new ServerSocket(0)) {
+            port = s.getLocalPort();
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+        return port;
+    }
+}
--- a/src/main/java/org/jenkinsci/test/acceptance/junit/AbstractClassBasedJUnitTest.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/junit/AbstractClassBasedJUnitTest.java
@@ -1,0 +1,51 @@
+package org.jenkinsci.test.acceptance.junit;
+
+import org.jenkinsci.test.acceptance.po.CapybaraPortingLayerImpl;
+import org.jenkinsci.test.acceptance.po.Jenkins;
+import org.junit.ClassRule;
+import org.openqa.selenium.WebDriver;
+
+import javax.inject.Inject;
+import java.io.IOException;
+import java.net.ServerSocket;
+
+/**
+ * Convenience base class to derive your plain-old JUnit tests from when using {@link JenkinsAcceptanceTestRule} as a {@link ClassRule}.
+ *
+ * @author Andres Rodriguez
+ */
+public class AbstractClassBasedJUnitTest extends CapybaraPortingLayerImpl {
+
+    @ClassRule
+    public static JenkinsAcceptanceTestRule rules = JenkinsAcceptanceTestRule.classRule();
+
+    /**
+     * Jenkins under test.
+     */
+    @Inject
+    public static Jenkins jenkins;
+
+    /**
+     * This field receives a valid web driver object you can use to talk to Jenkins.
+     */
+    @Inject
+    public static WebDriver driver;
+
+    public AbstractClassBasedJUnitTest() {
+        super(null);
+    }
+
+    /**
+     * @return finds an unused, available port on the test machine
+     */
+    public int findAvailablePort() {
+        // use port 65000 as fallback (but maybe there is something running)
+        int port = 65000;
+        try (ServerSocket s = new ServerSocket(0)) {
+            port = s.getLocalPort();
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+        return port;
+    }
+}
--- a/src/main/java/org/jenkinsci/test/acceptance/junit/AbstractClassBasedJUnitTest.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/junit/AbstractClassBasedJUnitTest.java
@@ -14,7 +14,7 @@ import java.net.ServerSocket;
  *
  * @author Andres Rodriguez
  */
-public class AbstractClassBasedJUnitTest extends CapybaraPortingLayerImpl {
+public class AbstractClassBasedJUnitTest extends AbstractATHTest {
 
     @ClassRule
     public static JenkinsAcceptanceTestRule rules = JenkinsAcceptanceTestRule.classRule();
@@ -32,20 +32,5 @@ public class AbstractClassBasedJUnitTest extends CapybaraPortingLayerImpl {
     public static WebDriver driver;
 
     public AbstractClassBasedJUnitTest() {
-        super(null);
-    }
-
-    /**
-     * @return finds an unused, available port on the test machine
-     */
-    public int findAvailablePort() {
-        // use port 65000 as fallback (but maybe there is something running)
-        int port = 65000;
-        try (ServerSocket s = new ServerSocket(0)) {
-            port = s.getLocalPort();
-        } catch (IOException e) {
-            e.printStackTrace();
-        }
-        return port;
     }
 }
--- a/src/test/java/plugins/FolderPluginTest.java
+++ b/src/test/java/plugins/FolderPluginTest.java
@@ -26,21 +26,29 @@ package plugins;
 
 import org.hamcrest.MatcherAssert;
 import org.jenkinsci.test.acceptance.Matchers;
+import org.jenkinsci.test.acceptance.junit.AbstractClassBasedJUnitTest;
 import org.jenkinsci.test.acceptance.junit.AbstractJUnitTest;
 import org.jenkinsci.test.acceptance.junit.WithPlugins;
 import org.jenkinsci.test.acceptance.po.FolderItem;
+import org.junit.Assert;
+import org.junit.FixMethodOrder;
 import org.junit.Test;
+import org.junit.runners.MethodSorters;
 
 /**
  * Acceptance tests for the CloudBees Folder Plugins.
  */
 @WithPlugins("cloudbees-folder")
-public class FolderPluginTest extends AbstractJUnitTest {
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
+public class FolderPluginTest extends AbstractClassBasedJUnitTest {
     /** Test folder name. */
     private static final String F01 = "F01";
     /** Test folder name. */
     private static final String F02 = "F02";
-    
+
+    /** First test folder. */;
+    private static FolderItem f01;
+
     /**
      * Checks that a folder exists and has the provided name.
      */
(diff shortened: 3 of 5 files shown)


src/main/java/org/jenkinsci/test/acceptance/junit/AbstractATHTest.java
```diff
@@ -0,0 +1,32 @@
...
+public abstract class AbstractATHTest extends CapybaraPortingLayerImpl {
+
+    protected AbstractATHTest() {
+        super(null);
+    }
+
+    /**
+     * @return finds an unused, available port on the test machine
+     */
+    public int findAvailablePort() {
```
olivergondza: This should be moved to a helper class, instead. We might not need this class at all.

src/test/java/plugins/FolderPluginTest.java
```diff
@@ -26,21 +26,29 @@
...
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
+public class FolderPluginTest extends AbstractClassBasedJUnitTest {
     /** Test folder name. */
     private static final String F01 = "F01";
     /** Test folder name. */
     private static final String F02 = "F02";
-    
+
+    /** First test folder. */;
+    private static FolderItem f01;
```
olivergondza: -1 on sharing state

src/main/java/org/jenkinsci/test/acceptance/junit/AbstractATHTest.java
andresrc: +1

src/test/java/plugins/FolderPluginTest.java
andresrc: Was included as an example. Will fix. Thanks!

amuniz:
:bee: I would like a confirmation from you that it does not break previous tests. A good check would be to execute the `AnalysisSuite` (`mvn test -Dtest=AnalysisSuite`).

src/main/java/org/jenkinsci/test/acceptance/junit/AbstractClassBasedJUnitTest.java
```diff
@@ -0,0 +1,37 @@
...
+public class AbstractClassBasedJUnitTest extends CapybaraPortingLayerImpl {
+
+    @ClassRule
+    public static JenkinsAcceptanceTestRule rules = JenkinsAcceptanceTestRule.classRule();
+
+    /**
+     * Jenkins under test.
+     */
+    @Inject
+    public static Jenkins jenkins;
```
olivergondza: Will this be correctly reassigned per every test (while `rules` will not)?

andresrc: It is assigned per-class for all the tests in the class, that being the idea of a `ClassRule`.

Please take into account that the mechanism provided in this PR does not substitute the existing `AbstractJUnitTest`-based tests but complement the approach for those cases *specifically designed* to work at class level. The use cases for it are:
* The main one when using the `existing` driver (tests designed specifically for the case in which the JUT is an external one, so they are staeful).
* Tests classes designed so that the different methods don't interfere with each other, as is the case with the modified `FolderPluginTest`.

In both cases they get to share initialization code.

olivergondza:
Is there anything else, aside from the Jenkins startup time optimization, you are trying to address?

> The main one when using the  existing  driver (tests designed specifically for the case in which the JUT is an external one, so they are stateful).

In such a case, setup should be pretty quick, right?

> Tests classes designed so that the different methods don't interfere with each other, as is the case with the modified  FolderPluginTest .

Arguably, tests should not be designed in this way.

---

After this fix, all `RuleAnnotation`s applied to methods will be silently ignored.

andresrc:
> Is there anything else, aside from the Jenkins startup time optimization, you are trying to address?

I would call that a side effect, although an interesting one: having fast turnaround times encourages adding additional tests.

In any case the main driver for the change are cases where an external jenkins instance is the subject under test, so they are stateful. These suites:
* Must not be used instead of the real ATHs (though can add to it)
* Can be run on isolation against the external jenkins (e.g. a soak environment).

I believe that it is quite a legit scenario.

> In such a case, setup should be pretty quick, right?

Not always, there are, e.g., setups involving explicit reboots after checking plugins.

> After this fix, all `RuleAnnotations` applied to methods will be silently ignored.

Only for tests using the `ClassRule`. I will prepare an additional commit documenting:
* The expected use case for the use of `ClassRule`.
* Its implications (`RuleAnnotation` won't be observed).

Thanks!

andresrc:
@amuniz Executed the CheckStyle suite, :+1:

olivergondza:
> In any case the main driver for the change are cases where an external jenkins instance is the subject under test, so they are stateful. These suites:
> - Must not be used instead of the real ATHs (though can add to it)
> - Can be run on isolation against the external jenkins (e.g. a soak environment).

Sorry, I do not understand what you mean.

> there are, e.g., setups involving explicit reboots after checking plugins.

This should happen iff the plugin was installed and restart is needed for successful deployment - this can not be optimized away. If the plugin is installed (after previous test run against `existing` instance or by you external instance setup), the restart should be performed at all.

> Only for tests using the  ClassRule . I will prepare an additional commit documenting:
> - Its implications ( RuleAnnotation  won't be observed).

It would be better if we manage to fail the tests (or issue a warning) in case the annotations are ignored because of `ClassRule`.

andresrc:
Hi,

I have added additional commits addressing:
* A rule to detect method annotations on classes using class rule.
* Additional comments on expected uses of ClassRule

Thanks

jglick:
I am not following the principal intended use case either.

> An example would be test suites to perform tasks soak [?] test environment: populate data, check status, schedule jobs.

But test cases in a suite are supposed to be able to be run independently. So the example you describe would have to be written as one test case, in which case a method rule would work just as well.

Do you have some concrete example of what you are trying to accomplish with this PR? The only thing we have to look at is `FolderPluginTest`, which uses `@FixMethodOrder` for no reason that I can discern (the tests do not appear to refer to one another’s state), and seems to be using `@ClassRule` merely to save startup time—which I think would be better handled by continuing to use `@Rule` for guaranteed isolation, but making `jut-server.sh` work again (Oliver broke this a few months ago and [AFAIK never fixed it](https://github.com/jenkinsci/acceptance-test-harness/commit/7aaa05cd31d6afbd45e5b77ef1067c98767a1e3f#commitcomment-11434758)).

andresrc:
Ok, let me try to describe the intended use case better.

###### Stateful Environments as Subjects under Test

Maybe the first question we should consider is: do we consider stateful environments as useful SUTs?
I'm using "stateful environment" do refer to a jenkins instance that exists before the test is run and continue to exist (without any reset to its `JENKINS_HOME`) after test. Besides, several test rounds may be "fired" at these environments.

Examples of these SUTs would be, e.g., soak test environments, environments set up by installers, etc.

Is this use case interesting? my opinion is *yes*.

###### Specific Test Cases for those SUTs

The use case described above require specific test "arrangements": populate a certain mix of jobs and other configurations, schedule jobs with a certain pattern, etc.

The point here is: is the current ATH infrastructure suited for developing this test cases/suites, etc.? my opinion is: *sure*.

Just a small clarification, does this PR wants to change how current ATH cases are developed: *no*

###### How to develop these cases

Starting "greyer" lands here:
* Test isolation is not real in these cases, as the SUT is completely stateful.
* Let's embrace that on the way the tests are developed and don't try to hide it.

What does that *embracing* mean: in the intended use case, test scenarios are very likely to involve long series of steps, so
* Why not separate those steps in smaller, more manageable methods that are reported independently? The downside is, being the SUT stateful, method order coordination is required ( @jglick that's why there are `@FixMethodOrder` in `FolderPluginTest`: in a previous state of the PR this dependency was illustrated, but afterwards removed to show that it was not needed but the annotations remained ).
* As a next step, why not share initialization logic among them?

I know that there are strong positions with this: I'm not here to revive the old junit/testng debate or something like that. I'm talking about a specific use case here: A stateful jenkins SUT, as defined above. I don't think that qualifies as typical unit testing.

###### Could this be useful for the ATH use case?

The provided advantage would be reduced test initialization times, above all for those groups of tests that demand installation of plugins for many methods. The drawback is that it would be done at the expense of making tests more stateful. It's clear to me what the current guideline is and, as stated before, it is not the goal of this PR to change it.

Having the tests run as fast as possible is always a good encouragement to grow the suite, but I see, that for the current use case, pursuing other ways to do it, such as jut improvements (is there some ATH CI job configured to use the jut?), is the way to go.

jglick:
> Why not separate those steps in smaller, more manageable methods that are reported independently?

You can if you want, though if one of the earlier test cases fail, the later ones will still run (to failure) rather than being marked skipped, which seems to defeat the purpose of clearer reporting. Perhaps there is some more exotic JUnit runner designed for this use case.

> in a previous state of the PR this dependency was illustrated, but afterwards removed to show that it was not needed but the annotations remained

Then I would suggest reverting the change to `FolderPluginTest`. Making tests run faster is certainly a laudable goal, but I would rather waste some time on a server farm somewhere than waste _my_ time debugging a failure that is specific to a stateful suite.

> is there some ATH CI job configured to use the jut?

There is not currently a CI job using it (AFAIK), because it does not currently work. Would be a nice thing to fix in a separate PR. Might not be easy but you can at least check out historical versions of the harness where it was working to see what it did.

jglick:
To be clear, I am not saying there is anything wrong with making `@ClassRule JenkinsAcceptanceTestRule` work—just looking for what the real use case is.

olivergondza:
Ad time of the test run: I understood cloudbees (cc @jtnord) are working on parallel test run using workflow plugin. Will it resolve @andresrc's concerns? Is it running somewhere public? How can we help?

jtnord:
@olivergondza we have something for running things in parallel but it is not quite where we (cloudbees) want it for our internal stuff (we need some more support from junit and split tests for workflow).  That said it is likely useable for OSS - but we do run this on special on premise slaves not on the Dev@cloud slaves so we may need to ask the Jenkins board to supply some AWS slaves for this purpose (I need to see if the ATH workflow bits need this or not).  I've just got back from vacation and about to go on one for Christmas/new year so nothing will happen until next year.  

I also didn't follow exactly what @andresrc concerns where - if it was just time to execute then yes this would help that - but if the individual tests where made faster then this speedup would be compounded!

rsandell:
I based our internal panellisation on what @Vlatombe already had done in OSS https://jenkins.ci.cloudbees.com/job/core/job/acceptance-test-harness-wf/

olivergondza:
@rsandell, that looks very good. I do not see a reason why not to start using that for triggered ATH tests.

rsandell:
So where is this PR actually, I see no direct up or down votes.
Should we close it or merge it?

andresrc:
I still feel this would be a good addition, though I clearly could not make my case and I understand the concern that it may be abused to create flaky scenarios with the excuse of some performance improvement.

IIRC the `JenkinsRule` in JTH can be used as a `ClassRule` and it has not caused much problems.

If there is some consensus I'd happily take the first two commits and some refactor (moving the `findAvailablePort` to some helper) and put it in another PR (keeping this one closed as it is for reference).

olivergondza:
> IIRC the JenkinsRule in JTH can be used as a ClassRule and it has not caused much problems.

Looking at Core tests `@ClassRule` is used 3 times in total but only once with `JenkinsRule` - in a class with a single test method.

> If there is some consensus I'd happily take the first two commits and some refactor (moving the findAvailablePort to some helper) and put it in another PR (keeping this one closed as it is for reference).

It seems to be used by LDAP tests only. Both `login_no_ldap` and `login_use_fallback_server` in fact needs local port where no LDAP is running (which we can assume is virtually any port except the one where we have just started LDAP). The functionality seems duplicated by `LocalController#randomLocalPort` already. I guess we are safe to get rid of `findAvailablePort`.

jtnord:
>  It seems to be used by LDAP tests only. Both login_no_ldap and login_use_fallback_server in fact needs local port where no LDAP is running (which we can assume is virtually any port except the one where we have just started LDAP

No they need a port where nothing is listening - be that an LDAP server a WebServer or anything else random with a listening socket...

`LocalController#randomLocalPort` is abstract and the method is protected and inefficiently implemented for obtaining a random port which the OS can do in one call (as done by `findAvailablePort`) as it uses a loop to check every port in sequence.

olivergondza:
@jtnord, good point. The implementation of `findAvailablePort` is clearly superior.

oleg-nenashev:
Any plans to proceed? Seems it's hanging

PR #69, opened 2015-12-04
Resync EMMA test with updated JaCoCo test resource

stephenc:
e68e9583e8441899376096cf2fce2c4e98d4cf09 changed the resource in JaCoCo that the emma plugin uses.

That change means that the maven build execution now requires the `jacoco.version` property be defined.

This change defines a version (we don't care which as we are running `emma:emma` all we care is that it is a valid version that exists.

@reviewbybees

### DIFF
--- a/src/test/java/plugins/EmmaPluginTest.java
+++ b/src/test/java/plugins/EmmaPluginTest.java
@@ -61,6 +61,8 @@ public class EmmaPluginTest extends AbstractJUnitTest {
 
         // In the maven build step an Emma goal is added to enable coverage reporting.
         MavenBuildStep mbs = job.addBuildStep(MavenBuildStep.class);
+        mbs.properties("jacoco.version=0.7.5.201505241946");
+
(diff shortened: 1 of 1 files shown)


olivergondza:
Right, I have not noticed it is shared fixture. I find it quite surprising that emma cares for jacoco version. Can we use separated and jacoco prone project? Or, if we want to share, define default value for the version?

rsandell:
:bee:

oleg-nenashev:
Not the best fix due to the hardcode, but better than nothing. :bee:

stephenc:
@reviewbybees done

olivergondza:
@stephenc, are you going to fix what I and success kid have agreed on?

olivergondza:
Replaced by https://github.com/jenkinsci/acceptance-test-harness/commit/04c1a9926a29a1a47f5d554ea5d29874c6e6eae1

PR #71, opened 2015-12-10
Download plugin by version

olivergondza:
User can suggest plugin version using for example `git.version=1.42` to override latest. Much like `git.jpi` works.


scoheb:
How will this be used? ENV var?

Maybe add a mention of this to the docs.

olivergondza:
@scoheb: Sure, I have restructured the docs a bit.

PR #72, opened 2016-01-05
Update Selenium

jglick:
Without this, I am unable to run tests using Firefox 43.

@reviewbybees


olivergondza:
I hit this just yesterday. +1 on excluding the problematic dependency.

scoheb:
Could we get a new ATH release which includes this? Tx!

olivergondza:
Released 1.21 just now.

PR #74, opened 2016-01-07
Add posibility to use remote webdriver as a browser

rsandell:
For example the selenium/standalone-firefox-debug docker image.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/FallbackConfig.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/FallbackConfig.java
@@ -4,11 +4,13 @@ import javax.inject.Named;
 
 import java.io.File;
 import java.io.IOException;
+import java.net.URL;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
+import org.apache.commons.lang.StringUtils;
 import org.eclipse.aether.RepositorySystem;
 import org.eclipse.aether.RepositorySystemSession;
 import org.eclipse.aether.artifact.DefaultArtifact;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/test/acceptance/FallbackConfig.java
```diff
@@ -123,6 +126,14 @@ private WebDriver createWebDriver(TestName testName) throws IOException {
...
             capabilities.setCapability(LANGUAGE_SELECTOR_PHANTOMJS, "en");
             return new PhantomJSDriver(capabilities);
+        case "remote-webdriver-firefox":
+            String u = System.getenv("REMOTE_WEBDRIVER_URL");
+            if (StringUtils.isBlank(u)) {
+                throw new Error("remote-webdriver-firefox requires REMOTE_WEBDRIVER_URL to be set");
+            }
+            return new RemoteWebDriver(
+                    new URL(u), //http://192.168.99.100:4444/wd/hub
+                    DesiredCapabilities.firefox());
```
lordofthejars: You are fixing the capabilities to firefox but if you check https://selenium.googlecode.com/git/docs/api/java/org/openqa/selenium/remote/DesiredCapabilities.html you can see that more capabilities can be provided, I think that since there is a docker image for Firefox and another for Chrome, at least both of them should be supported.

rsandell: Well since the `BROWSER` var is locally used for firefox or chrome it would be easy to add `remote-webdriver-chrome` in the future. I don't see a particular reason to add it here now?

lordofthejars: Agreed I only said this comment to add the feature in this PR so it is full featured but no problem.

lordofthejars:
:bee:

rsandell:
@reviewbybees done

PR #75, opened 2016-01-11
TestRecorder fails in headless mode.

lordofthejars:
@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
@@ -13,10 +13,7 @@ import org.monte.screenrecorder.ScreenRecorder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.awt.GraphicsEnvironment;
-import java.awt.GraphicsConfiguration;
-import java.awt.Dimension;
-import java.awt.AWTException;
+import java.awt.*;
 import java.io.IOException;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
```diff
@@ -13,10 +13,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.awt.GraphicsEnvironment;
-import java.awt.GraphicsConfiguration;
-import java.awt.Dimension;
-import java.awt.AWTException;
+import java.awt.*;
```
rsandell: :ant: We prefer to have no star imports since that hinders review somewhat.

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
```diff
@@ -83,8 +81,10 @@ private void startRecording(Description des) {
                     (gc, gc.getBounds(), getFileFormat(mimeType),
                             outputFormatForScreenCapture, null, null, diagnostics);
             this.screenRecorder.start();
+        } catch (HeadlessException e) {
+            logger.warn("Test recorder does not work with Headless mode");
+            this.headless = true;
```
rsandell: why not just set `this.screenRecorder = null;` so you don't need all the extra complexity in the logics below?

rsandell:
:bee:

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
lordofthejars: I thought the same but since what it is really happening is that it is headless I preferred to mark as is. Null is like an error situation, headless is not an error simply it is a valid scenario, that I think that if we run with debug images and with some hacks we would be able to record as well.

src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java
lordofthejars: aaaah IntelliJ, why IntelliJ.

PR #76, opened 2016-01-18
[JENKINS-32477] Stop using deprecated methods and test fix

armfergom:
[JENKINS-32477](https://issues.jenkins-ci.org/browse/JENKINS-32477)

* Stop using deprecated method <code>org.jenkinsci.test.acceptance.po.Build.shouldContainsConsoleOutput</code>
* Use <code>elasticSleep</code> method to fix test that fails. DOM tree is not regenerated fast enough.

@reviewbybees

### DIFF
--- a/src/test/java/plugins/SubversionPluginTest.java
+++ b/src/test/java/plugins/SubversionPluginTest.java
@@ -1,7 +1,11 @@
 package plugins;
 
 import com.google.inject.Inject;
+
+import java.util.regex.Pattern;
+
 import org.hamcrest.CoreMatchers;
+import org.jenkinsci.test.acceptance.Matchers;
 import org.jenkinsci.test.acceptance.docker.DockerContainerHolder;
 import org.jenkinsci.test.acceptance.docker.fixtures.SvnContainer;
 import org.jenkinsci.test.acceptance.junit.*;
@@ -9,6 +13,7 @@ import org.jenkinsci.test.acceptance.plugins.subversion.SubversionPluginTestExce
 import org.jenkinsci.test.acceptance.plugins.subversion.SubversionScm;
 import org.jenkinsci.test.acceptance.plugins.subversion.SubversionSvmAdvanced;
 import org.jenkinsci.test.acceptance.plugins.subversion.SvnRepositoryBrowserWebSvn;
+import org.jenkinsci.test.acceptance.po.Build;
 import org.jenkinsci.test.acceptance.po.Changes;
 import org.jenkinsci.test.acceptance.po.FreeStyleJob;
 import org.junit.Test;
@@ -23,7 +28,7 @@ import static org.junit.Assert.*;
  *
  * @author Matthias Karl
  */
-@WithPlugins("subversion@2.3")
+@WithPlugins("subversion@2.5.7")
 @WithDocker
 public class SubversionPluginTest extends AbstractJUnitTest {
     @Inject
@@ -49,7 +54,8 @@ public class SubversionPluginTest extends AbstractJUnitTest {
         f.addShellStep("test -d .svn");
         f.save();
 
-        f.startBuild().shouldSucceed().shouldContainsConsoleOutput("test -d .svn");
+        Build b = f.startBuild().shouldSucceed();
+        assertThat(b.getConsole(), Matchers.containsRegexp("test -d .svn", Pattern.MULTILINE));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


recena:
LGTM, let me try it.

olivergondza:
I am -1 on bumping *required* subversion version. When you run the harness without any special config, it will download the latest version available anyway. To quote the javadoc:

> [The variant with version] declares that running the test with older version is pointless, typically because of missing feature.

... which is not the case.

armfergom:
@olivergondza You are completely right. With no version specified, the latest version of the plugin is downloaded (if there is no previous installation of the plugin). 

In the case of the subversion plugin though, it is included in the core, and by default, if I remember correctly, the version that is included is 1.54. Hence, if no version is specified, the test will be run against 1.54.

Could I please have your thoughts on this?

olivergondza:
> Hence, if no version is specified, the test will be run against 1.54.

That is not what I observe:

```
$ BROWSER=firefox JENKINS_WAR=jenkins-lts.war mvn test -Dtest=SubversionPluginTest
...
INFO: subversion has version 1.54 but 2.3 was requested
...
Plugin /home/ogondza/.m2/repository/org/jenkins-ci/plugins/subversion/2.5.7/subversion-2.5.7.hpi installed
...
```

src/test/java/plugins/SubversionPluginTest.java
```diff
@@ -49,7 +54,8 @@ public void run_basic_subversion_build() throws SubversionPluginTestException {
         f.addShellStep("test -d .svn");
         f.save();
 
-        f.startBuild().shouldSucceed().shouldContainsConsoleOutput("test -d .svn");
+        Build b = f.startBuild().shouldSucceed();
+        assertThat(b.getConsole(), Matchers.containsRegexp("test -d .svn", Pattern.MULTILINE));
```
jglick: :ant: This will match, for example, `test -d xsvn`. I think you wanted to use `Matchers.containsString`.

jglick:
@olivergondza right, you need to specify an explicit version if the bundled version is newer than the minimum version you care to test, which would not be the case here unless core started bundling, say, 2.4.

recena:
:+1: when @jglick's is addressed.

olivergondza:
> @olivergondza right, you need to specify an explicit version if the bundled version is newer than the minimum version you care to test, which would not be the case here unless core started bundling, say, 2.4.

@jglick, I am not sure I follow.

src/test/java/plugins/SubversionPluginTest.java
armfergom: @jglick The deprecated method suggested the use of the <code>containsRegexp</code> method in the Javadoc, but you are absolutely right,in this case it may match other strings.

I was not able to find the method <code>Matcher.containsString</code>, so I have used the <code>String.contains</code> method instead.

armfergom:
@olivergondza You are right, the actual behaviour is to download and install the latest version of the plugin. For that reason, I have set it back to 2.3 and narrowed the scope of the pull request down to just the removing of the deprecated methods and the test fix.

In any case, I believe that being able to specify a concrete version of the plugin would be quite useful in some cases, but that is not to be discussed here anyways.

amuniz:
:bee: (and +1 to add a new annotation `@WithPluginInRange(lower="2.3", upper="2.5")` or `@WithPluginInRange("2.3")` to use a concrete version)

recena:
I disagree with add a new annotation. I propose to improve the current `@WithPlugins`

olivergondza:
I disagree that specifying a plugin/core version should be a concern of people who *write* the tests, rather than those who *run* it. (With the exception when the author of the tests is sure it is going to report false negatives - that is what `@Since("X")` \ `@WithPlugins("...@X")` is for.)

In fact, we are refactoring plugin parent pom to gain the same advantage for plugin unit tests. The version of core can be specified at run time (not in pom.xml or even in every single test).

recena:
@olivergondza We would move this discussion to the mailing list in order to know others opinions. Design test for specific version range is helpful. I'm thinking on Subversion Plugin for example.

src/test/java/plugins/SubversionPluginTest.java
olivergondza: It is in `org.hamcrest.Matchers`. I am +1 to merge this.

armfergom: I'm confused, sorry! Is it a +1 when I change it to use the <code>org.hamcrest.Matchers.containsString</code> or you find it ok already?

olivergondza: With the matcher, sorry. The disadvantage of `String.contains` is miserable defect localization. All you know is it failed - not what the content was.

armfergom: Updated. Thanks for the feedback

PR #77, opened 2016-01-18
Trying to make WorkflowPluginTest.linearFlow less flaky

jglick:
Observed that after the restart, the command launcher tries to reconnect the slave, but always fails with an `EOFException` (currently have no idea why). At some point `ComputerRetentionWork` kicks in and runs `RetentionStrategy.Always.check`. It seems like this should happen within one minute, but I have seen it take nearly two full minutes after startup before an attempt is made to reconnect. By that time the test timeout waiting for build completion may have expired.

@reviewbybees

### DIFF
--- a/src/test/java/plugins/WorkflowPluginTest.java
+++ b/src/test/java/plugins/WorkflowPluginTest.java
@@ -102,6 +102,9 @@ public class WorkflowPluginTest extends AbstractJUnitTest {
         jenkins.restart();
         visit(build.getConsoleUrl());
         clickLink("Proceed");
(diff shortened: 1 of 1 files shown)


amuniz:
:bee:

olivergondza:
Waiting until slave comes online after restart makes sense. Hope people will not find to too surprising.

andresrc:
@olivergondza the comment seems positive but the PR is closed. Could you elaborate?
Thanks!

olivergondza:
My apology, I merged in UI and click the other button. Integrated in c5e0e5bfd8d7d53a48eb5574f8f7fc6d6c7d66c5.

tfennelly:
@olivergondza so is this PR staying closed? It's not clear (to me anyway) from your previous post.

olivergondza:
While the PR says closed, the whole change was merged to master. Github does not seem to change the PR status from closed to merged then merge commit arrive. Sorry for the mess.

PR #78, opened 2016-01-22
[FIXED JENKINS-32569] - ArtifactManagement: Add a delay to the control cleanup cycle

oleg-nenashev:
The change gives slow JS engines a chance to cleanup the previous control before we click on the button again.
It fixes ArtifactManager and ParameterizedTriggerTest::triggerWithNonStandardArchiver() on my machine.

https://issues.jenkins-ci.org/browse/JENKINS-32569

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/ArtifactManagement.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/ArtifactManagement.java
@@ -46,8 +46,14 @@ public class ArtifactManagement extends PageAreaImpl {
 
     public void clear() {
         try {
-            while (true) {
+            while (control("artifactManagerFactories/repeatable-delete").exists()) {
                 control("artifactManagerFactories/repeatable-delete").click();
+                // We allow slow Javascripts to remove the deleted element before we click again (JENKINS-32569)
+                try {
+                    Thread.sleep(500);
+                } catch (InterruptedException ex) {
+                    // just ignore and continue the cycle
+                }
             }
         } catch (NoSuchElementException ignored) {
             //done no more buttons to push


olivergondza:
Looking at the stacktrace, it might be `Scroller` bug hiding the expected `NoSuchElement` exception.

```
...
at com.sun.proxy.$Proxy31.executeScript(Unknown Source)
at org.jenkinsci.test.acceptance.selenium.Scroller.scrollIntoView(Scroller.java:81)
at org.jenkinsci.test.acceptance.selenium.Scroller.beforeClickOn(Scroller.java:72)
...
```

`Scroller` should detect if the element (still) exists before calling JS methods on it.

---

Scroller is a utility that makes sure the element we are about to interact with is in the visible part of the document and scroll if necessary.

oleg-nenashev:
@olivergondza 
I was able to reproduce this issue in the debugger.
When the Scroller invokes the Javascript, the element really exists according to the the field data. So it seems to be a race condition with JS scripts cleaning up the page. Is there is a way to synchronize them somehow? Would be useful for other tests as well

olivergondza:
IMO the JS script should be made immune to disappearing elements. It is invoked when we are about to interact with it using selenium so it would throw the expected exception.

src/main/java/org/jenkinsci/test/acceptance/po/ArtifactManagement.java
```diff
@@ -46,8 +46,14 @@ public ArtifactManagement(JenkinsConfig config) {
...
     public void clear() {
         try {
-            while (true) {
+            while (control("artifactManagerFactories/repeatable-delete").exists()) {
                 control("artifactManagerFactories/repeatable-delete").click();
+                // We allow slow Javascripts to remove the deleted element before we click again (JENKINS-32569)
+                try {
+                    Thread.sleep(500);
+                } catch (InterruptedException ex) {
+                    // just ignore and continue the cycle
```
rsandell: I'm not sure it's really safe to catch it like this since an InterruptedException usually indicates something is actively trying to stop the thread from running anymore and tries to be nice about it, you should be nice back and exit gracefully.

stephenc: Or if you are not being nice, you could at least restore the interrupted flag ;-)

olivergondza: Even better use `CapybaraPortingLayerImpl#sleep()`/`CapybaraPortingLayerImpl#elasticSleep()`.

olivergondza:
I guess this was fixed already.

PR #79, opened 2016-02-24
[JENKINS-33107]  Docker Fixtures might fail using Docker-Machine

lordofthejars:
Add `DOCKER_HOST` ip resolution.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
@@ -5,6 +5,7 @@ import org.jenkinsci.utils.process.CommandBuilder;
 
 import java.io.File;
 import java.io.IOException;
+import java.net.URI;
 
 /**
  * Container image, a template to launch virtual machines from.
@@ -12,10 +13,14 @@ import java.io.IOException;
  * @author Kohsuke Kawaguchi
  */
 public class DockerImage {
+
+    public static final String DEFAULT_DOCKER_HOST = "127.0.0.1";
     public final String tag;
+    DockerHostResolver dockerHostResolver;
 
     public DockerImage(String tag) {
         this.tag = tag;
+        dockerHostResolver = new DockerHostResolver();
     }
 
     public <T extends DockerContainer> T start(Class<T> type, CommandBuilder options, CommandBuilder cmd,int portOffset) throws InterruptedException, IOException {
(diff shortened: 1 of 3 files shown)


tfennelly:
:bee: LGTM from what I can see

src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
```diff
@@ -12,10 +13,14 @@
...
  */
 public class DockerImage {
+
+    public static final String DEFAULT_DOCKER_HOST = "127.0.0.1";
     public final String tag;
+    DockerHostResolver dockerHostResolver;
 
     public DockerImage(String tag) {
         this.tag = tag;
+        dockerHostResolver = new DockerHostResolver();
```
rsandell: should probably use `@Inject` here instead for nicer IoC wiring during tests. But that might be a bit overkill in this case.

lordofthejars: Well this class has been added to make test possible, so as you said maybe an overkill of using `@Inject`

tfennelly: > should probably use `@Inject` here instead ...

Nooooooooo .... that sh1t is not nicer .... it's putrid ... sorry, I really hate how that crap ends up being used everywhere, just coz it can be used.

lordofthejars: hehehe then I suspect that you prefer my approach :stuck_out_tongue:

tfennelly: Totally !!!

[KISS](https://en.wikipedia.org/wiki/KISS_principle).

rsandell: Well one way of keeping it simple is to follow the conventions in the rest of the system, suddenly switching conventions in the middle of it all can confuse casual observers of the code. But I have no strong opinions of it in this particular case.

rsandell:
:bee:

lordofthejars:
@reviewbybees done

@olivergondza This PR is ready to be merged :)

src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
tfennelly: True, but `dockerHostResolver = new DockerHostResolver()` is just regular Java `new` syntax and is not a change of convention of any kind.

lordofthejars: I understand the point of @rsandell, but in this case the logic is quite simple to do a new.

rsandell:
@lordofthejars Do you need a release?

lordofthejars:
@rsandell Yes so I can update everywhere where it is used. Thanks :)

rsandell:
@lordofthejars should be in 1.24 it's always a bit iffy if the release actually succeeded or not :)

PR #81, opened 2016-02-29
[JENKINS-27152] Test of standardized temp directory with Git operations

jglick:
[JENKINS-27152](https://issues.jenkins-ci.org/browse/JENKINS-27152)

@reviewbybees


src/test/java/plugins/WorkflowPluginTest.java
```diff
@@ -158,4 +168,31 @@
         build.shouldContainsConsoleOutput("divided into 3 sets");
     }
 
+    @WithDocker
+    @WithPlugins({"workflow-aggregator@1.14", "docker-workflow@1.14-SNAPSHOT", "git", "ssh-agent@1.10-SNAPSHOT"})
+    @WithCredentials(credentialType=WithCredentials.SSH_USERNAME_PRIVATE_KEY, values={"git", "/org/jenkinsci/test/acceptance/docker/fixtures/GitContainer/unsafe"}, id="gitcreds")
+    @Issue("JENKINS-27152")
+    @Test public void sshGitInsideDocker() throws Exception {
+        // Pending https://github.com/jenkinsci/docker-workflow-plugin/pull/31 (and additional APIs in DockerFixture etc.)
```
jglick: https://github.com/jenkinsci/docker-workflow-plugin/pull/31 that is.

src/test/java/plugins/WorkflowPluginTest.java
```diff
@@ -158,4 +168,31 @@
         build.shouldContainsConsoleOutput("divided into 3 sets");
     }
 
+    @WithDocker
+    @WithPlugins({"workflow-aggregator@1.14", "docker-workflow@1.14-SNAPSHOT", "git", "ssh-agent@1.10-SNAPSHOT"})
```
jglick: Downstream of https://github.com/jenkinsci/docker-workflow-plugin/pull/33 and https://github.com/jenkinsci/ssh-agent-plugin/pull/11.

src/test/java/plugins/WorkflowPluginTest.java
```diff
@@ -158,4 +168,31 @@
...
+        // it is not possible to run this on a bind-mounted Docker slave, so we cannot verify that paths are local to the slave.
+        GitContainer container = gitServer.get();
+        String host = container.host();
+        int port = container.port();
+        GitRepo repo = new GitRepo();
+        repo.commit("Initial commit");
+        repo.transferToDockerContainer(host, port);
+        WorkflowJob job = jenkins.jobs.create(WorkflowJob.class);
+        job.script.set(
+            "node {ws('" + tmp.getRoot() + "') {\n" + // TODO UNIX_PATH_MAX workaround
```
jglick: Not happy about this but not sure what else to do @cyrille-leclerc: if you have a long workspace path, putting the SSH agent socket in or near it will not work, so the SSH Agent plugin will fall back to a global temp dir that will not be accessible from inside the Docker container.

Once https://github.com/jenkinsci/jenkins/pull/2066 makes it into an LTS and we call it from both sides, that method could be adjusted to give an option to use a shortened path, say `/tmp/<SHA-1-of-workspace>`, which would be deterministic and unique, but nowhere near the workspace. Or `/<slave-FS-root>/<some-hash>`, to retain the property of being under the slave FS root and thus amenable to cleanup, bulk `chmod` operations, etc.?

recena: @jglick I like this one `/tmp/<SHA-1-of-workspace>`

src/test/java/plugins/WorkflowPluginTest.java
```diff
@@ -158,4 +168,31 @@
...
+            "  docker.image('cloudbees/java-build-tools').inside {\n" +
+            "    git url: '" + container.getRepoUrlInsideDocker() + "', credentialsId: 'gitcreds'\n" +
+            "    sh 'mkdir ~/.ssh && echo StrictHostKeyChecking no > ~/.ssh/config'\n" +
```
recena: I would add something like `chmod 700 ~./ssh`

amuniz:
:bee:

BTW why is [acceptance-test-harness](https://jenkins.ci.cloudbees.com/job/core/job/acceptance-test-harness/) job building against 2.0? There is a [specific job](https://jenkins.ci.cloudbees.com/job/core/job/acceptance-test-harness-2.0/) for that.

recena:
LGTM :bee:

src/test/java/plugins/WorkflowPluginTest.java
jglick: There are no private files in this directory, only configuration, and anyway it is just a test.

jglick:
> why is acceptance-test-harness job building against 2.0?

You would have to ask @olivergondza, I am not sure.

olivergondza:
> why is acceptance-test-harness job building against 2.0?

We are copying the artifact from core/jenkins-core and in case it build RP against 2.0 we got these bits. ATH should run against master branch, ideally.

amuniz:
> We are copying the artifact from core/jenkins-core

Then we need an specific job to build master (excluding PRs) and make ATH job to use that one. Should we move this conversation to the mailing list?

PR #82, opened 2016-03-03
[JENKINS-33292] More test cases for LDAP plugin

armfergom:
Increate test coverage for the LDAP plugin.

* Add support for adding environment variables in tests.
* Fix minor issues when selecting cache options
* Add more tests to increase coverage

@reviewbybees

### DIFF
--- a/src/test/java/plugins/LdapPluginTest.java
+++ b/src/test/java/plugins/LdapPluginTest.java
@@ -89,6 +89,17 @@ public class LdapPluginTest extends AbstractJUnitTest {
         assertThat(jenkins, hasLoggedInUser("jenkins"));
     }
 
+    @Test
+    public void login_ok_anonymous_binding() {
+        // Given
+        useLdapAsSecurityRealm(createDefaultsWithoutManagerCred(ldap.get()));
+        // When
+        Login login = jenkins.login();
+        login.doLogin("jenkins", "root");
+        // Then
+        assertThat(jenkins, hasLoggedInUser("jenkins"));
+    }
+
     /**
      * Scenario: Login with ldap uid and a wrong password
      * Given I have a docker fixture "ldap"
@@ -205,14 +216,6 @@ public class LdapPluginTest extends AbstractJUnitTest {
         assertThat(jenkins, hasLoggedInUser("jenkins@jenkins-ci.org"));
     }
     
-    /**
-     * Scenario: login with email address
-     * Given I have a docker fixture "ldap"
-     * And Jenkins is using ldap as security realm with user search filter "invalid={0}"
-     * When I login with email "jenkins@jenkins-ci.org" and password "root"
-     * Or when I login with user "jenkins" and password "root"
-     * Then I will not be able to log in
-     */
     @Test
     public void invalid_user_search_filter() {
         // Given
@@ -280,14 +283,6 @@ public class LdapPluginTest extends AbstractJUnitTest {
         assertThat(u, mailAddressIs("jenkins@jenkins-ci.org"));
     }
 
-    /**
-     * Scenario: do not resolve email address
-     * Given I have a docker fixture "ldap"
-     * And Jenkins is using ldap as security realm and resolve email address is disabled
-     * When I login with user "jenkins" and password "root"
-     * Then I will be logged on as user "jenkins"
-     * And there will not be any resolved email address"
-     */
     @Test
     public void do_not_resolve_email() {
         // Given
@@ -303,14 +298,6 @@ public class LdapPluginTest extends AbstractJUnitTest {
         assertThat(u.mail(), nullValue());
     }
     
-   /**
-   * Scenario: do not resolve email address
-   * Given I have a docker fixture "ldap"
-   * And Jenkins is using ldap as security realm and cache is enabled
-   * When I login with user "jenkins" and password "root"
-   * Then I will be logged on as user "jenkins"
-   * Nothing much can be tested here apart from being able to use the security realm with the option activated
-   */
     @Test
     public void enable_cache() throws IOException {
         // Given
@@ -324,14 +311,6 @@ public class LdapPluginTest extends AbstractJUnitTest {
         assertThat(jenkins, hasLoggedInUser("jenkins"));
     }
     
-    /**
-     * Scenario: can use environment variables
-     * Given I have a docker fixture "ldap"
-     * And Jenkins is using ldap as security realm and the java.naming.ldap.typesOnly env var is se to true
-     * When I login with user "jenkins" and password "root"
-     * Then I will not be logged on as user "jenkins" because the values for the LDAP attributes
-     * have not been sent as defined in the environment variable.
-     */
     @Test
     public void use_environment_varibales() {
         // Given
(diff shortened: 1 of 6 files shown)


src/test/java/plugins/LdapPluginTest.java
```diff
@@ -73,6 +90,24 @@ public void login_ok() {
     }
 
     /**
+     * Scenario: Login with ldap uid and password
+     * Given I have a docker fixture "ldap" that allows anonymous binding
+     * And Jenkins is using ldap as security realm
+     * When I login with user "jenkins" and password "root" without manager credentials
+     * Then I will be successfully logged on as user "jenkins"
```
olivergondza: Note, these are just a relic from BDD days and should not be added to new junit tests. I guess I should just go delete these comments everywhere not to confuse people.

src/test/java/plugins/LdapPluginTest.java
```diff
@@ -240,6 +299,72 @@ public void resolve_email() {
...
+        // Given
+        LdapDetails details = createDefaultsWithoutManagerCred(ldap.get());
+        details.addEnvironmentVariable(new LdapEnvironmentVariable("java.naming.ldap.typesOnly", "true"));
+        useLdapAsSecurityRealm(details);
+        // When
+        Login login = jenkins.login();
+        login.doLogin("jenkins", "root");
+        // Then
+        assertThat(jenkins, not(hasLoggedInUser("jenkins")));
+        assertThat(getElement(by.tagName("pre")).getText(), Matchers.containsString("NoSuchElementException"));
```
olivergondza: This seems very fragile to me.

src/test/java/plugins/LdapPluginTest.java
armfergom: Thanks for the info Oliver. Nothing is mentioned in the Wiki:

> If your LDAP server doesn't support anonymous binding (IOW, if your LDAP server doesn't even allow a query without authentication), then Jenkins would have to first authenticate itself against the LDAP server, and Jenkins does that by sending "manager" DN and password.

And I obviously wasn't aware of that. Test case removed. Again, thanks for the heads up.

src/test/java/plugins/LdapPluginTest.java
armfergom: Yes... may be in case the exception thrown is different. In any case, what I wanted to test is that the environment variables are actually being used, and from [documentation](http://docs.oracle.com/javase/7/docs/technotes/guides/jndi/jndi-ldap.html) this one seemed like an easy one to test. Removed the exception assertion and just checked that the user was not logged in, that should be alright.

src/test/java/plugins/LdapPluginTest.java
olivergondza: Sorry, my comment was related to the Gherkin-like comments and not the test itself. It just duplicates whatever the test is doing anyway. If the test is hard to understand(which it is not in this case), we should fix the framework/POs so it is cloear what is going on and not to duplicate the information in text.

armfergom: My bad @olivergondza . Read the comment to quickly and I misunderstood. If the convention is not to add the comments, I'll remove them. I'm of the opinion that comments are always good though :)

andresrc:
:bee:

armfergom:
@reviewbybees done

src/test/java/plugins/LdapPluginTest.java
```diff
@@ -205,7 +249,7 @@ public void login_use_fallback_server() {
         int freePort = this.findAvailablePort();
         LdapDetails ldapDetails = new LdapDetails("", 0, ldapContainer.getManagerDn(), ldapContainer.getManagerPassword(), ldapContainer.getRootDn());
         // Fallback-Config: primary server is not running, alternative server is running docker fixture
-        ldapDetails.setHostWithPort("localhost:" + freePort + " localhost:" + ldapContainer.getPort());
+        ldapDetails.setHostWithPort(ldapContainer.getHost()+":" + freePort + " " + ldapContainer.getHost() + ":" + ldapContainer.getPort());
```
jtnord: the first change is incorrect - it should be attempting to connect against a server that does not exist - and you have no gaurantee that freePort isactually a freeport on the ldapContainer (hence it should use localhost).

armfergom: @jtnord Unless I am misunderstanding something, I think this is not incorrect. You cannot guarantee either that there is no LDAP server running on localhost. Actually, if the docker container is being run on Linux, the LDAP server will be running on localhost. The only way to actually be sure that there is no Ldap server listening on a given server and a given port is to use a port where you know no Ldap server will be listening. In this case 65000. The Ldap container exposes 636 and 389.

armfergom: I think it works just fine in both cases, using `ldapContainer.getHost()` or `localhost`. What matters is the port.

jtnord: > Unless I am misunderstanding something

Yes.  You are missing the fact that this does [`findAvailablePort()`](https://github.com/jenkinsci/acceptance-test-harness/blob/caddb01a1b21bf19f12de6cd8cf687b63bafcec2/src/main/java/org/jenkinsci/test/acceptance/junit/AbstractJUnitTest.java#L45-L54) which will find a port that is not in use on the local system so there is nothing running on that.  It also binds to it and does not set SO_REUSE_ADDR so it should not be re-cycled for a period of time (which should be longer than the time it takes to run the tests).

> I think it works just fine in both cases, using ldapContainer.getHost() or localhost. What matters is the port.

Not true - networking plays a role here

The `ldapContainer.getHost()` will return the host of the docker process that you are talking to (this may not be localhost).
You can not guarantee that there is not something listening on that specific port on a remote machine.

armfergom: Will change it. Just want to point out that this issue will still be present for localhost and linux. Because `ldapContainer.getHost() == localhost` in that case..

jtnord: > Just want to point out that this issue will still be present for localhost and linux.

That's not an issue are `findAvailableFreePort()` finds a port that there is nothing listening on - and as I stated there is guaranteed to not be anything listeing on this.  Userlandproxy which is used by docker to do the rewrites IIUC needs to bind a port.  So unless you have a firewall rewriting rule for some random ports (at which point I would balk at your setup) this is safe on linux with docker running on localhost.

armfergom: Thanks @jtnord . Makes sense.

abayer:
Closed and re-opened to rerun the build.

abayer:
:bee: assuming build passes.

jtnord:
:bee: (waiting fro an ATH build to pass - are you mad :-) )

armfergom:
@reviewbybees done

olivergondza:
Please resolve the conflicts so we can merge.

armfergom:
Resolved conflicts after @jtnord [PR](https://github.com/jenkinsci/acceptance-test-harness/pull/84) was merged.

@reviewbybees

andresrc:
:bee:

jtnord:
new code should really have a (c) header - but :bee:

armfergom:
@reviewbybees done 

...again

PR #83, opened 2016-03-08
[JENKINS-33382] MSBuild acceptance tests.

armfergom:
There was no test coverage for MSBuild. This PR includes a test suite for MSBuild plugin that covers the basic use cases. It has to be run on Windows with MSBuild installed, otherwise the test is ignored.

- [x] MSBuild plugin test suite.
- [x] Ability to copyFiles on Windows. (only Unix machines were supported). If RemoteJenkinsController is in use, error is thrown (_for now_, needs to be fixed)
- [x] New annotation `@WithOs`, that checks whether the test is being executed on the right OS. Otherwise, tests are skipped.
- [x] `@Native` annotation modified as discussed in the PR

@olivergondza Could I please have your thoughts here? There's room for improvement, part of the set up could be extracted to an annotation like `@WithDocker`. Any other piece of advice is welcome.

@reviewbybees


src/test/java/plugins/MSBuildPluginTest.java
```diff
@@ -0,0 +1,133 @@
...
+            return job;
+        } else {
+            throw new IllegalArgumentException("Workspace and buildFile must be different from null.");
+        }
+    }
+    
+    @Before
+    public void setUp() {
+         //Check environment configuration
+         if (System.getenv(MSBUILD_EXE_ENV) == null) {
```
olivergondza: `@TestActivation` should be used instead to skip Jenkins startup.

armfergom: I missed that annotation, that's exactly what I needed. Thanks for pointing this out. Also, I think I am going to create a new annotation to enforce specific OS (line 71)

src/test/java/plugins/MSBuildPluginTest.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    @Test
+    public void buildProjNoCmdLineArgumentsTest() {
+        FreeStyleJob job = msBuildJob("/msbuild_plugin/projProject/", "project.proj", null);
+
+        // Job should run successfully
+        Build b = job.scheduleBuild();
+        b.shouldSucceed();
+        assertThat(b.getConsole(), Matchers.containsString("Build succeeded."));
```
olivergondza: `build.shouldSucceed();`

src/test/java/plugins/MSBuildPluginTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+    @Test
+    public void buildProjCmdLineArgumentsTest() {
+        FreeStyleJob job = msBuildJob("/msbuild_plugin/projProject/", "project.proj", "/verbosity:detailed");
+
+        // Job should run successfully
+        Build b = job.scheduleBuild();
+        b.shouldSucceed();
+        String console = b.getConsole();
+        assertThat(console, Matchers.containsString("Build succeeded."));
```
olivergondza: Ditto

src/test/java/plugins/MSBuildPluginTest.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+    
+    @Test
+    public void buildSlnNoCmdLineArgumentsTest() {
+        FreeStyleJob job = msBuildJob("/msbuild_plugin/slnProject/", "project.sln", null);
+        
+        // Job should run successfully
+        Build b = job.scheduleBuild();
+        b.shouldSucceed();
+        assertThat(b.getConsole(), Matchers.containsString("Build succeeded."));
```
olivergondza: Ditto

src/test/java/plugins/MSBuildPluginTest.java
olivergondza: Actually, is this Jenkins output of output of the tool?

olivergondza:
Looks good aside from that.

src/test/java/plugins/MSBuildPluginTest.java
armfergom: With that string I am checking the output of the tool, not the output of Jenkins.

src/test/java/plugins/MSBuildPluginTest.java
armfergom: @olivergondza I am trying to use the annotation you suggested but I am seeing something strange.. From [this](https://github.com/jenkinsci/acceptance-test-harness/blob/b27c150f35a386dfced4a5960127bf62f9c34363/src/main/java/org/jenkinsci/test/acceptance/junit/TestActivation.java#L82), it seems like the environment variable that needs to be present must look like `{className.envVarName}`. I see two problems here:

- Unix does not allow dots in environment variables. See [here](http://unix.stackexchange.com/questions/93532/exporting-a-variable-with-dot-in-it#answer-93533)
- How can tests like [this](https://github.com/jenkinsci/acceptance-test-harness/blob/b27c150f35a386dfced4a5960127bf62f9c34363/src/test/java/core/JdkTest.java) be working. First, we have the issue with the dot. Second, look at line [15](https://github.com/jenkinsci/acceptance-test-harness/blob/b27c150f35a386dfced4a5960127bf62f9c34363/src/test/java/core/JdkTest.java#L15), the variable that tries to get does not include the `className.`before the environment variable name. So, for this test, there has to be two environment variables set. One with the class name (to pass the activation) and one without the dot (so that the value can be retrieved). And in Unix, the one with the dot cannot be set.

Any thoughts? Am I missing something?

olivergondza: Yeah, The JdkTest seems broken. The good thing is that nobody seems to be running it as nobody noticed so it should be safe to change the envvar named to the form we require.

Note that it is bash that does not handle dots in envvar names, not unix. The delimiter was brought in by KK who I presume is using different shell. Anyway, this can be avoided by replacing `f.o=42 b.r=43 mvn ...` with `env f.o=42 b.r=43 mvn`.

armfergom: I am afraid I had already tried that and it does not work. I wonder why this is being restricted this way. It just makes it more difficult to run tests and can cause problems depending on the CI machine that runs them. 

Currently, all tests using `@TestActivation`are being ignored on the CI server. See [1](https://jenkins.ci.cloudbees.com/job/core/job/acceptance-test-harness/1630/testReport/plugins/BeakerBuilderPluginTest/), [2](https://jenkins.ci.cloudbees.com/job/core/job/acceptance-test-harness/1630/testReport/plugins/OpenstackCloudPluginTest/), [3](https://jenkins.ci.cloudbees.com/job/core/job/acceptance-test-harness/1630/testReport/plugins.jclouds/JCloudsPluginTest/) and [4](https://jenkins.ci.cloudbees.com/job/core/job/acceptance-test-harness/1630/testReport/core/JdkTest/). But this is probably because the needed [environment variables](https://jenkins.ci.cloudbees.com/job/core/job/acceptance-test-harness/1630/injectedEnvVars/) are not set. What I don't know is whether they are not set because nobody did it, or because nobody was able to.

I think it would make sense to change the `.` for a `_` and I am happy to do it. I would also update the only 4 plugin tests that use this annotation. What do you think?

scoheb: I too have struggled with the '.' issue in env vars for Test Activation.

+1 to change it to '_'

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
```diff
@@ -188,10 +197,14 @@ public void copyFile(File file) {
             ZipUtil.pack(file, tmp);
             byte[] archive = IOUtils.toByteArray(new FileInputStream(tmp));
 
-            addShellStep(String.format(
-                    "base64 --decode << ENDOFFILE > archive.zip && unzip -o archive.zip \n%s\nENDOFFILE",
-                    new String(Base64.encodeBase64Chunked(archive))
-            ));
+            if (SystemUtils.IS_OS_WINDOWS) {
+                addBatchStep("xcopy " + file.getAbsolutePath() + " %cd% /E");
```
jglick: :bug: IIUC this will not work at all for a `RemoteJenkinsController`—hence the complex trick for Unix. If you do not want to implement it correctly for Windows now, OK, but make sure this method throws a meaningful error when a remote controller is in use.

src/test/java/plugins/MSBuildPluginTest.java
armfergom: @olivergondza @scoheb I've been assuming that `@TestActivation` is used to set environment variables (as I initially wanted) but it turns out that it is not used for that but to set [system properties](https://github.com/jenkinsci/acceptance-test-harness/blob/b27c150f35a386dfced4a5960127bf62f9c34363/src/main/java/org/jenkinsci/test/acceptance/junit/TestActivation.java#L83) that do allow `.`. It is not quite the functionality I was looking for but it will do... So really, nothing to do here. 

If needed, another annotation for environment variables could be added.

armfergom:
@jglick @olivergondza comments addressed. I have added a new annotation that checks that the test is being run on the right OS to avoid starting up Jenkins otherwise (`@WithOS`)

src/main/java/org/jenkinsci/test/acceptance/junit/WithOS.java
```diff
@@ -0,0 +1,84 @@
...
+    
+    OS[] os();
+
+    public class RuleImpl implements TestRule {
+        @Override
+        public Statement apply(final Statement base, final Description d) {
+           return new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                Class<?> testCase = d.getTestClass();
```
jglick: Generally in JUnit this would be the “suite”, as opposed to the method which would be the “case”.

src/main/java/org/jenkinsci/test/acceptance/junit/WithOS.java
```diff
@@ -0,0 +1,84 @@
...
+            }
+
+            private void check(WithOS withos, Class<?> testCase) {
+                if (withos == null) return;
+                
+                String errorMsg = "Test must be run on a machine running any of the following operating systems: " + Arrays.toString(withos.os());
+                for (OS _os : withos.os()) {
+                    switch (_os) {
+                    case LINUX:
+                        if (!SystemUtils.IS_OS_LINUX) {
```
jglick: Probably not easy to fix, so should at least be documented: this is checking the OS of the machine _running the test_, which may not be the same as the OS of the machine _running Jenkins_. If you are using `LocalController` of course there is no distinction, but for other controllers there certainly is.

src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
```diff
@@ -43,6 +44,12 @@
     @Inject
     public Injector injector;
 
+    /**
+     * The controller that starts/stops Jenkins
+     */
+    @Inject
+    public JenkinsController controller;
```
jglick: Better to put this directly in `Job` where it is used.

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
```diff
@@ -188,10 +199,18 @@ public void copyFile(File file) {
             ZipUtil.pack(file, tmp);
             byte[] archive = IOUtils.toByteArray(new FileInputStream(tmp));
 
-            addShellStep(String.format(
-                    "base64 --decode << ENDOFFILE > archive.zip && unzip -o archive.zip \n%s\nENDOFFILE",
-                    new String(Base64.encodeBase64Chunked(archive))
-            ));
+            if (SystemUtils.IS_OS_WINDOWS) {
+                if (controller instanceof RemoteJenkinsController) {
```
jglick: Rather you want to check that it is _not_ `instanceof LocalController`: there are several kinds of controller that would not work with this trick.

src/test/java/plugins/MSBuildPluginTest.java
```diff
@@ -0,0 +1,133 @@
...
+ *      - MSBUild.exe is not present in the specified location
+ */
+@WithPlugins("msbuild")
+@TestActivation({"MSBUILD_EXE"})
+@WithOS(os = {WithOS.OS.WINDOWS})
+public class MSBuildPluginTest extends AbstractJUnitTest {
+
+    @Inject
+    @Named("MSBuildPluginTest.MSBUILD_EXE") 
+    private String MS_BUILD_EXECUTABLE;
```
jglick: Why not just mark the test suite

```java
@Native("msbuild")
```

and that would avoid the need for `@WithOS`?

src/main/java/org/jenkinsci/test/acceptance/junit/WithOS.java
armfergom: Javadoc and error message modified to reflect this.

src/main/java/org/jenkinsci/test/acceptance/junit/WithOS.java
armfergom: Changed variable name.

src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
armfergom: My idea was to put it in the top level element so that it could be used everywhere if needed. Moved it to `Job` anyway.

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
armfergom: Good point. Fixed.

src/test/java/plugins/MSBuildPluginTest.java
armfergom: IIUC, what `@Native` would do is execute `which msbuild` that will always fail in both Unix and Windows machines. First case because `msbuild` does not exist on Unix machines, in the second case because `which` does not exist on Windows machines.

armfergom:
@jglick Pushed changes and replied to your comments. Cheers.

andresrc:
:bee:

jglick:
> My idea was to put it in the top level element so that it could be used everywhere if needed.

If that was your idea, then you should have _deleted_ the corresponding field in existing subclasses.

src/test/java/plugins/MSBuildPluginTest.java
jglick: Well obviously you would need to fix the annotation handler to work on Windows as well, I suppose to search `System.getenv("Path")` directly. (Probably by searching `getenv("PATH")` you could write a single implementation which would work on both Unix and Windows.)

src/main/java/org/jenkinsci/test/acceptance/junit/WithOS.java
```diff
@@ -0,0 +1,89 @@
...
+ * 
+ * Indicates that the test and Jenkins instance must be running 
+ * on one of the operating systems provided. If this condition is not
+ * met, this test will be skipped.
+ * 
+ * <p>
+ * This check can only be <strong>trusted to do its job correctly</strong> when tests 
+ * are being run with a LocalController. When using a different controller, 
+ * the machine running Jenkins can be different from the one running the tests so
+ * the test may pass or fail when it should not.
```
jglick: Yet you are not actually issuing a runtime warning when this is the case, which you probably [could](https://github.com/jenkinsci/acceptance-test-harness/blob/6681f400d9a84ded4c39e8a8c4c982c1a2196295/src/main/java/org/jenkinsci/test/acceptance/junit/JenkinsAcceptanceTestRule.java#L169) [be](https://github.com/jenkinsci/acceptance-test-harness/blob/6681f400d9a84ded4c39e8a8c4c982c1a2196295/src/main/java/org/jenkinsci/test/acceptance/junit/WithPlugins.java#L73-L77).

src/test/java/plugins/MSBuildPluginTest.java
armfergom: IMHO and IIUC we face here the same issue as in `@WithOS`. We would be searching the PATH env variable for the machine running the tests, not _necessarily_ the machine running Jenkins. Actually, same concern is present ATM for the `@Native` annotation.

Just to be sure I get this right anyway, the idea here would be to scan all directories present in the path looking for files named as the value passed to the `@Native` annotation (i.e "msbuild" or "java", correct?  

I am happy to implement the solution you think it's best given your expertise on the matter (plus I will need your bee at some point :)), so could you please let me know what you prefer (`@WithOS` or tweaking `@Native` annotation)? and then I will proceed with the implementation.

andresrc: In that case we may be getting into the `LocalController` constraint again.

src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
```diff
@@ -9,6 +9,7 @@
 import java.util.concurrent.TimeUnit;
 
 import org.hamcrest.StringDescription;
+import org.jenkinsci.test.acceptance.controller.JenkinsController;
```
amuniz: Unused?

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
```diff
@@ -188,10 +209,18 @@ public void copyFile(File file) {
...
-            addShellStep(String.format(
-                    "base64 --decode << ENDOFFILE > archive.zip && unzip -o archive.zip \n%s\nENDOFFILE",
-                    new String(Base64.encodeBase64Chunked(archive))
-            ));
+            if (SystemUtils.IS_OS_WINDOWS) {
+                if (!(controller instanceof LocalController)) {
+                    // TODO: Make it work for RemoteJenkinsController like in Unix (below)
+                    throw new AssumptionViolatedException("Copying files in Windows is only supported if a LocalController is in use. Test will be skipped.");
+                }
+                addBatchStep("xcopy " + file.getAbsolutePath() + " %cd% /E");
```
amuniz: As per [this commit](https://github.com/armfergom/acceptance-test-harness/commit/4264b16e09412febd736bd185ee95a084d7d5ea0) I'd say that the implementation for *NIX systems (code below) was done such weird in order to avoid issues with special characters in the path.

`xcopy " + file.getAbsolutePath() + " %cd% /E"` seems more weak to me, but yes, it covers 99% of use cases, so it's ok IMO.

amuniz:
LGTM :+1:
Are you planning to do some more fixes/improvements? If not, :bee: for me.

src/test/java/plugins/MSBuildPluginTest.java
jglick: > We would be searching the PATH env variable for the machine running the tests, not necessarily the machine running Jenkins

True. The annotation does not currently seem to differentiate whether you expect to run the executable locally or from Jenkins, though I think the expectation in most cases is that it is something run from Jenkins. Could be checked by running a Groovy closure remotely.

I do not want to distract too much from the PR at hand, which is useful, just pointing out areas that need to be strengthened in the harness generally.

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
jglick: > I'd say that the implementation for *NIX systems (code below) was done such weird in order to avoid issues with special characters in the path.

@amuniz No, the filename has nothing to do with this. The problem is that with a remote Jenkins controller you cannot simply copy a file from a known path into the workspace, because the source is on the machine running the test code. So the Unix trick is to take the bytestream of the file, Base-64 encode it, and then create a command which _when run on any machine_ will cause that exact binary file to reappear in the current directory (in this case a Jenkins workspace). Whether there is a (portable) Windows equivalent to this trick, I have no idiom.

src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
armfergom: Yes. Just a remainder of a previous change that I reverted. Will fix.

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
```diff
@@ -188,10 +209,18 @@ public void copyFile(File file) {
...
             byte[] archive = IOUtils.toByteArray(new FileInputStream(tmp));
 
-            addShellStep(String.format(
-                    "base64 --decode << ENDOFFILE > archive.zip && unzip -o archive.zip \n%s\nENDOFFILE",
-                    new String(Base64.encodeBase64Chunked(archive))
-            ));
+            if (SystemUtils.IS_OS_WINDOWS) {
+                if (!(controller instanceof LocalController)) {
+                    // TODO: Make it work for RemoteJenkinsController like in Unix (below)
+                    throw new AssumptionViolatedException("Copying files in Windows is only supported if a LocalController is in use. Test will be skipped.");
```
jtnord: not true.  you can have base64 on windows and the test should run if this is the case.

olivergondza:
As I see how elaborate it is to make this work on win, I am tempted to incorporate the `path-element-plugin` into this repo and turn in into a ATH helper plugin. Sometimes, it would be so much easier to implement the thing in Java as `BuildStep` or `ComputerLauncher`.

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
armfergom: @jtnord I personally don't know right now how to implement the same functionality for Windows. base64 is the only thing you have natively on Windows AFAIK. What about unzip and [this](http://stackoverflow.com/questions/2500436/how-does-cat-eof-work-in-bash#answer-2500451)? If you can point me to the right direction, I would definitely like to add full support for copying files on Windows.

jtnord: http://cygwin.redhat.com  you also need docker and a load of other stuff to run the tests on windows...
see PR#84

What I am saying is that whilst you may not have these tools installed - if they are installed the test should not be skipped. 
Currently I beleive this would be working for me on my setup (I don't think it is the cause of any of the still failing tests that I am seeing at least - although still a number to go through)

armfergom: The scope of #84 is to make tests work on Windows.

The scope of my PR is to implement a test suite that _has_ to be executed on a Windows machine (to test MSBuild). 

If I remove the LocalController check, and a non LocalController is used and tests are being run on Windows, this piece of code will not do what is expected and we would start getting non informative error messages. So, I am not sure what you are suggesting to do here. Can you explain?

armfergom:
Pushed some new changes.

I have improved `@Native` so that it can be used in both win and *nix machines following @jglick suggestion (run other tests using the annotation and worked ok). Since I have already implemented it, I believe we could keep the `@WithOS` annotation although the `MSBuildPluginTest` is not using it anymore.

 `MSBuildPluginTest` is now using the `@Native` annotation only and it is much cleaner than before, which is good. 

I think there is still one concern raised by @jtnord for which I am waiting feedback. Not sure yet if it will involve making more changes.

@amuniz @andresrc when the whole discussion comes to an end, I will probably need some re-bees here. Will let you know.

Thanks.

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
jglick: Since (this part of) this PR is strictly adding functionality, I think we can approve it as is, and use separate PRs to expand the coverage to more combinations such as Windows + remote controller, or to different OSs locally vs. remotely.

jglick:
> I am tempted to incorporate the path-element-plugin into this repo and turn in into a ATH helper plugin.

Actually I think that the functionality of `path-element` should simply be baked into core. It is harmless (AFAIK) and useful for many purposes, including `JenkinsRule.WebClient` and possibly even non-test clients.

armfergom:
@jglick do you have any other suggestions/comments for this PR? If not, could you unblock the review? Thanks!

src/main/java/org/jenkinsci/test/acceptance/junit/Native.java
```diff
@@ -45,9 +63,31 @@ public void evaluate() throws Throwable {
...
+                    if (!(controller instanceof LocalController)) {
+                        throw new AssumptionViolatedException("Test skipped. Native should be used with a local controller, otherwise it cannot be fully trusted to work as expected.");
+                    }
+                    
+                    List<String> supportedCmds = Lists.newLinkedList();
+                    // Get PATH. Valid for both Unix and Windows
+                    String path = System.getenv("PATH");
+                    if (path != null) {
+                        // Get all directories in PATH
+                        String[] dirsInPath = SystemUtils.IS_OS_WINDOWS ? path.split(";") : path.split(":");
```
jglick: `File.pathSeparator` you mean?

src/main/java/org/jenkinsci/test/acceptance/junit/Native.java
```diff
@@ -45,9 +63,31 @@ public void evaluate() throws Throwable {
...
+                            if (dir.isDirectory()) {
+                                supportedCmds.addAll(Arrays.asList(dir.list()));
+                            }
+                        }
+                    }
+                    // An cmd is not Native if it cannot be found in the PATH
                     for (String cmd : n.value()) {
-                        if (new CommandBuilder("which",cmd).system()!=0) {
-                            throw new AssumptionViolatedException(cmd + " is needed for the test but doesn't exist in the system");
+                        if (!supportedCmds.contains(cmd)) {
```
jglick: On Windows, we want to try adding `.bat` and `.exe` extensions to `cmd`.

src/main/java/org/jenkinsci/test/acceptance/junit/Native.java
```diff
@@ -45,9 +63,31 @@ public void evaluate() throws Throwable {
...
+                    String path = System.getenv("PATH");
+                    if (path != null) {
+                        // Get all directories in PATH
+                        String[] dirsInPath = SystemUtils.IS_OS_WINDOWS ? path.split(";") : path.split(":");
+                        
+                        // Iterate over PATH directories to get all supported commands
+                        for (String dirName : dirsInPath) {
+                            File dir = new File(dirName);
+                            if (dir.isDirectory()) {
+                                supportedCmds.addAll(Arrays.asList(dir.list()));
```
jglick: Theoretically possible NPE here.

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
armfergom: @jglick The bug was here.

src/test/java/plugins/MSBuildPluginTest.java
```diff
@@ -0,0 +1,94 @@
...
+import org.jenkinsci.test.acceptance.msbuild.MSBuildStep;
+import org.jenkinsci.test.acceptance.po.Build;
+import org.jenkinsci.test.acceptance.po.FreeStyleJob;
+import org.junit.Test;
+
+/**
+ * This test is designed to be run on a Windows machine with MSBuild installed and in the PATH
+ * 
+ * Tests will be skipped if:
+ *      - MSBUild.exe is in the path. This implies that machine running Jenkins is windows.
```
jglick: You mean, it will be skipped if that is _not_ in the path.

jglick:
:bee:

src/main/java/org/jenkinsci/test/acceptance/junit/Native.java
```diff
@@ -45,9 +63,34 @@ public void evaluate() throws Throwable {
...
                     if (n==null)        return;
+                    
+                    // Checks performed in this annotation only makes sense if Jenkins and Tests are being executed in the same machine
+                    if (!(controller instanceof LocalController)) {
+                        throw new AssumptionViolatedException("Test skipped. Native should be used with a local controller, otherwise it cannot be fully trusted to work as expected.");
+                    }
+                    
+                    List<String> supportedCmds = Lists.newLinkedList();
+                    // Get PATH. Valid for both Unix and Windows
+                    String path = System.getenv("PATH");
```
jtnord: No - just revert this hunk.  I have a much better fix in #84 [`Native.java#47-62`](https://github.com/jtnord/acceptance-test-harness/blob/jenkins-33520/src/main/java/org/jenkinsci/test/acceptance/junit/Native.java#L47-L62)

src/main/java/org/jenkinsci/test/acceptance/junit/Native.java
```diff
@@ -45,9 +63,34 @@ public void evaluate() throws Throwable {
...
+                                    supportedCmds.addAll(Arrays.asList(filesInDir));
+                                }
+                            }
+                        }
+                    }
+                    // An cmd is not Native if it cannot be found in the PATH
                     for (String cmd : n.value()) {
-                        if (new CommandBuilder("which",cmd).system()!=0) {
-                            throw new AssumptionViolatedException(cmd + " is needed for the test but doesn't exist in the system");
+                        if (!supportedCmds.contains(cmd) && !supportedCmds.contains(cmd + ".bat") && !supportedCmds.contains(cmd + ".exe")) {
```
jtnord: `.cmd` `.com` and then there is  `.vbs` `.msf`........

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
jtnord: `powershell` is everyone's new best friend.

src/main/java/org/jenkinsci/test/acceptance/junit/Native.java
```diff
@@ -45,9 +63,34 @@ public void evaluate() throws Throwable {
 
                 private void verifyNativeCommandPresent(Native n) throws IOException, InterruptedException {
                     if (n==null)        return;
+                    
+                    // Checks performed in this annotation only makes sense if Jenkins and Tests are being executed in the same machine
+                    if (!(controller instanceof LocalController)) {
+                        throw new AssumptionViolatedException("Test skipped. Native should be used with a local controller, otherwise it cannot be fully trusted to work as expected.");
```
jtnord: I'm not sure that because you don't know something is true that you should treat it as false here in this case.

We do not run the tests like this in the CI system - so this would mainly hit local users, and I or others may want to run the tests against a `RemoteJenkinsController` so I can run Unix based tests.  Now I have no way to do that, before it could do this by virtue of creating a dummy file in my path, if it had not picked up something from cygwin to begin with...

src/test/java/plugins/MSBuildPluginTest.java
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * This test is designed to be run on a Windows machine with MSBuild installed and in the PATH
+ * 
+ * Tests will be skipped if:
+ *      - MSBUild.exe is not in the path. This implies that machine running Jenkins is windows.
+ *      - The test and Jenkins are being run on different machines
+ */
+@WithPlugins("msbuild")
+@Native({"MSBuild"})
+public class MSBuildPluginTest extends AbstractJUnitTest {
```
jtnord: do you not want this to be `withOS(Windows)`?

src/test/java/plugins/MSBuildPluginTest.java
```diff
@@ -0,0 +1,94 @@
...
+    @Test
+    public void buildSlnCmdLineArgumentsTest() {
+        FreeStyleJob job = msBuildJob("/msbuild_plugin/slnProject/", "project.sln", "/verbosity:detailed");
+        
+        // Job should run successfully
+        Build b = job.scheduleBuild();
+        b.shouldSucceed();
+        String console = b.getConsole();
+        assertThat(console, Matchers.containsString("Build succeeded."));
+        assertThat(console, Matchers.containsString("Done Building Project"));
```
jtnord: btw - allOf(...) is a good thing when asserting multiple matchers against the same object.

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
jtnord: Well arguably it breaks existing functionality if you are running on windows have cygwin and using a `RemoteJenkinsController`

what you give in one hand you take away with the other :-)

jtnord:
:bug: on introducing `@WithOs` and not using it for the test that apparently needs it.

I don't think the implementation of searching the PATH is a good one, as it is inefficient at the file system level and memory level - it also does not search for tools correctly - as it is ignoring PATHEXT which sets what extensions the command interpreter will launch.
but if you merge that I can rip it all out later :-) (but if I merge first then it's up to you to not undo it!)

Otherwise OK.

src/main/java/org/jenkinsci/test/acceptance/junit/Native.java
jglick: > before it could do this by virtue of creating a dummy file in my path

This seems like a use case we need not support. If you care, fix `Native` to actually query the remote system.

jtnord:
BTW - not to far into the future (now?) this needs to work on Linux as well.
[msbuild porting to .NET core](https://blogs.msdn.microsoft.com/dotnet/2016/02/23/porting-msbuild-to-net-core/)

src/main/java/org/jenkinsci/test/acceptance/junit/Native.java
jtnord: So I was going to do that - but in order to fix it properly you need to have Jenkins running - the point of Native was to avoid the overhead of starting Jenkins.

However I am fine if we decide that reliably detecting tools on the SUT is better than waving our arms in the air and doing nothing more than guessing.

armfergom:
@jtnord `@WithOs` was initially used. With the use of `@Native` later on it was not really needed anymore but was already implemented. I see no harm on having it for whoever wants to use it. And I see no harm in use it together with `@Native` too if that's what it takes to get the bug away..

Anyway, I'll just leave this on hold until [this](https://github.com/jenkinsci/acceptance-test-harness/pull/84) is merged and then I'll adapt this to it. It's pointless to keep working on the same things in parallel...

jtnord:
:bee: based on feedback

armfergom:
Resolved conflicts after @jtnord [PR](https://github.com/jenkinsci/acceptance-test-harness/pull/84) was merged.

Use `@WithOS` in addition of `@Native`

@reviewbybees

src/main/java/org/jenkinsci/test/acceptance/po/BatchBuildStep.java
```diff
@@ -0,0 +1,21 @@
+package org.jenkinsci.test.acceptance.po;
```
jtnord: Use [atchCommandBuildStep.java](https://github.com/jenkinsci/acceptance-test-harness/blob/2f526cec2f6d268734a2bdc914d5cbf539d0830c/src/main/java/org/jenkinsci/test/acceptance/po/BatchCommandBuildStep.java)

src/test/java/plugins/MSBuildPluginTest.java
```diff
@@ -0,0 +1,96 @@
...
+    
+    @Test
+    public void buildProjNoCmdLineArgumentsTest() {
+        FreeStyleJob job = msBuildJob("/msbuild_plugin/projProject/", "project.proj", null);
+
+        // Job should run successfully
+        Build b = job.scheduleBuild();
+        b.shouldSucceed();
+        assertThat(b.getConsole(), Matchers.containsString("Build succeeded."));
+        assertThat(b.getConsole(), Matchers.not(Matchers.containsString("Done building project \"project.proj\".")));
```
jtnord: it's generally better to import static on Hamcrest matchers so it reads better  `not(containsString("foobar)))`

Is this actually the right test - running msbuild without arguments [will search](https://msdn.microsoft.com/en-us/library/ms164311.aspx) for the .proj to run in the CWD.

jtnord:
:bug: on the now duplicate `BatchBuildStep.java` some minro NITs otherwise LGTM

armfergom:
@jtnord Addressed comments and removed `BatchBuildStep.java`. I was using `BatchCommandBuildStep.java` though, but forgot to remove my own version. 

Also, added a test to check that msbuild works without specifying the .proj file. Run it and passed.

andresrc:
:bee:

jtnord:
:bee:

armfergom:
@reviewbybees done

jtnord:
@andresrc seems to have caused a regression
```
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /scratch/jenkins/workspace/core/acceptance-test-harness-stable/src/main/java/org/jenkinsci/test/acceptance/msbuild/MSBuildInstallation.java:[11,15] incompatible types: org.jenkinsci.test.acceptance.po.JenkinsConfig cannot be converted to org.jenkinsci.test.acceptance.po.Jenkins
[INFO] 1 error
[INFO] -------------------------------------------------------------
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
```

armfergom:
@jtnord I noticed. Creating a PR for this.

The problem is that there was another [PR](https://github.com/jenkinsci/acceptance-test-harness/pull/89/) merged 11 days ago, that modified the `ToolInstallation` class from which the `MSBuildToolInstallation` class extends. This PR was opened like a month ago so the PR branch did not contain that. The final reviewbybees was done 12 days ago, before the problematic PR was merged.

PR #86, opened 2016-03-23
[JENKINS-33741] [SSH Slaves Plugin] Improve test coverage 

armfergom:
Adding more test cases for SSH Slaves Plugin.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_slaves/SshSlaveLauncher.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_slaves/SshSlaveLauncher.java
@@ -1,6 +1,8 @@
 package org.jenkinsci.test.acceptance.plugins.ssh_slaves;
 
+import org.jenkinsci.test.acceptance.plugins.credentials.UserPwdCredential;
 import org.jenkinsci.test.acceptance.plugins.ssh_credentials.SshCredentialDialog;
+import org.jenkinsci.test.acceptance.plugins.ssh_credentials.SshPrivateKeyCredential;
 import org.jenkinsci.test.acceptance.po.ComputerLauncher;
 import org.jenkinsci.test.acceptance.po.Control;
 import org.jenkinsci.test.acceptance.po.Describable;
@@ -13,7 +15,13 @@ import org.jenkinsci.test.acceptance.po.PageObject;
 public class SshSlaveLauncher extends ComputerLauncher {
     public final Control host = control("host");
     public final Control credentialsId = control("credentialsId");
-
+    public final Control prefixCmd = control("prefixStartSlaveCmd");
+    public final Control suffixCmd = control("suffixStartSlaveCmd");
+    public final Control timeout = control("launchTimeoutSeconds");
+    public final Control retries = control("maxNumRetries");
+    public final Control javaPath = control("javaPath");
+    public final Control jvmOptions = control("jvmOptions");
+        
     public SshSlaveLauncher(PageObject context, String path) {
         super(context, path);
     }
@@ -32,5 +40,37 @@ public class SshSlaveLauncher extends ComputerLauncher {
 
     private void ensureAdvancedOpen() {
         control("advanced-button").click();
+    }
+    
+    /**
+     * Add username/password based credentials to the configuration
+     * 
+     * @param username to use
+     * @param password for the username
+     * @return the SshSlaveLauncher to be configured
+     */
+    public SshSlaveLauncher pwdCredentials(String username, String password) {
+        SshCredentialDialog dia = this.addCredential();
+        UserPwdCredential cred = dia.select(UserPwdCredential.class);
+        cred.username.set(username);
+        cred.password.set(password);
+        cred.add();
+        return this;
+    }
+
+    /**
+     * Add username/key based credentials to the configuration
+     * 
+     * @param username to use
+     * @param key for the private key to use
+     * @return the SshSlaveLauncher to be configured
+     */
+    public SshSlaveLauncher keyCredentials(String username, String key) {
+        SshCredentialDialog dia = this.addCredential();
+        SshPrivateKeyCredential cred = dia.select(SshPrivateKeyCredential.class);
+        cred.username.set(username);
+        cred.enterDirectly(key);
+        cred.add();
+        return this;
     }
 }
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_slaves/SshSlaveLauncher.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_slaves/SshSlaveLauncher.java
@@ -25,8 +25,6 @@ public class SshSlaveLauncher extends ComputerLauncher {
     public final Control javaPath = control("javaPath");
     public final Control jvmOptions = control("jvmOptions");
         
-    private String credentialUsername;
-    
     public SshSlaveLauncher(PageObject context, String path) {
         super(context, path);
     }
@@ -60,7 +58,7 @@ public class SshSlaveLauncher extends ComputerLauncher {
         cred.username.set(username);
         cred.password.set(password);
         cred.add();
-        credentialUsername = username;
+        waitForCredentialVisible(username);
         return this;
     }
 
@@ -77,7 +75,7 @@ public class SshSlaveLauncher extends ComputerLauncher {
         cred.username.set(username);
         cred.enterDirectly(key);
         cred.add();
-        credentialUsername = username;
+        waitForCredentialVisible(username);
         return this;
     }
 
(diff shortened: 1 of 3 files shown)


src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/JavaContainer/Dockerfile
```diff
@@ -18,5 +18,8 @@ ENTRYPOINT ["/usr/sbin/sshd", "-D", "-e"]
 
 # JDK is from Universe
 RUN echo deb http://archive.ubuntu.com/ubuntu precise universe >> /etc/apt/sources.list
+RUN apt-get install --no-install-recommends -y software-properties-common
+RUN add-apt-repository ppa:openjdk-r/ppa
 RUN apt-get update
+RUN apt-get install --no-install-recommends -y openjdk-8-jdk
```
jtnord: it is better to run all installations in a single line as it produces less intermediary images.

src/test/java/plugins/SshSlavesPluginTest.java
```diff
@@ -99,4 +129,20 @@ private void verify() {
         job.save();
         job.startBuild().shouldSucceed();
     }
+    
+    private void verifyLog(String message) {
+        elasticSleep(3000);
```
jtnord: despite its name elasticSleep is not and this will pause all tests for 3 seconds.
Better to use waitFor(... 3) which will wait 3 seconds for the log to contain the matcher that you want, but in the fast case will return as soon as the data is written.

jtnord: or 
```
slave.waitUntilOnline();
assertTrue("Slave did not come online within default timeout", slave.isOnline());
assertTrue(slave.getLog().contains(message));
```

olivergondza:
LGTM, once James' comments are incorporated.

armfergom:
@jtnord comments addressed. There are two test cases where I cannot wait until slave is online because what is being tested is that it cannot be connected. So, in that case there is still a "elastic" sleep in place so that the connection attempt has enough time to fail.

There's another sleep [here](https://github.com/jenkinsci/acceptance-test-harness/pull/86/files#diff-0e999aca3636d0fbfa2ac895e5d5cd78R67) as well to avoid test flakiness. Slave is not able to connect with this error in the logs if we don't wait some time before saving the slave configuration:

```
hudson.AbortException: Cannot find SSH User credentials with id: null
	at hudson.plugins.sshslaves.SSHLauncher.openConnection(SSHLauncher.java:1210)
	at hudson.plugins.sshslaves.SSHLauncher$2.call(SSHLauncher.java:711)
	at hudson.plugins.sshslaves.SSHLauncher$2.call(SSHLauncher.java:706)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

If anybody has any other suggestion to avoid this, it will be welcome. Another workaround is to re-save the slave configuration. In that case, it works too. So it seems to be a problem when saving the configuration "too quickly".

jtnord:
> There's another sleep here as well to avoid test flakiness

Sounds like a valid failure being detected (ie a race condition in core / plugin)?

src/test/java/plugins/SshSlavesPluginTest.java
```diff
@@ -58,35 +55,74 @@
...
-        launcher.port(sshd.port(22));
-
-        SshCredentialDialog dia = launcher.addCredential();
-        SshPrivateKeyCredential cred = dia.select(SshPrivateKeyCredential.class);
-        cred.username.set("test");
-        cred.enterDirectly(sshd.getPrivateKeyString());
-        cred.add();
+        configureDefaultSSHSlaveLauncher().keyCredentials("test", sshd.getPrivateKeyString());
+        // To avoid test flakiness
+        elasticSleep(1000);
```
jtnord: would appear like a valid error scenario in the underlying SUT?

src/test/java/plugins/SshSlavesPluginTest.java
```diff
@@ -58,35 +55,74 @@
...
 
         verify();
     }
 
+    @Test public void unableToConnectWrongPort() {
+        configureSSHSlaveLauncher(sshd.ipBound(22), 1234).pwdCredentials("test", "test");
+        slave.save();
+        
+        // Wait for connection attempt to fail
+        elasticSleep(3000);
```
jtnord: No need to sleep - just waitUntil the log contains...  with a timeout,

src/test/java/plugins/SshSlavesPluginTest.java
```diff
@@ -58,35 +55,74 @@
...
+    }
+    
+    
+    @Test public void unableToConnectWrongCredentials() {
+        configureDefaultSSHSlaveLauncher().pwdCredentials("unexsisting", "unexsisting");
+        slave.save();
+        
+        // Wait for connection attempt to fail
+        elasticSleep(3000);
+        verifyLog("Authentication failed");
```
jtnord: No need to sleep - just waitUntil the log contains... with a timeout,

jtnord:
something inside me says this should be inside a separate plugin which depends on the AWS stuff and the backup plugin.

"As a user who does not use AWS I should not have more fluff in the UI than I need (credential types, extra AWS configuration)."  Adding the dependency here will pull in these plugins which will add their configuration items.

src/test/java/plugins/SshSlavesPluginTest.java
armfergom: @jtnord Without waiting some time to save the configuration this is what I get in the slave configuration xml:

```
...
  <launcher class="hudson.plugins.sshslaves.SSHLauncher" plugin="ssh-slaves@1.10">
    <host>192.168.99.100</host>
    <port>32859</port>
    <maxNumRetries>0</maxNumRetries>
    <retryWaitTime>0</retryWaitTime>
  </launcher>
...
```

If I then re-save the configuration, or if I wait for some time before actually saving the slave configuration, I get this instead:

```
...
  <launcher class="hudson.plugins.sshslaves.SSHLauncher" plugin="ssh-slaves@1.10">
    <host>192.168.99.100</host>
    <port>32859</port>
    <credentialsId>c2605d23-3e92-4c64-907d-5d90f2cdfd0f</credentialsId>
    <maxNumRetries>0</maxNumRetries>
    <retryWaitTime>0</retryWaitTime>
  </launcher>
...
```

And that's happening because the DOM is not generated quick enough and a `null` is sent for the credentials field. I don't see an actual issue here, just a test that needs to take this into account and has to wait until the dropdown has been filled with the credential that has just been created. So that is what I have done instead of sleeping for 1 second which, I agree, is not very elegant nor reliable. If you see it differently, please do let me know.

Other comments are also addressed.

jtnord: Hmmm, I don't see why this path is different to using password credentials (which works without a sleep)?

The DOM would be empty if the UI has not populated the dropdown yet with the newly created item - so I get that but why it's only in this case is jumping out at me. 

If you do need to wait waiting for the element (credentials) to be visible would be a better option as that can shave time off the test (everything adds up!)

jtnord:
:bee:

andresrc:
:bee:

armfergom:
@reviewbybees done

src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_slaves/SshSlaveLauncher.java
```diff
@@ -71,6 +77,20 @@ public SshSlaveLauncher keyCredentials(String username, String key) {
...
         cred.add();
+        credentialUsername = username;
         return this;
     }
+
+    /**
+     * Once a credential has been created for a given slave, this method can be used 
+     * to check whether it has already been rendered in the dropdown.
+     */
+    public void waitForCredentialVisible() {
```
olivergondza: Note that when it will not contain the text (but the control would exist) you will never know as the return value of `until` is ignored. I am not sure what are trying to do here (why are you waiting for visual feedback) but you should probably assert it contains the text.

armfergom: See discussion [here](https://github.com/jenkinsci/acceptance-test-harness/pull/86#diff-for-comment-57709373) (this same pull request). 

@olivergondza Summing up, it happens that the credentials are added and immediately after the slave config is saved without time for the credential to appear in the combobox, resulting in a slave configuration with `null` credentials. With this, I ensure the credentials are already there and the slave configuration is created with them.

olivergondza: I see, I prefer to make this part of the method that sets the value so the caller do not have to worry about this problem at all.

armfergom: @olivergondza yeah, makes complete sense. Changed and pushed. Cheers

PR #89, opened 2016-03-31
Abstract away the fact tools config page was moved away from global config

olivergondza:
@tfennelly, it is still a work in progress


src/test/java/plugins/GroovyPluginTest.java
```diff
@@ -109,11 +111,10 @@ public void use_custom_groovy_version() {
...
     public void use_native_groovy() {
-        jenkins.configure();
-        GroovyInstallation groovy = jenkins.getConfigPage().addTool(GroovyInstallation.class);
-        groovy.name.set("local-groovy");
-        groovy.useNative();
-        jenkins.save();
+        GroovyInstallation ant = ToolInstallation.addTool(jenkins, GroovyInstallation.class);
+        ant.name.set("local-groovy");
+        ant.useNative();
+        ant.getPage().save();
```
olivergondza: :)

PR #90, opened 2016-04-04
[JENKINS-33994] Improve ATH test coverage for Credentials Plugin

armfergom:
Improve ATH test coverage for Credentials Plugin

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/credentials/Domain.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/credentials/Domain.java
@@ -9,8 +9,8 @@ import org.jenkinsci.test.acceptance.po.PageObject;
 @Describable("Credential Domain")
 public class Domain extends PageAreaImpl {
 
-    public Control name = control("/domain/name");
-    public Control description = control("/domain/description");
+    public final Control name = control("/domain/name");
+    public final Control description = control("/domain/description");
     public final Control addCredentialButton = control("/hetero-list-add[credentials]");
     public final Control addSpecificationButton = control("/domain/hetero-list-add[specifications]");
     
@@ -21,16 +21,9 @@ public class Domain extends PageAreaImpl {
     public Domain(PageArea area, String relativePath) {
         super(area, relativePath);
     }
-
-    /**
-     * Add a domain
-     */
-    public void add() {
-        find(by.button("Add Domain")).click();
-    }
     
     /**
-     * Adds a new credential and bind it to the page ae object.
+     * Adds a new credential under the scope of this domain
      */
     public <T extends Credential> T addCredential(Class<T> type) {
         addCredentialButton.selectDropdownMenu(type);
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/credentials/Domain.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/credentials/Domain.java
@@ -1,0 +1,42 @@
+package org.jenkinsci.test.acceptance.plugins.credentials;
+
+import org.jenkinsci.test.acceptance.po.Control;
+import org.jenkinsci.test.acceptance.po.Describable;
+import org.jenkinsci.test.acceptance.po.PageArea;
+import org.jenkinsci.test.acceptance.po.PageAreaImpl;
+import org.jenkinsci.test.acceptance.po.PageObject;
+
+@Describable("Credential Domain")
+public class Domain extends PageAreaImpl {
+
+    public Control name = control("/domain/name");
+    public Control description = control("/domain/description");
+    public final Control addCredentialButton = control("/hetero-list-add[credentials]");
+    public final Control addSpecificationButton = control("/domain/hetero-list-add[specifications]");
+    
+    public Domain(PageObject context, String path) {
+        super(context, path);
+    }
+    
+    public Domain(PageArea area, String relativePath) {
+        super(area, relativePath);
+    }
+
+    /**
+     * Add a domain
+     */
+    public void add() {
+        find(by.button("Add Domain")).click();
+    }
+    
+    /**
+     * Adds a new credential and bind it to the page ae object.
+     */
+    public <T extends Credential> T addCredential(Class<T> type) {
+        addCredentialButton.selectDropdownMenu(type);
+
+        String path = last(by.xpath("//div[@name='credentials']")).getAttribute("path");
+
+        return newInstance(type, this, path);
+    }
+}
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/credentials/Domain.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/credentials/Domain.java
@@ -2,9 +2,7 @@ package org.jenkinsci.test.acceptance.plugins.credentials;
 
 import org.jenkinsci.test.acceptance.po.Control;
 import org.jenkinsci.test.acceptance.po.Describable;
-import org.jenkinsci.test.acceptance.po.PageArea;
 import org.jenkinsci.test.acceptance.po.PageAreaImpl;
-import org.jenkinsci.test.acceptance.po.PageObject;
 
 @Describable("Credential Domain")
 public class Domain extends PageAreaImpl {
@@ -14,12 +12,8 @@ public class Domain extends PageAreaImpl {
     public final Control addCredentialButton = control("/hetero-list-add[credentials]");
     public final Control addSpecificationButton = control("/domain/hetero-list-add[specifications]");
     
-    public Domain(PageObject context, String path) {
+    public Domain(ManagedCredentials context, String path) {
         super(context, path);
-    }
-    
-    public Domain(PageArea area, String relativePath) {
-        super(area, relativePath);
     }
     
     /**
@@ -28,7 +22,7 @@ public class Domain extends PageAreaImpl {
     public <T extends Credential> T addCredential(Class<T> type) {
         addCredentialButton.selectDropdownMenu(type);
 
-        String path = last(by.xpath("//div[@name='credentials']")).getAttribute("path");
+        String path = last(by.name("credentials")).getAttribute("path");
 
         return newInstance(type, this, path);
     }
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/test/acceptance/plugins/credentials/Domain.java
```diff
@@ -0,0 +1,35 @@
...
+    public final Control name = control("/domain/name");
+    public final Control description = control("/domain/description");
+    public final Control addCredentialButton = control("/hetero-list-add[credentials]");
+    public final Control addSpecificationButton = control("/domain/hetero-list-add[specifications]");
+    
+    public Domain(PageObject context, String path) {
+        super(context, path);
+    }
+    
+    public Domain(PageArea area, String relativePath) {
```
olivergondza: If I am not mistaken, the `Domain` page area is scoped to `ManagedCredentials` page object only. In that case it should have only one constructor: `Domain(ManagedCredentials parent, String relativePath)`.

src/test/java/core/CredentialsTest.java
```diff
@@ -1,55 +1,227 @@
...
+        upc.scope.select(systemScope);
+        upc.setId(CRED_ID);
+        c.save();
+
+        // verify credential was created
+        verifyInputValue(CRED_ID, USERNAME_INPUT, CRED_USER);
+        
+        // Update credential
+        c.open();
+        upc = c.get(UserPwdCredential.class, CRED_ID);
```
olivergondza: Is it really needed to look up the `PageArea` like that (the implementation of the `get`methods sound fragile to me)? Why not reuse the instance you already have? The fact you navigated away and back should not matter.

src/test/java/core/CredentialsTest.java
```diff
@@ -1,55 +1,227 @@
...
+       try {
+           // Find domain div
+            findIfNotVisible(by.input(domainName))
+           .findElement(by.ancestor("div"))
+            // Find credential div inside domain div
+           .findElement(by.input(credId))
+           .findElement(by.ancestor("div"))
+           // Find input with credential id passed as parameter
+           .findElement(by.input(credId));
+       } catch (NoSuchElementException e) {
```
olivergondza: This will generate false positives as soon as surrounding markup changes. Making sure the name/id is not present on the whole page anywhere is safer.

I wish there ware more pragmatic way to verify the credentials are set correctly (IOW, actually use it not just verify it is the the form), but all seems to be too complicated.

src/test/java/core/CredentialsTest.java
armfergom: Well, I don't see the test would be very useful if I just reuse the instance I created. The idea is to create the instance, navigate to the page again (reload it), get the instance, update it, and then perform the checks. Why do you think the implementation is fragile?

src/test/java/core/CredentialsTest.java
armfergom: My first idea was to actually use it. I discarded that solution because it would involve requiring another plugin that uses credentials for that and I did not want to add another variable to the test case, that is, adding some other failure point not directly related with the credentials plugin. I could do it though, for example, require the Git plugin and when creating a FreeStyleProject, checking that the credentials created appear in the credentials dropdown. What do you think?

src/test/java/core/CredentialsTest.java
```diff
@@ -1,55 +1,227 @@
...
+        c.save();
+        
+        // verify credential is not present
+        verifyCredentialNotPresent(domainName, CRED_ID);
+    }
+
+    /**
+     * Verifies that a given input for a given credential id 
+     * is equals to the value expected. If there are two credentials
+     * with the same id (that should not happen), the first one is verified.
```
olivergondza: Is there a reason not to use the (properly nested) `PageArea` to access the value? There is no need to rely on the fact it will be the first element found, no need to depend on current markup and no need to duplicate what the `PageArea` is doing already.

src/test/java/core/CredentialsTest.java
```diff
@@ -1,55 +1,227 @@
...
+     */
+    private void verifyCredentialNotPresent(String credId) {
+        assertThat(getElement(by.input(credId)), nullValue());        
+    }
+    
+    /**
+     * Verifies that a given input for a given credential id 
+     * is equals to the value expected under the scope of a given domain.
+     * If there are two credentials with the same id in the same domain,
+     * the first one is verified.
```
olivergondza: Ditto.

src/main/java/org/jenkinsci/test/acceptance/plugins/credentials/Domain.java
armfergom: Sounds rights :) Will fix it once I get some more input from you in the other comments in case we thing there are more things to improve/add. Cheers

src/test/java/core/CredentialsTest.java
olivergondza: Agreed, too complicated.

src/test/java/core/CredentialsTest.java
armfergom: Well, basically the same reason mentioned in another comment. I believe that for the test to be meaningful, the page should be revisited and the value checked. Also, it is basically what was being done before my changes. See [here](https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/test/java/core/CredentialsTest.java#L34)

src/test/java/core/CredentialsTest.java
olivergondza: > Why do you think the implementation is fragile?

They depend on surrounding markup that might change and they duplicate the functionality as noted in other comment.

> Well, I don't see the test would be very useful if I just reuse the instance I created. The idea is to create the instance, navigate to the page again (reload it), get the instance, update it, and then perform the checks.

Note that `PageObject`s/`PageArea`s are supposed to be mare handles to the page. What matters is that you have reloaded the page, getting new *handle* should not be necessary.

armfergom: @olivergondza I was typing when I got your second comment :). Actually, I agree. Get method is useless here. Can (and will) remove it. 

As for the `verify` methods, I am not so sure. I agree that relying on the markup is not ideal, but it seems to me inevitable in many cases when implementing acceptance tests. What, I think, proves this, is that lots of tests are going to have to be fixed with Jenkins 2.0 because of changes to the markup, so I think this is just inherent to the nature of these tests. 

Also, when saving the credentials, the configuration page is exited. The idea behind the `verify` methods is to visit the page again and actually check that the inputs contains the values expected, as you would do if you were a real user basically. 

As for [this](https://github.com/jenkinsci/acceptance-test-harness/pull/90/files#diff-30e169630196f1c675dc380646da7e5aR210), I could remove the exception check and just check for the id in the whole page. But I could get false positives here as well. If I create a credential inside of a domain with username `bob` and another credential outside of the domain with username `bob` as well, and then I checked for the username in the whole page, I would be checking for the first occurrence of `usernmae==bob` and that could not be the global one (not the one inside the domain), so I would not be checking what I want. It's up to the test implementation to reproduce this, but I just wanted to be sure that this would never happen. I don't see a big issue in just being safer (markup wise) and just check for the occurrence in the whole page as suggested though, just wanted to explain the logic behind it.

So, what do you think?

olivergondza: > As for the verify methods, I am not so sure. I agree that relying on the markup is not ideal, but it seems to me inevitable in many cases when implementing acceptance tests. What, I think, proves this, is that lots of tests are going to have to be fixed with Jenkins 2.0 because of changes to the markup, so I think this is just inherent to the nature of these tests. 

Even if it ware, there is no reason to make things worse. (@tfennelly might want to  correct me here, though most of the 2.0 effort addresses problems on higher level: unbundled plugins not installed, security on by default, whole parts of forms moved, text labels changed.)

> Also, when saving the credentials, the configuration page is exited. The idea behind the verify methods is to visit the page again and actually check that the inputs contains the values expected,

Agreed, the `PageArea` will do the heavy lifting for you so you do not have to locate the input manually.

> and ideally we want to check a concrete input, in the same div as the credential id.

Divs and inputs are implementation details the test should known nothing about on the first place. You want to make sure there is a *credential* scoped to some *domain* for certain *user* with given *configuration*, you do not care in how many `div`s is any of that nested.

> As for this, I could remove the exception check and just check for the id in the whole page. But I could get false positives here as well. If I create a credential inside of a domain with username bob and another credential outside of the domain with username bob as well, and then I checked for the username in the whole page, I would be checking for the first occurrence of usernmae==bob and that could not be the global one (not the one inside the domain), so I would not be checking what I want. It's up to the test implementation to reproduce this, but I just wanted to be sure that this would never happen. This actually applies to all verifies that are being performed, not just the one for the method stated.

A problem that will be fixed by referring to the `PageArea` as there will be no ambiguity.

armfergom:
@olivergondza addressed your comments. Thanks

armfergom:
@olivergondza addressed your comments. Thanks

armfergom:
Accidentally closed the PR. Reopening

olivergondza:
LGTM, though some of the failures are related: https://jenkins.ci.cloudbees.com/job/core/job/acceptance-test-harness-stable/664/console

armfergom:
Apparently the `Advanced...` button has been removed in 1.651.1-SNAPSHOT which is the SUT.

**Screenshot from 1.642.3 (Jenkins version I was developing with):**

screen shot 2016-04-05 at 12 38 09
**Screenshot from 1.651.1-SNAPSHOT (SUT):**

screen shot 2016-04-05 at 12 37 50
This is why it's failing in the CI server. The only field under advanced was the `id` which is what is failing and is not really used anywhere. So I am removing it.

andresrc:
:bee:

armfergom:
Pushed a change to stop setting the credential id, since in SUT it is not there yet. Could have also bumped the minimum required plugin version, but I think it's better to test whatever is currently bundled with Jenkins.

armfergom:
Ok, now they are passing as expected: https://jenkins.ci.cloudbees.com/job/core/job/acceptance-test-harness-stable/666/console

olivergondza:
Thanks

PR #91, opened 2016-04-04
Fix element waits

olivergondza:
- Prefer elastic waits
- Refactor some explicit waiting loops to use selenium API.


olivergondza:
This seems broken.

PR #93, opened 2016-04-05
[JENKINS-34004] Add identity configuration when creating repo.

andresrc:
See [JENKINS-34004](https://issues.jenkins-ci.org/browse/JENKINS-34004).

Adds a local user and email to the created repos. This makes the tests pass even if the global config is not set.

@reviewbybees


amuniz:
:bee:

> in case the global config is not set

It seems to be set unconditionally, no?

andresrc:
> It seems to be set unconditionally, no?

yes, to make the tests self contained. The values are not significant for the tests.

armfergom:
🐝

andresrc:
@reviewbybees done

PR #95, opened 2016-04-15
[JENKINS-34285] Script and signature approval tests.

andresrc:
See [JENKINS-34285](https://issues.jenkins-ci.org/browse/JENKINS-34285)

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/groovy_postbuild/GroovyPostBuildStep.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/groovy_postbuild/GroovyPostBuildStep.java
@@ -1,0 +1,70 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/test/acceptance/plugins/groovy_postbuild/GroovyPostBuildStep.java
```diff
@@ -0,0 +1,70 @@
...
+import org.jenkinsci.test.acceptance.po.Control;
+import org.jenkinsci.test.acceptance.po.Describable;
+import org.jenkinsci.test.acceptance.po.Job;
+import org.jenkinsci.test.acceptance.po.PostBuildStep;
+
+/**
+ * Groovy post build step. Use {@code form-element-path} plugin.
+ * Assumes "path" form element availability.
+ * Paths are usually available when using the default Jenkins controller.
+ * (Available thanks to pre-installed form-element-path plugin.)
```
olivergondza: All steps does. I do not think that the testsuite is of much of a vale without the plugin.

src/test/java/plugins/ScriptSecurityPluginTest.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    private void shouldSucceed(FreeStyleJob job) {
+        login(USER);
+        job.scheduleBuild().shouldSucceed();
+    }
+
+    @Test
+    public void scriptNeedsApproval() throws Exception {
+        final FreeStyleJob job = createFailedJob("def a = 4", false);
```
rsandell: why would this fail??

andresrc: because it does not use the sandbox, so the script requires approval.

rsandell:
:bee:

src/main/java/org/jenkinsci/test/acceptance/plugins/script_security/PendingSignature.java
```diff
@@ -0,0 +1,57 @@
...
+public class PendingSignature extends CapybaraPortingLayerImpl {
+    private final WebElement block;
+
+    PendingSignature(Injector injector, WebElement block) {
+        super(injector);
+        this.block = block;
+    }
+
+    private void click(String buttonName) {
+        block.findElement(by.button(buttonName)).click();
```
jglick: What if there are multiple approvals pending? This API will not work.

jglick: I see, `findSignature` does that lookup.

jglick:
retro:bee:

PR #96, opened 2016-04-21
Fix toggle in JUnit build results.

andresrc:
@reviewbybees


recena:
🐝  since I've been talking with @andresrc about it.

olivergondza:
This reflect a change in a plugin and fix like this is going to break the tests for the old version. Given the cosmetic change that broke this, we should imho rewrite the assert to visit the test page and get the message from there to be less fragile.

andresrc:
@olivergondza agree, will fix, thanks!

recena:
@olivergondza I agree, we can use another selector less fragile. Other option would be to set the minimum version of JUnit where the change was done.

armfergom:
LGTM 🐝

abayer:
:bee:, merging.

PR #98, opened 2016-05-04
MultipleScmsPluginTest still referred to svn.jenkins-ci.org. Fixed.

abayer:
cc @reviewbybees


jglick:
Safe until Ant’s SVN repo is dropped I guess. :bee:

abayer:
Good point. I got a better option. =)

abayer:
Updated to use SVN's own SVN repo, because if we can't count on that... cc @reviewbybees

PR #99, opened 2016-05-05
PageObject for MsTestRunner build step

armfergom:
To encourage the implementation of tests for this plugin.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/mstestrunner/MSTestRunnerBuildStep.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/mstestrunner/MSTestRunnerBuildStep.java
@@ -1,3 +1,26 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 CloudBees
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/test/acceptance/plugins/mstestrunner/MSTestRunnerBuildStep.java
```diff
@@ -0,0 +1,36 @@
...
+    private final Control categoriesControl = control("categories");
+    private final Control resultFileControl = control("resultFile");
+    private final Control cmdLineArgsControl = control("cmdLineArgs");
+    private final Control continueOnFailControl = control("continueOnFail");
+    
+    public MSTestRunnerBuildStep(Job parent, String path) {
+        super(parent, path);
+    }
+
+    public void configure(String testFiles, String resultFile, String categories, String cmdLineArgs, boolean ignoreFailingTests) {
```
olivergondza: Please split this into separate methods or just expose the `Controls` so it can be configured independently (provided there is no reason to always set these in bulk).

armfergom: Done. I have just exposed the controls. I have left the method because I still see it useful for whoever wants to use it. For those test cases where all controls are needed, it'd be cleaner to have one line of code rather than 5, one for each control.

andresrc:
:bee:

rsandell:
:nit: you should add a MIT header to the new file.

armfergom:
@rsandell added MIT header. Not all new files have it though.

PR #100, opened 2016-05-06
Fix ACE editor detection in workflow projects.

andresrc:
* Fix timeout figure.
* Support old editor `id`.


armfergom:
🐝 good catch

PR #103, opened 2016-05-18
apt-get update alone can lead to Docker caching the results.

abayer:
Run another command on the same line to make sure that doesn't happen.

We're hitting this on an internal test run where there's a stale apt cache on an image and this is the best workaround I can find.

See https://github.com/docker/docker/issues/3313 

cc @reviewbybees


andresrc:
:bee:

It is the recommended practice btw,

https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#apt-get

jtnord:
yeah why not, sigh... :bee: although I must confess I don't understand how this works...
the result of `apt-get update && apt-get install -y unzip` is still one command as far as docker is concerned and will result in an image that can be cached...

abayer:
*shrug* I have no idea, but it works this way. =)

abayer:
@reviewbybees done

PR #105, opened 2016-05-20
JIRA image now needs JDK8.

abayer:
cc @reviewbybees


abayer:
Note that the test still fails even with this, but it at least gets further!

PR #106, opened 2016-05-23
PageArea/PageObjects for plain-credentials and credentials-binding

armfergom:
`PageArea`s and `PageObject`s for `plain-credentials` and `credentials-binding` plugins so that acceptance tests can be written for them.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/credentialsbinding/CredentialsBinding.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/credentialsbinding/CredentialsBinding.java
@@ -1,0 +1,32 @@
+package org.jenkinsci.test.acceptance.plugins.credentialsbinding;
+
+import org.jenkinsci.test.acceptance.po.ContainerPageObject;
+import org.jenkinsci.test.acceptance.po.Control;
+import org.jenkinsci.test.acceptance.po.PageArea;
+import org.jenkinsci.test.acceptance.po.PageAreaImpl;
+
+/**
+ * Represents a credential binding in the job configuration page.
+ */
+public class CredentialsBinding extends PageAreaImpl {
+
+    public Control credentialId = control("credentialId");
+    public Control variable = control("variable");
+    
+    public CredentialsBinding(PageArea area, String path) {
+        super(area, path);
+    }
+    
+    public CredentialsBinding(ContainerPageObject po, String path) {
+        super(po, path);
+    }
+    
+    /**
+     * Checks whether there are credentials in the credentials drop down
+     * 
+     * @return true if there are no credentials, false otherwise
+     */
+    public boolean noCredentials() {
+        return getElement(by.tagName("options")) == null;
+    }
+}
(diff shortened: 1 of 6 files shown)


src/main/java/org/jenkinsci/test/acceptance/plugins/credentialsbinding/CredentialsBinding.java
```diff
@@ -0,0 +1,32 @@
...
+        super(po, path);
+    }
+    
+    /**
+     * Checks whether there are credentials in the credentials drop down
+     * 
+     * @return true if there are no credentials, false otherwise
+     */
+    public boolean noCredentials() {
+        return getElement(by.tagName("options")) == null;
```
olivergondza: This is fragile as it will report any `select`/`options` tag in the whole page. Btw, I presume you meant `option`, not `options`.

armfergom: oops, you are right, I meant `option`, that's a typo. And you are right, it is looking for `<option>` tags in the whole page. I did not intend to do that, I wanted to look for it in the PageArea context. Will fix. Thanks

olivergondza:
As I just realized there is a lot of UI changes between credentials v1 and v2. To make sure new credentials code works with both, I intend to postpone merging this until there are tests that use this (and prove it really does what it seems to).

armfergom:
@olivergondza existing tests are also in danger... This morning I filed [JENKINS-35090 ](https://issues.jenkins-ci.org/browse/JENKINS-35090) because of the new credentials versions.

There are many tests that use `ManagedCredentials` class to create credentials and that class is only compatible with v1 unfortunately. I am working on making it compatible with v2 as well, although I am not sure if both version lines should be supported. WDYT?

Problem is that `credentials-plugin` is bundled up until Jenkins 2.0 and the ATH does not upgrade the plugin to the latest version unless it is a dependency of another required plugin that is not installed.

For example, if `credentials-binding` (not bundled) plugin is required with `@WithPlugins`, even though the `credentials-plugin` is installed, it gets upgraded to the latest version because it is a dependency of `credentials-binding` (which is not installed). For that reason, such tests using `ManagedCredentials` class to add new credentials **will fail.** 

On the other hand, if `@WithPlugins` required the `ssh-credentials` plugin, since `ssh-credentials` is bundled, it will not be installed and `credentials-plugin` would not be upgraded. In this case, `ManageCredentials` **will work** because the bundled version is older than 2.x.

Making `ManagedCredentials` work for both version lines may be hacky. Should both lines be supported or only the latest?

By the way, I believe this PR is not affected by the newer versions of `credentials-plugin`

armfergom:
@olivergondza I have updated the PR to address your comment.

stephenc:
Once I finish the shake-down of credentials 2.0 (currently I will deem that as 2 weeks from the last release with no S1/S2 regressions... so that is two weeks from today thanks to the bug that is fixed in 2.0.3) then I will be recommending the credentials 2.0 line for everything as this gives users access to the critical domain and store selection functionality that was missing for a long time.

Up to you to decide what you actually want to do with the ATH

olivergondza:
> Making ManagedCredentials work for both version lines may be hacky. Should both lines be supported or only the latest?

I will definitely need the old version support for a while. Everybody who runs the tests against older versions of the plugin (I presume cloudbees does that).

armfergom:
@olivergondza This PR is not affected by the recent changes of credentials plugins. This only adds page objects to manage the credentials bindings in the job configuration and that has not changed. Can we get it merged?

andresrc:
:bee:

armfergom:
@reviewbybees done

PR #111, opened 2016-06-02
[JENKINS-35275] Don't force unneccessary dependencies to get upgraded in ATH

armfergom:
[JENKINS-35275](https://issues.jenkins-ci.org/browse/JENKINS-35275)

The proposed fix tries to make the ATH upgrade policy the same as the one in real Jenkins. See ticket for better explanation.

@olivergondza your input here would be much appreciated. Thanks.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
@@ -154,7 +154,7 @@ public class PluginManager extends ContainerPageObject {
             if (!someChangeRequired) {
                 return false;
             }
-            List<PluginMetadata> pluginToBeInstalled = ucmd.get().transitiveDependenciesOf(jenkins.getVersion(), candidates);
+            List<PluginMetadata> pluginToBeInstalled = ucmd.get().transitiveDependenciesOf(jenkins, candidates);
             for (PluginMetadata newPlugin : pluginToBeInstalled) {
                 final String name = newPlugin.getName();
                 String claimedVersion = candidates.get(name);
--- a/src/main/java/org/jenkinsci/test/acceptance/update_center/UpdateCenterMetadata.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/update_center/UpdateCenterMetadata.java
@@ -15,6 +15,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.jenkinsci.test.acceptance.po.Jenkins;
+
 /**
  * Databinding for Update Center metadata
  *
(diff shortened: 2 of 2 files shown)


olivergondza:
:+1:

abayer:
🐛 - still seeing failures on GitPluginTest#checkout_local_branch, e.g., on both 1.651.2 and 2.7 - it's still installing Credentials 2.0.7 -
```
12:09:20 INFO: credentials has version 1.18 but 2.0.7 was requested
12:09:20 Plugin /root/.m2/repository/org/jenkins-ci/plugins/credentials/2.0.7/credentials-2.0.7.hpi installed
```

armfergom:
Credentials version is 1.18 because whatever Jenkins you are using bundles it. The plugin under test (I don't know which one it is) has surely a dependency to a version of credentials greater than 1.18, so credentials plugin needs to be updated. The behaviour in a real Jenkins would be to upgrade to latest version of credentials and so it should be in the ATH.

This fix is about not upgrading the plugin when it is not needed, that is if the dependency of the plugin under test was credentials 1.18 or older.

abayer:
Any chance we could have it upgrade to the required version rather than latest in that scenario?

abayer:
So https://github.com/abayer/acceptance-test-harness/commit/67f8e4bad97c79c5f529513c85f3afb7294d1127 ensures that it only uses the highest explicit dependency version of a plugin, if it's specified as a dependency somewhere - before that, it'd end up installing the newest version of the plugin any time it wasn't already installed, including if you explicitly specified the plugin version in `@WithPlugins` *after* a plugin that depended on the first one.

That said, I'm not sure this is a good idea. It's great at preventing a dependency plugin from breaking tests for other plugins, but as a result, it masks problems caused by that newer dependency version that might not be tested except in those other plugins' tests.

armfergom:
@abayer I think this PR is about making the ATH to behave as similarly as possible to a real Jenkins instance. If a required dependency is not installed in a real Jenkins, the latest version would be installed. If an installed dependency is valid it does not get updated, that's all.

That said, when specifying a version number in `@WithPlugins` like `@WithPlugins("credentials@1.25")`, what you are doing is telling the ATH that, if the installed versions of credentials **is older than 1.25**, upgrade it to the **latest**. If the installed dependency of credentials is **1.25 or greater**, it matches the condition and it does not get upgraded. 

Now, being able to set a specific version of a plugin to test with might be useful but I think that is something that should be configurable in `@WithPlugins` annotation or _elsewhere_ but we should not make it the default behaviour.

recena:
:bee: and 👍

oleg-nenashev:
Looks reasonable to me 👍 
BTW it would be great to run the tests in the "bleeding edge" mode by specifying a proper flag

recena:
@oleg-nenashev There are some ideas around that. I hope to get good news soon.

abayer:
:bee:

armfergom:
@reviewbybees done

jglick:
As noted in #118, I would rather just have ATH simulate updating all plugins to the latest available version, and remove the version option from `@WithPlugins`.

armfergom:
We would not be able to test a given Jenkins instance as it is then, right? It could cause problems as some I have seen where credentials plugins get updated and ATH starts failing (because of major UI changes) while the real instance of Jenkins works as expected with the plugin versions that are bundled.

How about having a system property that tells the ATH to always update to latest versions available for the baseline under test and leave as default the actual behaviour in a real Jenkins?

olivergondza:
Note that the version in `@WithPlugins` (and `@Since`) is the _minimal_ version, not required one. I guess we will have to collect use-cases and create several strategies. Right now it is not straightforward how to configure ATH to do what you want, and as it gets more complex I am less and less confident it really tests what I want it to. Some strategies we might be interested in:

- Always use latest version of plugins (For CI build to make sure latest Jenkins and plugins works together).
- Never touch plugins provided (For testing particular plugin collection (including plugin versions)).
- Do not update plugin with sufficient version (Acts like Jenkins)
- ...

armfergom:
👍 for configurable strategies for ATH

armfergom:
@olivergondza , is there any other feedback I should be addressing or is this PR already valid for merge?

armfergom:
@olivergondza are there any concerns about something related to this PR or needs some extra work?

PR #114, opened 2016-06-02
[JENKINS-35134] Ensure no focus lib is used for Firefox.

abayer:
Fixes a ton of job creation failures on Linux - doesn't make a
difference on OS X, and there are still a number of failures mentioned
in the JIRA that are credentials-related, but this is real progress.

cc @reviewbybees @recena @uhafner @olivergondza


andresrc:
:bee:

kwhetstone:
:bee:

PR #118, opened 2016-06-06
[JENKINS-34687] Fix XML comparison for 2.x lines

armfergom:
This [commit](https://github.com/stapler/stapler/commit/17f9e496ce24365befecf69251e46169e25924af) in Stapler modified the XML that the Jenkins Remote API produces introducing a new attribute called `_class`. The first version of Stapler including the change was `1.241`. This change was never introduced in Jenkins 1.x line but it is now in Jenkins 2.x line since Stapler was bumped to version `1.243`. 

This has caused some ATH tests to start failing agains Jenkins 2.x. This PR fixes these tests by ignoring the attributes differences. 

@olivergondza WDYT?

@reviewbybees esp. @abayer


abayer:
:bee: We might be getting to the point where it's worth moving the 1.x ATH to a separate branch and just having master concerned with 2.x...

jglick:
> We might be getting to the point where it's worth moving the 1.x ATH to a separate branch and just having master concerned with 2.x

I think 1.x vs. 2.x is a red herring. In general, ATH tends to break for all sorts of reasons whenever we update core or various plugins, and we have no coherent way of dealing with that.

jglick:
Perhaps ATH should just be versioned according to core LTS line, where for plugins we assume that the latest plugin release compatible with that core is the one we need to be testing.

abayer:
That's a good idea - 1.x vs 2.x is still relevant given UI changes, but yeah, LTS line does seem like an appropriate way to branch.

olivergondza:
I do not think that expecting 100% match on `api/xml` is a sane thing to rely on in test. The entities can be in different order based on collection implementation and JDK version used (7 vs 8), for example. What is the actual difference here? Is it part of the "interesting data" (from the test's point of view)?

armfergom:
This was not and is not checking that the XMLs are identical. In fact, what this is doing is checking that the two XMLs are **similar**. That's what the `true` [here](https://github.com/jenkinsci/acceptance-test-harness/pull/118/files#diff-fc3a7583cae9f42abed87b75c9786b4fR640) means.

What this PR is addressing is the fact that from Jenkins 2.x, the XMLs tags generated have an extra attribute called `_class`, and it is being excluded in the comparison so that the check is valid for both 1.x and 2.x lines.

Not sure if this addresses your comment though.

olivergondza:
> This was not and is not checking that the XMLs are identical. In fact, what this is doing is checking that the two XMLs are similar.

Right. I am fine fine with the fix as it gets the tests green again. @uhafner, can you verify that ignoring attributes (IIUC) will not make your tests toothless?

uhafner:
For me it would make more sense to have 2 different files to compare with: one for 1.x and one for 2.x. These can be selected in the individual test cases by evaluating `jenkins.getVersion()`. When LTS is finally on 2.x we can drop the old ones.

armfergom:
Yeah, I considered that option too, but I just wanted to have a single fix for the 4 tests that were failing. Plus, the method can still behave as before if `ignoreAttributesDiffs` parameter is false. 

That said, I'll implement whatever solution the community feels is better. Please, advise! :)

uhafner:
Well, since I'm the owner of the analysis plug-ins I would like to pick the second solution: this solution will retain the original behavior (and brevity) as soon as LTS is on the 2.x line.

armfergom:
@uhafner implemented the second option. I have also left my previous changes since there is a parameter (`ignoreAttributesDiffs`) in `assertXmlApiMatchesExpected` that needs to be set to `true` to ignore attributes. Left it to `false` and we now have 2 different xml files.

uhafner:
Thanks for fixing!

PR #119, opened 2016-06-06
Fixing GlobalToolConfig source location

jglick:
Fixes an apparent mistake in #89.

@reviewbybees & @olivergondza


pom.xml
```diff
@@ -441,7 +441,7 @@
       <plugin>
         <groupId>org.codehaus.gmaven</groupId>
         <artifactId>gmaven-plugin</artifactId>
-        <version>1.5</version>
+        <version>1.5-jenkins-3</version>
```
jglick: Uses a more canonical location under `target/` for generated sources, which is helpful for NetBeans at least.

jglick: Hmm, CI builder does not like this, but I am not sure how to reproduce locally.

olivergondza:
Thanks for catching this.

PR #121, opened 2016-06-08
Put Docker logs in diagnostics dir

olivergondza:
- Log `docker run` and `docker build` separately into diagnostics directory.
- Introduce builder for `DockerImage` starting to keep the signatures sane.
- As a bonus, make sure `cid` file gets deleted.

@reviewbybees esp. @jtnord

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
@@ -91,15 +91,22 @@ public class Docker {
             return new DockerImage(full);
         }
 
-        ProcessBuilder buildCmd = cmd("build").add("-t", full, dir).build();
+        CommandBuilder buildCmd = cmd("build").add("-t", full, dir);
+        ProcessBuilder processBuilder = buildCmd.build();
         if (log != null) {
-            buildCmd.redirectError(log).redirectOutput(log);
+            processBuilder.redirectError(log).redirectOutput(log);
         } else {
-            buildCmd.redirectError(ProcessBuilder.Redirect.INHERIT).redirectOutput(ProcessBuilder.Redirect.INHERIT);
+            processBuilder.redirectError(ProcessBuilder.Redirect.INHERIT).redirectOutput(ProcessBuilder.Redirect.INHERIT);
         }
 
-        if (buildCmd.start().waitFor() != 0) {
-            throw new Error("Failed to start image: " + tag);
+        StringBuilder sb = new StringBuilder("Building Docker image `").append(buildCmd.toString()).append("`");
+        if (log != null) {
+            sb.append(": logfile is at ").append(log);
+        }
+        System.out.println(sb.toString());
+
+        if (processBuilder.start().waitFor() != 0) {
+            throw new Error("Failed to build image: " + tag);
         }
         return new DockerImage(full);
     }
--- a/src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
@@ -7,6 +7,7 @@ import org.jenkinsci.test.acceptance.utils.SHA1Sum;
 import org.jenkinsci.utils.process.CommandBuilder;
 import org.jvnet.hudson.annotation_indexer.Index;
 
+import javax.annotation.CheckForNull;
 import javax.inject.Named;
 import javax.inject.Singleton;
 import java.io.File;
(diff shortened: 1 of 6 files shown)


src/main/java/org/jenkinsci/test/acceptance/junit/FailureDiagnostics.java
```diff
@@ -74,6 +76,17 @@ public void write(String filename, String content) {
         }
     }
 
+    /**
+     * Copy existing file to diagnostics directory.
+     */
+    public void archvie(String targetFilename, File source) {
```
scoheb: typo

olivergondza: Right, let's get rid of that. I ended up not needing it anyway,

src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
```diff
@@ -91,13 +91,24 @@ public DockerImage build(String image, File dir) throws IOException, Interrupted
...
-            throw new Error("Failed to build image: " + tag);
+        ProcessBuilder buildCmd = cmd("build").add("-t", full, dir).build();
+        if (log != null) {
+            buildCmd.redirectError(log).redirectOutput(log);
+        } else {
+            buildCmd.redirectError(ProcessBuilder.Redirect.INHERIT).redirectOutput(ProcessBuilder.Redirect.INHERIT);
+        }
+
+        if (buildCmd.start().waitFor() != 0) {
+            throw new Error("Failed to start image: " + tag);
```
jtnord: this is a failure to build the image not start it.

src/main/java/org/jenkinsci/test/acceptance/docker/DockerContainerHolder.java
```diff
@@ -22,15 +25,35 @@
...
+    /**
      * Lazily starts a container and returns the instance.
      */
     @Override
     public synchronized T get() {
-        if (container==null)
-            container = docker.start((Class<T>)type.getRawType());
+        if (container==null) {
+            Class<T> fixture = (Class<T>) type.getRawType();
+            File buildlog = diag.touch("docker-" + fixture.getSimpleName() + ".build.log");
```
jtnord: so this change means a failure to build an image for a test will not be in the junit log for that specific test, so you have to go looking in multiple places.
Therefore to get this log we now need to archive all these logs, and this means for the case of successful building we will not throw these logs away (like we can do with the junit trim successful tests output).

jtnord: Right this is a FailureDiagnostics, I missed that the first time around, so should not have this issue.

jtnord:
Looks ok with the exception of one incorrect log line.  I won't be able to test this on windows for ~1 week but it looks sane and not to be changing any of the docker specific bits and bobs.

src/main/java/org/jenkinsci/test/acceptance/docker/DockerContainerHolder.java
```diff
@@ -22,15 +25,35 @@
...
+            File runlog = diag.touch("docker-" + fixture.getSimpleName() + ".run.log");
+            try {
+                System.out.println("Setting up docker container for " + fixture.getSimpleName());
+                container = docker.build(fixture, buildlog).start(fixture).withPortOffset(portOffset).withLog(runlog).start();
+            } catch (InterruptedException | IOException e) {
+                throw new AssertionError("Failed to start container " + fixture.getName(), e);
```
jtnord: if the `docker.build(fixture, buildlog)` fails then it throws an Error, which would not be translated into an AssertionError.

I think this is correct as I would want the test to be in `Error` not `Failed` - so I am wondering why we treat any error at all here as `AssertionError` rather than a different type of `Error` that would cause the test to be marked as `error` rather than `failed`?

src/main/java/org/jenkinsci/test/acceptance/docker/DockerContainerHolder.java
```diff
@@ -22,15 +25,34 @@
...
-        if (container==null)
-            container = docker.start((Class<T>)type.getRawType());
+        if (container==null) {
+            Class<T> fixture = (Class<T>) type.getRawType();
+            File buildlog = diag.touch("docker-" + fixture.getSimpleName() + ".build.log");
+            File runlog = diag.touch("docker-" + fixture.getSimpleName() + ".run.log");
+            try {
+                container = docker.build(fixture, buildlog).start(fixture).withPortOffset(portOffset).withLog(runlog).start();
+            } catch (InterruptedException | IOException e) {
+                throw new AssertionError("Failed to start container " + fixture.getName(), e);
```
rsandell: Another place where you might not want to throw AssertionError?

olivergondza:
Right, will change that to `Error`.

PR #122, opened 2016-06-08
Make sure the minimal browser window is 1280x960

olivergondza:


src/main/java/org/jenkinsci/test/acceptance/FallbackConfig.java
```diff
@@ -148,9 +148,13 @@ private WebDriver createWebDriver(TestName testName) throws IOException {
     @Provides @TestScope
     public WebDriver createWebDriver(TestCleaner cleaner, TestName testName, ElasticTime time) throws IOException {
         WebDriver base = createWebDriver(testName);
+
+        // Make sue the window have minimal resolution set, even when out of the visible screen. Try maximizing first so
+        // it has a chance to fit the screen nicely if big enough.
+        base.manage().window().maximize();
```
scoheb: maximize should come after the setSize() if block

olivergondza: That is not what I intended to do. If you are watching the test or taking recording it helps if whole browser is in screen visible area. As browser can come up not maximized and might not be in upper left corner, stretching its dimensions can cause that lower right corner will be off screen (even though it could have fit in the screen if moved).

That is why it maximizes first. If the size become sufficient by doing that - great, if it does not we will stretch off the screen but we will keep as much as we can in visible area (since maximizing should ensure the browser is attached to upper left corner).

olivergondza:
@scoheb, I am merging this as it improves the test results. Let's adjust later if needed.

PR #123, opened 2016-06-08
[JENKINS-35090] - Credentials plugin upgrade/fix

kwhetstone:
Bumped the required version of the credentials plugin to 2.0.7.  Added some new classes to handle the different layout, and changed classes that were dependent on the credentials.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/junit/WithCredentials.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/junit/WithCredentials.java
@@ -3,6 +3,7 @@ package org.jenkinsci.test.acceptance.junit;
 import com.google.inject.Inject;
 import com.google.inject.Injector;
 
+import org.jenkinsci.test.acceptance.plugins.credentials.CredentialsPage;
 import org.jenkinsci.test.acceptance.plugins.credentials.ManagedCredentials;
 import org.jenkinsci.test.acceptance.plugins.credentials.UserPwdCredential;
 import org.jenkinsci.test.acceptance.plugins.ssh_credentials.SshPrivateKeyCredential;
@@ -43,6 +44,7 @@ import org.jenkinsci.test.acceptance.plugins.credentials.BaseStandardCredentials
 @Target({METHOD, TYPE})
 @Inherited
 @Documented
+@WithPlugins("credentials@2.0.7")
 @RuleAnnotation(value=WithCredentials.RuleImpl.class, priority = WithPlugins.PRIORITY + 1) // requires plugins
 public @interface WithCredentials {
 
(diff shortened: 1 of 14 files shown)


abayer:
Still seeing failures on CreateSlaveTest#newSlave...

scoheb:
PR 122 should fix that test.

The credentials inlay is not scrollable and the default resolution in VNC
is too small.

kwhetstone:
I did check out that test and it's actually not related to the rest of the credentials set.  Apparently there are a few different routes to create a credential.  To complicate things, the save/add/ok button is different there too!  If there's a fix for that default resolution, then that will be one way to solve the problem.  I think the other way is to tell Selenium to scroll on the page.

abayer:
Ah, ok. I'll throw some other tests at it tomorrow AM.

jglick:
Does this fix `WorkflowPluginTest.sshGitInsideDocker`? If so, thank you!

src/main/java/org/jenkinsci/test/acceptance/junit/WithCredentials.java
```diff
@@ -43,6 +44,7 @@
 @Target({METHOD, TYPE})
 @Inherited
 @Documented
+@WithPlugins("credentials@2.0.7")
```
jglick: Does this work on a rule? I thought it was only for actual tests. [reference](https://github.com/jenkinsci/acceptance-test-harness/blob/9e3ea2d23d825b1afd64bba5f359f763b64c4dd7/src/main/java/org/jenkinsci/test/acceptance/junit/WithPlugins.java#L92-L93)

src/main/java/org/jenkinsci/test/acceptance/plugins/credentials/BaseStandardCredentials.java
```diff
@@ -45,12 +48,13 @@ protected BaseStandardCredentials(PageArea area, String relativePath) {
...
     public void setId(String id) {
-        control("advanced-button").click();
-        control("id").set(id);
+        //control("advanced-button").click();
+        control(by.name("_.id")).set(id);
     }
     
     /**
      * Clicks the delete credential button and wait until it is not present anymore
+     * TO REMOVE - not possible with new credentials
```
jglick: Then should you not deprecate/delete this method?

abayer:
Just threw `WorkflowPluginTest#sshGitInsid​eDocker`, `GitPluginTest#checkout_branch` and `SubversionPluginTest#run_basic​_subversion_build_svn_userPwd` at this branch and they all passed - they'd been failing since the credentials change, so yay! :bee:

src/main/java/org/jenkinsci/test/acceptance/junit/WithCredentials.java
kwhetstone: That's what I thought too; but I threw the annotation on there and tried running a test without credentials included in it's list of plugins, and it installed 2.0.7 for that test.  It feels a bit off using that annotation on another annotation, but I don't expect people writing new tests using the WithCredentials rule to know to add the credentials plugin in their list of required plugins.

src/main/java/org/jenkinsci/test/acceptance/plugins/credentials/BaseStandardCredentials.java
kwhetstone: You're right, I missed deleting this with the second round of cleanups.  Removing...

src/main/java/org/jenkinsci/test/acceptance/junit/WithCredentials.java
olivergondza: Your motivation makes sense, though the reason why I and Jesse are surprised is that noone _implemented_ that to work. Unfortunately, what you are observing is an accidental behavior - ATH just installs latest version regardless of your annotation here. If you change the annotation to `@WithPlugins("credentials@3.0.7")` (version that does not exist), it installs 2.0.7 (latest) instead of failing the test.

Throwing `AssumptionViolatedException` from Statement#evaluate is the way to go.

kwhetstone: I wrapped the credentials creation with a try-catch since I figure most problems here would be related to plugin version.  The original error will still be reported for diagnosis.

abayer:
:bee: wooo

PR #124, opened 2016-06-09
Automate vnc setup for local testing

olivergondza:
ATH now takes the display for browser from `BROWSER_DISPLAY` property. The advantage compared to propagating plain old `DISPLAY` from shell exported envvar (that worked before) is it will not affect any other UI applications launched from the shell.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/FallbackConfig.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/FallbackConfig.java
@@ -42,6 +42,7 @@ import org.openqa.selenium.UnsupportedCommandException;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.chrome.ChromeDriver;
 import org.openqa.selenium.chrome.ChromeOptions;
+import org.openqa.selenium.firefox.FirefoxBinary;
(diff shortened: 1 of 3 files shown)


rsandell:
:+1:

src/main/java/org/jenkinsci/test/acceptance/FallbackConfig.java
```diff
@@ -95,7 +96,9 @@ private WebDriver createWebDriver(TestName testName) throws IOException {
 
             profile.setPreference(LANGUAGE_SELECTOR, "en");
 
-            return new FirefoxDriver(profile);
+            FirefoxBinary binary = new FirefoxBinary();
+            binary.setEnvironmentProperty("DISPLAY", getBrowserDisplay());
```
jglick: @varmenise claims this throws a `WebDriverException` on non-X11 platforms.

olivergondza: Even if the feature is not turned on?

varmenise: https://github.com/jenkinsci/acceptance-test-harness/pull/166

PR #126, opened 2016-06-15
[JENKINS-35703] Fix test 'CompressArtifactsPluginTest#archiveLargerThan4GInTotal

mafraba:
[JENKINS-35703](https://issues.jenkins-ci.org/browse/JENKINS-35703)

@reviewbybees esp @armfergom, @olivergondza

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/ArtifactManagement.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/ArtifactManagement.java
@@ -24,6 +24,7 @@
 package org.jenkinsci.test.acceptance.po;
 
 import org.openqa.selenium.NoSuchElementException;
+import org.openqa.selenium.StaleElementReferenceException;
 
 /**
  * Global page area to configure Artifact management.
@@ -45,12 +46,12 @@ public class ArtifactManagement extends PageAreaImpl {
     }
 
     public void clear() {
-        try {
-            while (true) {
+        while (control("artifactManagerFactories/repeatable-delete").exists()) {
+            try {
                 control("artifactManagerFactories/repeatable-delete").click();
+            } catch (NoSuchElementException | StaleElementReferenceException e) {
+                // control of race condition, the element might have been deleted but selenium cache not updated?
             }
-        } catch (NoSuchElementException ignored) {
-            //done no more buttons to push
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/test/java/plugins/CompressArtifactsPluginTest.java
```diff
@@ -93,12 +93,12 @@ public void archiveLargerThan4GInTotal() throws Exception {
         job.addPublisher(ArtifactArchiver.class).includes("*");
         if (SystemUtils.IS_OS_UNIX) {
             job.addBuildStep(ShellBuildStep.class).command( // Generate archive larger than 4G
-                "#!/bin/bash\nwget $JENKINS_URL/jnlpJars/jenkins-cli.jar -O stuff.jar; for i in {0..7000}; do cp -l stuff.jar stuff.${i}.jar; done"
+                "#!/bin/bash\nwget $JENKINS_URL/jnlpJars/jenkins-cli.jar -O stuff.jar; for i in {0..7000}; do cp stuff.jar stuff.${i}.jar; done"
```
mafraba: Not described in the original issue, but this also makes the test fail on Mac OS since the '-l' option is not valid on this OS.

armfergom:
If removing `-l` option does not break the test in Linux, this is a 🐝 from me

src/test/java/plugins/CompressArtifactsPluginTest.java
amuniz: `-l` creates a link instead of a full copy, but I guess it does not matter in this case.

src/main/java/org/jenkinsci/test/acceptance/po/ArtifactManagement.java
```diff
@@ -45,12 +46,12 @@ public ArtifactManagement(JenkinsConfig config) {
     }
 
     public void clear() {
-        try {
-            while (true) {
+        while (control("artifactManagerFactories/repeatable-delete").exists()) {
```
amuniz: Under what conditions the delete button does not exists?

mafraba: What the clear method does is to look for a possibly existing 'Delete' button, that is, the delete button for an item in the Artifact Management section, and clicks on it in order to remove the item. It does it iteratively until no more items (i.e. delete-buttons) are found, when the section is considered *cleared*.

amuniz:
:bee:

mafraba:
I see the test is still failing in the CI run, but with a different error that seems related to the environment setup itself?

I have successfully run the test on both Mac OS and Linux machines.

mafraba:
@reviewbybees done

abayer:
Have we verified that this works when there isn't already an artifact manager configured?

abayer:
Oh, and this also fixes `ParameterizedTriggerTest#triggerWithNonStandardArchiver` when there's a configured archiver already as well. Woo!

PR #129, opened 2016-06-15
Increase script timeout in Firefox

carlossg:
To prevent 'Unresponsive script' warnings

It would be better to allow any custom property, any suggestions?


carlossg:
I have changed it to create a WebDriverProvider and allow the injection of the FirefoxProfile

it can then be done like

```
@Extension
public class DefaultConfig extends AbstractModule implements ExtensionModule {
    @Provides
    @TestScope
    public FirefoxProfile buildFirefoxProfile() {
        FirefoxProfile profile = new FirefoxProfile();
        profile.setPreference("dom.max_script_run_time", 300);
        return profile;
    }
}
```

Vlatombe:
:bee

src/main/java/org/jenkinsci/test/acceptance/WebDriverProvider.java
```diff
@@ -0,0 +1,209 @@
...
+    private TestCleaner cleaner;
+
+    @Inject
+    private TestName testName;
+
+    @Inject
+    private ElasticTime time;
+
+    @Inject
+    private FirefoxProfile firefoxProfile;
```
olivergondza: How will this work with other browsers? Are they expected to inject their configuration holders too? It is not clear to me why we need this refactorization and the flexibility it provides.

carlossg: something similar can be done for other browsers (`ChromeOptions`, `DesiredCapabilities`,...)

Without a way to customize the FirefoxProfile there is no way for tests to change the behavior of the browser. In this particular case a script taking more than the default 30s makes the test fail.

olivergondza: I am not sure it is actually desirable to change that from test method. What is the problem this is addressing? To my understanding this change as implemented does not change the profile for any test, right?

carlossg: we are extending the test suite to add more tests, some JS takes more than the default 30 seconds and the tests fail. So we need a way to customize the Firefox profile and increase that value.
You are correct this change does not affect this project as it is right now, it only allows customizing the FF profile, which is something we need downstream but could be useful in the future to change any FF properties

olivergondza: I see. Though, I prefer to inject the actual timeout over injecting whole profile. Consider someone needs to tweak the timeout but is forced to do that for the supported browser profiles (and figure out how to do so for all the browsers). Also, there will be problem if 2 parts fo codebase wants to specify different profile attribute each, they would end up fighting over who provide the whole profile leaving other one ignored.

rsandell:
:bee: But I think I agree with Oliver.

PR #130, opened 2016-06-17
[JENKINS-36036] Fix flaky test 'ScriptSecurityPluginTest#scriptNeedsApproval'

mafraba:
Fix for [JENKINS-36036](https://issues.jenkins-ci.org/browse/JENKINS-36036)

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
@@ -24,7 +24,6 @@ import com.google.common.base.Joiner;
 import com.google.inject.Injector;
 
 import static java.util.Arrays.*;
-import org.openqa.selenium.interactions.Actions;
 
 /**
  * For assisting porting from Capybara.
--- a/src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
@@ -24,6 +24,7 @@ import com.google.common.base.Joiner;
 import com.google.inject.Injector;
 
 import static java.util.Arrays.*;
+import org.openqa.selenium.interactions.Actions;
 
 /**
  * For assisting porting from Capybara.
@@ -245,9 +246,7 @@ public class CapybaraPortingLayerImpl implements CapybaraPortingLayer {
      */
     @Override
     public void check(WebElement e) {
-        if (!e.isSelected()) {
-            e.click();
-        }
+        check(e, true);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


olivergondza:
In case horizontal scrolling is the problem, it should be fixed in https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/main/java/org/jenkinsci/test/acceptance/selenium/Scroller.java or https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/main/resources/org/jenkinsci/test/acceptance/selenium/scroller.js respectively.

mafraba:
@olivergondza thanks for pointing it out.

I wasn't aware of the existence of the 'Scroller' class. However, I see it only takes into account vertical scrolling. Looks like the support for horizontal scrolling was introduced but has been reverted later: https://github.com/jenkinsci/acceptance-test-harness/commit/1cdd0988e3175f929f784366cdb71398d916341b.

Anyway, I've been trying out some alternatives using [org.openqa.selenium.interactions.Actions#moveToElement](https://seleniumhq.github.io/selenium/docs/api/java/org/openqa/selenium/interactions/Actions.html#moveToElement-org.openqa.selenium.WebElement-) and some javascript to try to try to bring the elements into view, similarly as how the reverted commit did, and at least in this case they didn't solve the reported issue: even if the element was brought into view the 'e.click()' had no effect.

I found the proposed javascript 'click' function to be effective as a workaround, at least in this case. I can agree it's not pretty though, but can't find any more elegant solution. Anyway, if you see the code it will only come into play if the first 'selenium' click fails.

Looks like @abayer fought a similar bug. Andrew, any comments on this would be appreciated.

andresrc:
I did the original https://github.com/jenkinsci/acceptance-test-harness/commit/254d0151d26d3bf5e71e891650c2303b49b91d49

I seemed to break other test, but the thing is that horizontal scrolling is a real problem for some tests.

Do we have any complete run using this approach?

armfergom:
@andresrc the one linked to this PR is a complete run, isn't it?

armfergom:
🐝 LGTM. I think this is not intrusive and do not rely on actual scrolling. I have checked the errors in the CI server and do not seem to be related.

mafraba:
@andresrc unfortunately the CI server seems to not being behaving correctly, but as Armando indicated, this modification should only come into play when the original selenium 'click' fails. It really doesn't try to fix the wider horizontal scrolling problem, it just makes it unnecessary to scroll when ticking a checkbox.

If we have a larger set of tests that fail because of horizontal scrolling issues, and they are well known, please share the list and I can open a new issue to tackle the problem.

andresrc:
:bee:

mafraba:
@reviewbybees done

recena:
@olivergondza What do you think?

olivergondza:
I still do not like the idea of using JS for this. We should either fix the real cause - element is not visible or give up on the *visible area* business and bend selenium to permit interacting with elements out of visible area. Anyway, I am ok to have this in to fix the test.

PR #131, opened 2016-06-17
[JENKINS-36042] Fix plugins.MatrixAuthPluginTest.projectMatrixAuth

raul-arabaolaza:
* Give some time to OK button to become enabled after project type selection in create job page
* Give some time after login to load the job listing

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/JobsMixIn.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/JobsMixIn.java
@@ -34,6 +34,7 @@ public class JobsMixIn extends MixIn {
                 }
             }
         }).click();
+        elasticSleep(3000);
 
         clickButton("OK");
 
--- a/src/test/java/plugins/MatrixAuthPluginTest.java
+++ b/src/test/java/plugins/MatrixAuthPluginTest.java
@@ -52,10 +52,12 @@ public class MatrixAuthPluginTest extends AbstractJUnitTest {
 
         // if we login as Bob, he shouldn't see the job
         jenkins.login().doLogin("bob");
+        elasticSleep(2000);
         assertNull(getElement(by.href("job/"+j.name+"/")));
 
         // contorl assertion: alice shoudl see the link
         jenkins.login().doLogin("alice");
+        elasticSleep(2000);
         assertNotNull(getElement(by.href("job/"+j.name+"/")));
 
         // TODO: variant of href that takes laxed match
@@ -90,11 +92,13 @@ public class MatrixAuthPluginTest extends AbstractJUnitTest {
 
         // bob shouldn't be able to see it without adding a permission for him
         jenkins.login().doLogin("bob");
+        elasticSleep(2000);
         assertNull(getElement(by.href("job/"+j.name+"/")));
 
 
         // alice will expose this job to bob
         jenkins.login().doLogin("alice");
+        elasticSleep(2000);
         j.configure();
         {
             ProjectMatrixProperty p = new ProjectMatrixProperty(j);
@@ -107,6 +111,7 @@ public class MatrixAuthPluginTest extends AbstractJUnitTest {
 
         // bob should see this job
         jenkins.login().doLogin("bob");
+        elasticSleep(2000);
         assertNotNull(getElement(by.href("job/"+j.name+"/")));
     }
 }


armfergom:
Might be better to use `CapybaraPortingLayerImpl.waitFor(final By selector, final int timeoutSec)` instead of `Thread.sleep()` that is more fragile.

src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
```diff
@@ -85,6 +85,24 @@ public void clickButton(String text) {
...
 
+   /**
+    * Waits till the specified button is enabled and clicks it
+    *
+    * @param caption The button's text
+    * @param timeout The time to wait for the button to become enabled, in seconds
+    */
+    public void waitTillEnabledAndClick(String caption, int timeout, TimeUnit timeoutTimeUnit) {
+        final WebElement e = find(by.button(caption));
+        waitFor().withTimeout(timeout,timeoutTimeUnit)
```
recena: `waitFor().withTimeout(timeout, timeoutTimeUnit)`

raul-arabaolaza:
@armfergom Can you check if the last changes are more like you were talking about??

src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
```diff
@@ -233,6 +254,25 @@ public WebElement getElement(By selector) {
...
+    /**
+     * Works like  {@link #getElement(org.openqa.selenium.By)} but it waits some time for it to exist
+     *
+     * @param selector The selector for the element to get
+     * @param timeoutSec The maximun time to wait for the element to appear in seconds
+     * @return The finded element or null if not found in the specified time
+     */
+    public WebElement waitAndGetElement(By selector, int timeoutSec) {
+        WebElement element = getElement(selector);
+        if(element != null) {
```
recena: `if (element != null) {`

src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
```diff
@@ -233,6 +254,25 @@ public WebElement getElement(By selector) {
         return all.get(0);
     }
 
+    /**
+     * Works like  {@link #getElement(org.openqa.selenium.By)} but it waits some time for it to exist
```
recena: `Works like {@link #getElement(org.openqa.selenium.By)}`

recena:
I think @olivergondza's opinion would be nice here.

LGTM, though I'm not sure about the method names `waitAndGetElement` and `waitTillEnabledAndClick`

src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
```diff
@@ -233,6 +254,25 @@ public WebElement getElement(By selector) {
...
+     *
+     * @param selector The selector for the element to get
+     * @param timeoutSec The maximun time to wait for the element to appear in seconds
+     * @return The finded element or null if not found in the specified time
+     */
+    public WebElement waitAndGetElement(By selector, int timeoutSec) {
```
olivergondza: Is is needed to introduce this method. Why is client not calling `waitFor(selector, timeoutSec)`? Hiding any exception behind `null` will actually make it harder to understand _why_ the element was not found.

raul-arabaolaza:
@olivergondza I made that method because sometimes (like in this test case) we don't always expect the element to appear. But when we expect it to appear we need to give it some time. This code was inspired by the already existing `getElement` but assuming some delay may occur. I don't like the idea of the tests having to deal with exception catching if using `waitFor(selector, timeoutSec)`, so that's the reason I created that method on the class.

Maybe `getElement` would be a better name

olivergondza:
> I don't like the idea of the tests having to deal with exception catching if using waitFor(selector, timeoutSec), so that's the reason I created that method on the class.

They should not. If the test fail with `TimeoutExcpetion`, `NullPointerException` or `WebDriverException` you have a decent chance to understand the cause and fix it. If the message, stacktrace and causes is throw away this way JUnit tels you "non-null expected, null given". You have no clue where to look.

raul-arabaolaza:
> They should not. If the test fail with TimeoutExcpetion, NullPointerException or WebDriverException you have a decent chance to understand the cause and fix it. If the message, stacktrace and causes is throw away this way JUnit tels you "non-null expected, null given". You have no clue where to look.

The test implements the following logic

1. Create an admin user and a regular user
2. Create a job using the admin user
3. Check that regular user can not see the job
4. Give permission to the regular user
5. Check that the regular user can now see the job

If using waitFor then the expected outcome in step 3 is a `TimeoutException`and in step 5 to not give any exception at all. So, to implement it with waitFor you have to use code like this (in pseudocode):

```
try {
   waitFor
   //failTest
catch (TimeoutException e) {
   //continue testing without further catching exceptions
}
``` 

In my opinion the `assertNull` and `assertNotNull`implemented variants are more readable than the previous pseudocode. I can add some description to the failures to clarify it a bit if you want.

Anyway I don't have very strong feelings for this, if you still prefer the waitFor version I ' ll do it like that :)

Regards

Mobrockers:
> If using waitFor then the expected outcome in step 3 is a TimeoutException and in step 5 to not give any exception at all. So, to implement it with waitFor you have to use code like this (in pseudocode):

The underlying driver behind this is selenium webdriver right? In selenium webdriver it is the standard to never use findElement(by) when the expected outcome is for the element to not be found. Instead you are supposed to use findElements(by).size() == 0. I see that in this project findElements is wrapped by the method all(), so why are you not using all().size() == 0 on step number 3?

raul-arabaolaza:
Because I need to give it some time to appear, test was failing in some
machines due to the list of jobs not having enough time to render before
the existence check

Mobrockers:
Then you need to wait for an element you are sure will exist when the page is fully rendered, before checking step number 3.

raul-arabaolaza:
No, I need to check that element does not exist in step 3 and that it
exists in step 5. Giving it some time to render in both cases

Mobrockers:
That does not change anything though. First you wait until the page is fully rendered by waiting for an element that is sure to be available, then you check if the element does not exist in step 3 by using all().size() == 0, and find() in step 5. This gives it time to render the elements, does not break selenium convention, and does not throw an unnecessary exception.

raul-arabaolaza:
@Mobrockers Oh I understand now. Thanks for your patience :)  I will update this PR ASAP, but currently I am travelling so it may take me a pair of days.

raul-arabaolaza:
@Mobrockers I've updated the PR, could you please check if this is more aligned with selenium conventions??

Mobrockers:
This is more in line with what I would expect from a selenium test, yes! Do keep in mind that I'm not a maintainer for the test harness so I can't comment on if it's now also enough in line with what's expected by Jenkins and or cloudbees.

src/test/java/plugins/MatrixAuthPluginTest.java
```diff
@@ -48,14 +49,23 @@ public void helloWorld() throws Exception {
...
+        //wait for main panel to appear to make sure page is rendered
+        waitFor(by.id("main-panel"));
+
         assertNull(getElement(by.href("job/"+j.name+"/")));
 
         // contorl assertion: alice shoudl see the link
         jenkins.login().doLogin("alice");
+
+        //wait for main panel to appear to make sure page is rendered
+        waitFor(by.id("main-panel"));
```
recena: `waitFor(by.id("main-panel"));` I think this is not stable. If the layout changes.... Is not there other way to check if the page was loaded?

raul-arabaolaza: Are you referring to not wait for a know element to appear or to not use id as identifier??

recena: This id `main-panel` belongs to the layout. If Jenkins changes the layout (his class ID for example), the test won't work.

By the way, the rest of the inline comment leave an space after `//`...

amuniz: The layout could change but not the ID of that element. And if it changes [this is not going to be the only test failing](https://github.com/jenkinsci/acceptance-test-harness/search?utf8=%E2%9C%93&q=main-panel).

src/test/java/plugins/MatrixAuthPluginTest.java
```diff
@@ -48,14 +49,23 @@ public void helloWorld() throws Exception {
...
 
         FreeStyleJob j = jenkins.jobs.create();
+
         j.save();
 
         // if we login as Bob, he shouldn't see the job
         jenkins.login().doLogin("bob");
+
+        //wait for main panel to appear to make sure page is rendered
+        waitFor(by.id("main-panel"));
```
amuniz: Better ` waitFor(final By selector, final int timeoutSec)`, the test does not hang forever (or until test time limit, which is 3 minutes I think).

src/test/java/plugins/MatrixAuthPluginTest.java
amuniz: I mean, the ID can change, of course it's a text file and can be edited 😄 but I don't think anybody want to change it in the next 20 years.

recena: @amuniz For that reason, nobody changes things that should be changed. In fact, this `ID` will be changed soon by me. There are other alternatives to do the check. For example, checking if the HTTP request has finished (page loaded)

Mobrockers: @recena selenium waits for page load by default, but the page having loaded does not guarantee that the browser has finished rendering, especially because of javascript actions that are only started after page load. Therefore it is recommended to wait until an element you know should exist has rendered on the page, before continuing with your test.

src/main/java/org/jenkinsci/test/acceptance/po/JobsMixIn.java
```diff
@@ -35,7 +35,11 @@ public JobsMixIn(ContainerPageObject context) {
             }
         }).click();
 
-        clickButton("OK");
+        //Sometimes we need to give some time to ok button to become enabled
+        waitTillEnabledAndClick("OK", 10, TimeUnit.SECONDS);
+
+        //sometimes job creation is not fast enough, so make sure it's finished before continue
+        waitFor(by.name("config"));
```
olivergondza: Is this waiting for config form to be interaction ready? In case it is, I prefer to do so in `Job.configure()` if really needed.

src/test/java/plugins/MatrixAuthPluginTest.java
recena: @Mobrockers 

> And besides, let's face it, this is a UI test. UI's change, and when UI's change tests have to be fixed. Such is life.

Well, we have to try finding the most robust way.

I'm going to check it to provide an alternative.

Mobrockers: @recena I'll be interested in what you find.

src/main/java/org/jenkinsci/test/acceptance/po/JobsMixIn.java
raul-arabaolaza: @recena I doubt that changing that `waitFor` for some JS magic (that anyway would probably depend also on the UI or worst be browser dependant) would  make a test, that already depends upon a button with a certain caption to exist, more robust or proof safe. That said, I truly believe ids are one of the most robust selectors out there, if the test fails the error message will be very informative and the fail easy to track. 

Changing that selector would mean probably a lot of collateral changes to be done, like CSS rules, JS code and so on. On that context, changing the test should not be difficult. So I must agree with @amuniz here. 

The main problem here is not tests being more or less fragile, but the lack of communication between jenkinsci and the acceptance-test-harness when UI changes, unlike unit tests there is no compiler here, so there is no easy way to ensure changes in the UI are propagated to tests. Honestly I believe a PR on this project is not the place to try to solve that problem, and that a separate Jira ticket should be created to discuss the actual fragility of this suite in general 

@olivergondza The problem is that without that `waitFor` is the `create` method itself the one that randomly fails in low resources machines (8 GB RAM), resulting in jobs not being created and this particular test failing.

src/test/java/plugins/MatrixAuthPluginTest.java
raul-arabaolaza: I remember reading about instead of using elements on the next page you could check already found elements of the previous page for a staled status. I believe this does not solves the main fragility problem but may be a little more robust because there are no new elements to depend uppon

recena: @raul-arabaolaza Let me to investigate it. IMO, `main-panel` is not the right `ID`.

Mobrockers: If you have a reference to a WebElement, and anything changes in this element, as soon as you try to use it, you'll get a StaleElementReferenceException. But this means you'll have to hold on to a reference of your webelement in your test. I suppose in the test harness this could be done since as far as I can see no Page Object Model is used.

You can then do `new WebDriverWait(driver, timeOut).until(ExpectedConditions.stalenessOf(element))` or whatever the test-harness equivalent is (if there is one).

src/main/java/org/jenkinsci/test/acceptance/po/JobsMixIn.java
```diff
@@ -35,7 +35,11 @@ public JobsMixIn(ContainerPageObject context) {
             }
         }).click();
 
-        clickButton("OK");
+        // Sometimes we need to give some time to ok button to become enabled
+        waitTillEnabledAndClick("OK", 10, TimeUnit.SECONDS);
+
+        // Sometimes job creation is not fast enough, so make sure it's finished before continue
+        waitFor(by.name("config"), 3);
```
amuniz: 3 seconds only? I would give it 10 like the previous one.

raul-arabaolaza: @amuniz Done!!

abayer:
:bug: Tried this branch out and it's still failing for me on the final `assertTrue` - the `bob` user can't see any jobs.

abayer:
Note that my test was on a local run - doing a CI run now.

abayer:
And on CI, I'm still seeing the same original error:
```
09:50:42 org.openqa.selenium.WebDriverException: Element is not clickable at point (1258, 7). Other element would receive the click: <ul id="breadcrumbs"></ul>
09:50:42 Command duration or timeout: 80 milliseconds
09:50:42 Build info: version: '2.53.0', revision: '35ae25b1534ae328c771e0856c93e187490ca824', time: '2016-03-15 10:43:46'
09:50:42 System info: host: 'jenkins', ip: '127.0.0.1', os.name: 'Linux', os.arch: 'amd64', os.version: '3.13.0-86-generic', java.version: '1.8.0_40'
09:50:42 Driver info: org.openqa.selenium.firefox.FirefoxDriver
09:50:42 Capabilities [{applicationCacheEnabled=true, rotatable=false, handlesAlerts=true, databaseEnabled=true, version=22.0, platform=LINUX, browserConnectionEnabled=true, nativeEvents=false, acceptSslCerts=true, webStorageEnabled=true, locationContextEnabled=true, browserName=firefox, takesScreenshot=true, javascriptEnabled=true, cssSelectorsEnabled=true}]
09:50:42 Session ID: d2ba944f-2d4a-4d7f-93dd-5b47b90af821
09:50:42 	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
09:50:42 	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
09:50:42 	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
09:50:42 	at java.lang.reflect.Constructor.newInstance(Constructor.java:422)
09:50:42 	at org.openqa.selenium.remote.ErrorHandler.createThrowable(ErrorHandler.java:206)
09:50:42 	at org.openqa.selenium.remote.ErrorHandler.throwIfResponseFailed(ErrorHandler.java:158)
09:50:42 	at org.openqa.selenium.remote.RemoteWebDriver.execute(RemoteWebDriver.java:678)
09:50:42 	at org.openqa.selenium.remote.RemoteWebElement.execute(RemoteWebElement.java:327)
09:50:42 	at org.openqa.selenium.remote.RemoteWebElement.click(RemoteWebElement.java:85)
09:50:42 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
09:50:42 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
09:50:42 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
09:50:42 	at java.lang.reflect.Method.invoke(Method.java:497)
09:50:42 	at org.openqa.selenium.support.events.EventFiringWebDriver$EventFiringWebElement$1.invoke(EventFiringWebDriver.java:335)
09:50:42 	at com.sun.proxy.$Proxy39.click(Unknown Source)
09:50:42 	at org.openqa.selenium.support.events.EventFiringWebDriver$EventFiringWebElement.click(EventFiringWebDriver.java:348)
09:50:42 	at org.jenkinsci.test.acceptance.po.CapybaraPortingLayerImpl.check(CapybaraPortingLayerImpl.java:280)
09:50:42 	at org.jenkinsci.test.acceptance.po.Control.check(Control.java:87)
09:50:42 	at org.jenkinsci.test.acceptance.plugins.matrix_auth.MatrixRow._check(MatrixRow.java:27)
09:50:42 	at org.jenkinsci.test.acceptance.plugins.matrix_auth.MatrixRow.on(MatrixRow.java:17)
09:50:42 	at plugins.MatrixAuthPluginTest.projectMatrixAuth(MatrixAuthPluginTest.java:118)
```

src/test/java/plugins/MatrixAuthPluginTest.java
```diff
@@ -48,14 +49,23 @@ public void helloWorld() throws Exception {
...
 
         // if we login as Bob, he shouldn't see the job
         jenkins.login().doLogin("bob");
+
+        // wait for main panel to appear to make sure page is rendered
+        waitFor(by.id("main-panel"), 10);
+
         assertNull(getElement(by.href("job/"+j.name+"/")));
 
         // contorl assertion: alice shoudl see the link
```
recena: typo (*unrelated*)

raul-arabaolaza:
@abayer Could you give some details about environment? Specially browser version, because it woks on my local machine...

src/test/java/plugins/MatrixAuthPluginTest.java
```diff
@@ -52,22 +53,24 @@ public void helloWorld() throws Exception {
...
 
+        jenkins.logout();
+
         // if we login as Bob, he shouldn't see the job
         jenkins.login().doLogin("bob");
 
-        // wait for main panel to appear to make sure page is rendered
-        waitFor(by.id("main-panel"), 10);
+        // check for job's existence
+        driver.get(jenkins.getCurrentUrl() + "job/" + j.name + "/");
```
recena: @raul-arabaolaza Have you found something to get the Jenkins URL? `jenkins.getCurrentUrl()` may not be sure.

recena:
The PoC provided here https://github.com/jenkinsci/acceptance-test-harness/pull/144/commits/7325974cf455e20f022c4f25f33576cfa054295a, has been included in this PR. I0m closing the PR https://github.com/jenkinsci/acceptance-test-harness/pull/144 now.

src/test/java/plugins/MatrixAuthPluginTest.java
abayer: :bug: Yeah, that just gets your current URL, not the root Jenkins URL.

raul-arabaolaza:
After a lot of digging it seems that the PO responsible of job creation may have some problems that affect this (and maybe other) tests. So I am going to create an issue for that and I have simplified this PR to address only the issues relative to this ticket.

src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
```diff
@@ -47,6 +47,7 @@
     @Inject
     protected ElasticTime time;
 
+
```
recena: Unrelated.

src/test/java/plugins/MatrixAuthPluginTest.java
```diff
@@ -1,5 +1,8 @@
 package plugins;
 
+import com.trilead.ssh2.log.Logger;
+import org.apache.commons.logging.Log;
```
recena: Unnecessary 🐛

recena:
🐝

abayer:
Do we need this any more after #146?

recena:
@abayer Sure. I think this PR improve both tests. Anyway, [this](https://github.com/jenkinsci/acceptance-test-harness/pull/146) should not be considered a final solution.

abayer:
Fair enough!

I'm testing this out without #146 to make sure it does the trick.

abayer:
And no joy - same failure as before:
```
12:50:43 org.openqa.selenium.WebDriverException: Element is not clickable at point (1258, 7). Other element would receive the click: <ul id="breadcrumbs"></ul>
12:50:43 Command duration or timeout: 96 milliseconds
12:50:43 Build info: version: '2.53.0', revision: '35ae25b1534ae328c771e0856c93e187490ca824', time: '2016-03-15 10:43:46'
12:50:43 System info: host: 'jenkins', ip: '127.0.0.1', os.name: 'Linux', os.arch: 'amd64', os.version: '3.13.0-86-generic', java.version: '1.8.0_40'
12:50:43 Driver info: org.openqa.selenium.firefox.FirefoxDriver
12:50:43 Capabilities [{applicationCacheEnabled=true, rotatable=false, handlesAlerts=true, databaseEnabled=true, version=22.0, platform=LINUX, browserConnectionEnabled=true, nativeEvents=false, acceptSslCerts=true, webStorageEnabled=true, locationContextEnabled=true, browserName=firefox, takesScreenshot=true, javascriptEnabled=true, cssSelectorsEnabled=true}]
12:50:43 Session ID: d1145ddb-2dd8-43a0-8a9e-7e823b2ba41c
12:50:43 	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
12:50:43 	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
12:50:43 	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
12:50:43 	at java.lang.reflect.Constructor.newInstance(Constructor.java:422)
12:50:43 	at org.openqa.selenium.remote.ErrorHandler.createThrowable(ErrorHandler.java:206)
12:50:43 	at org.openqa.selenium.remote.ErrorHandler.throwIfResponseFailed(ErrorHandler.java:158)
12:50:43 	at org.openqa.selenium.remote.RemoteWebDriver.execute(RemoteWebDriver.java:678)
12:50:43 	at org.openqa.selenium.remote.RemoteWebElement.execute(RemoteWebElement.java:327)
12:50:43 	at org.openqa.selenium.remote.RemoteWebElement.click(RemoteWebElement.java:85)
12:50:43 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
12:50:43 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
12:50:43 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
12:50:43 	at java.lang.reflect.Method.invoke(Method.java:497)
12:50:43 	at org.openqa.selenium.support.events.EventFiringWebDriver$EventFiringWebElement$1.invoke(EventFiringWebDriver.java:335)
12:50:43 	at com.sun.proxy.$Proxy39.click(Unknown Source)
12:50:43 	at org.openqa.selenium.support.events.EventFiringWebDriver$EventFiringWebElement.click(EventFiringWebDriver.java:348)
12:50:43 	at org.jenkinsci.test.acceptance.po.CapybaraPortingLayerImpl.check(CapybaraPortingLayerImpl.java:259)
12:50:43 	at org.jenkinsci.test.acceptance.po.Control.check(Control.java:87)
12:50:43 	at org.jenkinsci.test.acceptance.plugins.matrix_auth.MatrixRow._check(MatrixRow.java:27)
12:50:43 	at org.jenkinsci.test.acceptance.plugins.matrix_auth.MatrixRow.on(MatrixRow.java:17)
12:50:43 	at plugins.MatrixAuthPluginTest.projectMatrixAuth(MatrixAuthPluginTest.java:129)
```
So still a :bug: from me - it doesn't fix the problem.

raul-arabaolaza:
@abayer The intention of this PR was not to fix the same problems that #146 solves, that said it may fail because of window size and scrolling as any other selenium test in the suite.The question is that it solves another problems that may arise apart of the scrolling issue. For example sometimes jobs are not correctly created due to lack of time in JobsMixIn also it checks for jobs visibility in a less error prone way than before. Given the fact that #146 is already merged could you test again to see if it solves the bug is intended to??

raul-arabaolaza:
@abayer I' ve just merged the latest changes in the PR and it works, Could you please check it when you have some time??

abayer:
:bee: Ok, I haven't ever seen the other failure cases, but as long as this doesn't cause problems, it should be ok.

src/test/java/plugins/MatrixAuthPluginTest.java
```diff
@@ -48,17 +57,27 @@ public void helloWorld() throws Exception {
...
 
+        jenkins.logout();
+
         // if we login as Bob, he shouldn't see the job
         jenkins.login().doLogin("bob");
-        assertNull(getElement(by.href("job/"+j.name+"/")));
 
-        // contorl assertion: alice shoudl see the link
+        // check for job's existence
+        driver.get(controller.getUrl() + "job/" + j.name + "/");
```
olivergondza: `j.open()` is the convenience method for this.

raul-arabaolaza:
@reviewbybees done

src/test/java/plugins/MatrixAuthPluginTest.java
raul-arabaolaza: @olivergondza I can't use `Matchers.pageObjectExists` because to access the url we need to login first, so `Matchers.pageObjectExists`receives a 403 even in the case the job exist. I've switched to `j.open()` because i believe modifying `Matchers.pageObjectExists` to take into account login credentials is out of context for this PR

abayer:
re-:bee:

recena:
:bee:

recena:
@reviewbybees done

PR #132, opened 2016-06-17
[FIXED JENKINS-36045] Bump sleep for waiting for cidfile

abayer:
The half-second sleep seems to hit race conditions periodically,
especially if the host is under load. Moving it to a second got rid of
the issues in my testing - from seeing it error out due to an empty
cidfile roughly half the time in 10 tests to zero times in 10 tests.

cc @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
@@ -117,7 +117,7 @@ public class DockerImage {
 
     private String waitForCid(CommandBuilder docker, File cidFile, File logfile, Process p) throws InterruptedException, IOException {
         for (int i = 0; i < 10; i++) {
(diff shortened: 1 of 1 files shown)


raul-arabaolaza:
This seems a little arbitrary, maybe in other systems we need to wait more or less.

Have you evaluated the possibility to check the availability of the file instead of waiting a fixed (arbitrary) amount of time?

andresrc:
:bee:

is there some condition that we could use to do something like the `waitUntil` ? (file existence or something like that)

mafraba:
🐝 
I see it ok to unblock the issue. 
Anyway I think the proper way to do it would be to wait for the file **with a timeout** ?

abayer:
The problem is that the file actually gets created but is empty for a brief moment, and it seems that if we open it at that time, Docker itself is unable to write to it, so it stays empty forever.

PR #133, opened 2016-06-17
[JENKINS-36046] Fixing FileCredentials and StringCredentials controls

armfergom:
Fixing `FileCredentials` and `StringCredentials` controls to be compatible with credentials 2.x line.

See [JENKINS-36046](https://issues.jenkins-ci.org/browse/JENKINS-36046) 

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/credentials/FileCredentials.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/credentials/FileCredentials.java
@@ -9,7 +9,7 @@ import org.jenkinsci.test.acceptance.po.PageObject;
 @Describable("Secret file")
 public class FileCredentials extends BaseStandardCredentials {
 
-    public Control file = control("file");
+    public Control file = control(by.name("file"));
     
     public FileCredentials(PageObject context, String path) {
         super(context, path);
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/credentials/StringCredentials.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/credentials/StringCredentials.java
@@ -9,7 +9,7 @@ import org.jenkinsci.test.acceptance.po.PageObject;
 @Describable("Secret text")
 public class StringCredentials extends BaseStandardCredentials {
 
-    public Control secret = control("secret");
+    public Control secret = control(by.name("_.secret"));
     
     public StringCredentials(PageObject context, String path) {
         super(context, path);


src/main/java/org/jenkinsci/test/acceptance/plugins/credentials/FileCredentials.java
```diff
@@ -9,7 +9,7 @@
 @Describable("Secret file")
 public class FileCredentials extends BaseStandardCredentials {
 
-    public Control file = control("file");
+    public Control file = control(by.name("file"));
```
amuniz: Is `control` not searching by name already?

src/main/java/org/jenkinsci/test/acceptance/plugins/credentials/StringCredentials.java
```diff
@@ -9,7 +9,7 @@
 @Describable("Secret text")
 public class StringCredentials extends BaseStandardCredentials {
 
-    public Control secret = control("secret");
+    public Control secret = control(by.name("_.secret"));
```
amuniz: Don't you want `control(by.name("_.secret"), by.name("secret"));` to keep compatibility with previous versions.

recena:
🐝

src/main/java/org/jenkinsci/test/acceptance/plugins/credentials/FileCredentials.java
armfergom: `control` searches by the `path` attribute.

In credentials 1.x, the `path` attribute was `file`. In credentials 2.x the `path` attribute is `definition/file`. Same for `secret`.

However, in both 1.x and 2.x, the name attribute is unchanged (for both secret and file)

src/main/java/org/jenkinsci/test/acceptance/plugins/credentials/StringCredentials.java
armfergom: See inline reply to your first comment.

amuniz: Ok, so if `name` attribute was unchanged, why are you changing it here? ` "_.secret"` != `"secret"`.

Anyway, my question is: is this working on 1.x? If so, I'm ok.

armfergom: What changes from 1.x to 2.x is the `path` attribute. `control` by default uses the `path` attribute.

What has NOT changed from 1.x to 2.x is the `name` attribute. So, the simplest way to make this work with both 1.x and 2.x lines is to look by `name` instead of by `path`.

Answering your question, yes, works for both.

src/main/java/org/jenkinsci/test/acceptance/plugins/credentials/FileCredentials.java
olivergondza: The name attribute is not guaranteed to be unique on the page, though. Not sure it is a problem in this case.

olivergondza: `control("definition/file", "file")` would work.

armfergom: Well, I really doubt we will have conflicts here since it is a very simple page the one that's rendered when creating a credential.

That said, I wasn't aware of that use of the `control` method, and it does seem like the most reasonable way to implement this. Very useful, thanks!

src/main/java/org/jenkinsci/test/acceptance/plugins/credentials/StringCredentials.java
armfergom: @amuniz just letting you know that I will be using something similar to what you've initially suggested (`control(by.name("_.secret"), by.name("secret"))`) but using the `path` instead of the name as suggested by Oliver. Thanks for the review!

armfergom:
Superseded by #139

PR #135, opened 2016-06-20
Switches the runDockerTests profile to use a JUnit Category.

abayer:
The time spent instantiating each test and launching Firefox, etc when
you're just trying to run the Docker tests is...impressively
large. That's because the FilterRule doesn't kick in until after the
WebDriver has been injected and therefore the browser has been
launched. That also results in a lot of skipped tests rather than
those tests just not showing up at all in the Maven output, which IMO
is preferable.

cc @reviewbybees @olivergondza @kohsuke - yes, it's an otherwise redundant annotation alongside every occurrence of `@WithDocker`, but it makes my life so much better. =)


svanoort:
🐝

olivergondza:
I guess that is simpler. Can you delete `docker-only.groovy`?

olivergondza:
:+1:

jtnord:
:bee:

rsandell:
:bee:

PR #138, opened 2016-06-22
[JENKINS-36045] Rework acquiring Docker CID.

abayer:
Also effectively rolls back 74ec1821ced1d2ef2b2567e92e699117370ff1bf
as unneeded.

NOTE: Don't merge this until I've had a chance to run a few more full runs to be sure it's doing the trick - it certainly appears to be fixing things so far, but I've been bit by false positives on this issue before. =) It's still just a band-aid at the core - I can't for the life of me figure out why it's happening in the first place, but this is the best workaround I've got yet.

EDIT:
I've left the retry in for now, but may end up opting to remove it entirely. The new approach (as described below) reworks how we get the container ID and logs entirely and doesn't seem to run into problems.

```
More testing needed to verify this actually does the trick
consistently, but it does *work* in the non-failure case for
sure. We're now detaching the container, getting the container ID from
the "docker run" output (since that's the output from a detached run)
and running a side process piping "docker logs --follow (cid)" to the
log file. That side process gets automatically killed at the end of
the test when we kill and rm the container as well.
```

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
@@ -3,6 +3,7 @@ package org.jenkinsci.test.acceptance.docker;
 import com.google.inject.Inject;
 import hudson.remoting.Which;
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.IOUtils;
 import org.jenkinsci.test.acceptance.utils.SHA1Sum;
 import org.jenkinsci.utils.process.CommandBuilder;
 import org.jvnet.hudson.annotation_indexer.Index;
@@ -62,6 +63,26 @@ public class Docker {
         } catch (InterruptedException | IOException e) {
             return false;
         }
+    }
+
+    /**
+     * Checks if a given container is currently running.
+     *
+     * @param container Container id
+     * @return false if the container is not running, true otherwise
+     */
+    public boolean isContainerRunning(String container) throws IOException, InterruptedException {
+        ProcessBuilder pBuilder = cmd("ps").add("-q", "--filter", "\"id=" + container + '"').build();
+        Process psProcess = pBuilder.start();
+        psProcess.waitFor();
+        if (psProcess.exitValue() == 0) {
+            String psOutput = IOUtils.toString(psProcess.getInputStream());
+            if (psOutput.contains(container)) {
+                return true;
+            }
+        }
+
+        return false;
     }
 
     /**
--- a/src/main/java/org/jenkinsci/test/acceptance/docker/DockerContainerHolder.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/docker/DockerContainerHolder.java
@@ -47,10 +47,31 @@ public class DockerContainerHolder<T extends DockerContainer> implements Provide
             Class<T> fixture = (Class<T>) type.getRawType();
             File buildlog = diag.touch("docker-" + fixture.getSimpleName() + ".build.log");
             File runlog = diag.touch("docker-" + fixture.getSimpleName() + ".run.log");
-            try {
-                container = docker.build(fixture, buildlog).start(fixture).withPortOffset(portOffset).withLog(runlog).start();
-            } catch (InterruptedException | IOException e) {
-                throw new Error("Failed to start container " + fixture.getName(), e);
+            Exception launchException = null;
+            boolean keepTrying = true;
+            int i = 0;
+            for (; i < 5 && keepTrying; i++) {
+                try {
+                    container = docker.build(fixture, buildlog).start(fixture).withPortOffset(portOffset).withLog(runlog).start();
+                    launchException = null;
+                    break;
+                } catch (InterruptedException | IOException e) {
+                    launchException = e;
+                    // Only keep trying if the error is cidFile related.
+                    if (!e.getMessage().contains("docker didn't leave CID file")) {
+                        keepTrying = false;
+                    }
+                }
+                try {
+                    // If we've got this far, that means there was a failure - sleep for 5 seconds and try again.
+                    Thread.sleep(5000);
+                } catch (InterruptedException e2) {
+                    // Swallow it
+                }
+            }
+
+            if (launchException != null) {
+                throw new Error("Failed to start container after " + i + " tries - " + fixture.getName(), launchException);
             }
         }
         return container;
--- a/src/main/java/org/jenkinsci/test/acceptance/docker/DockerContainerHolder.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/docker/DockerContainerHolder.java
@@ -47,31 +47,10 @@ public class DockerContainerHolder<T extends DockerContainer> implements Provide
             Class<T> fixture = (Class<T>) type.getRawType();
             File buildlog = diag.touch("docker-" + fixture.getSimpleName() + ".build.log");
             File runlog = diag.touch("docker-" + fixture.getSimpleName() + ".run.log");
-            Exception launchException = null;
-            boolean keepTrying = true;
-            int i = 0;
-            for (; i < 5 && keepTrying; i++) {
-                try {
-                    container = docker.build(fixture, buildlog).start(fixture).withPortOffset(portOffset).withLog(runlog).start();
-                    launchException = null;
-                    break;
-                } catch (InterruptedException | IOException e) {
-                    launchException = e;
-                    // Only keep trying if the error is cidFile related.
-                    if (!e.getMessage().contains("docker didn't leave CID file")) {
-                        keepTrying = false;
-                    }
-                }
-                try {
-                    // If we've got this far, that means there was a failure - sleep for 5 seconds and try again.
-                    Thread.sleep(5000);
-                } catch (InterruptedException e2) {
-                    // Swallow it
-                }
-            }
-
-            if (launchException != null) {
-                throw new Error("Failed to start container after " + i + " tries - " + fixture.getName(), launchException);
+            try {
+                container = docker.build(fixture, buildlog).start(fixture).withPortOffset(portOffset).withLog(runlog).start();
+            } catch (InterruptedException | IOException e) {
+                throw new Error("Failed to start container - " + fixture.getName(), e);
             }
         }
         return container;
--- a/src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
@@ -119,23 +119,16 @@ public class DockerImage {
     }
 
     private String waitForCid(CommandBuilder docker, Process p) throws InterruptedException, IOException {
-        for (int i = 0; i < 10; i++) {
-            Thread.sleep(500);
+        p.waitFor();
 
-            String output = IOUtils.toString(p.getInputStream());
+        String output = IOUtils.toString(p.getInputStream());
 
-            try {
-                if (p.exitValue() != 0) {
-                    throw new IOException("docker died unexpectedly: " + docker + "\n" + output);
-                }
-            } catch (IllegalThreadStateException e) {
-                // Docker is still launching, so we move on.
-            }
+        if (p.exitValue() != 0) {
+            throw new IOException("docker died unexpectedly: " + docker + "\n" + output);
+        }
 
-            if (output != null && output.length() != 0) {
-                return output.trim();
-            }
-
+        if (output != null && output.length() != 0) {
+            return output.trim();
         }
 
         throw new IOException("docker didn't output a container id, yet is still running. Huh?: "+docker+"\n"+IOUtils.toString(p.getInputStream()));
--- a/src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
@@ -82,7 +82,6 @@ public class DockerImage {
         cidFile.delete();
         cidFile.deleteOnExit();
         docker.add("--cidfile="+cidFile);//strange behaviour in some docker version cidfile needs to come before
-
         for (int p : starter.ports) {
             docker.add("-p", starter.getPortMapping(p));
         }
@@ -117,7 +116,7 @@ public class DockerImage {
 
     private String waitForCid(CommandBuilder docker, File cidFile, File logfile, Process p) throws InterruptedException, IOException {
         for (int i = 0; i < 10; i++) {
-            Thread.sleep(1000);
+            Thread.sleep(500);
 
             String cid = FileUtils.readFileToString(cidFile);
             if (cid != null && cid.length() != 0) return cid;
--- a/src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
@@ -1,6 +1,7 @@
 package org.jenkinsci.test.acceptance.docker;
 
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.SystemUtils;
 import org.jenkinsci.utils.process.CommandBuilder;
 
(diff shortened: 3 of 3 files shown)


abayer:
ping @reviewbybees and @olivergondza - I'm fairly confident this one is right. =)

andresrc:
:bee:

src/main/java/org/jenkinsci/test/acceptance/docker/DockerContainerHolder.java
```diff
@@ -47,10 +47,31 @@ public synchronized T get() {
...
+                    // Only keep trying if the error is cidFile related.
+                    if (!e.getMessage().contains("docker didn't leave CID file")) {
+                        keepTrying = false;
+                    }
+                }
+                try {
+                    // If we've got this far, that means there was a failure - sleep for 5 seconds and try again.
+                    Thread.sleep(5000);
+                } catch (InterruptedException e2) {
+                    // Swallow it
```
jtnord: and so If I am trying to abort the test run with Ctrl+C or the ilk you will tell me to take a long walk of a short pier?

src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
```diff
@@ -65,6 +66,26 @@ public boolean isAvailable() {
     }
 
     /**
+     * Checks if a given container is currently running.
+     *
+     * @param container Container id
+     * @return false if the container is not running, true otherwise
+     */
+    public boolean isContainerRunning(String container) throws IOException, InterruptedException {
+        ProcessBuilder pBuilder = cmd("ps").add("-q", "--filter", "\"id=" + container + '"').build();
```
jtnord: Go away with your UNIX only nonsense.

src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
```diff
@@ -92,45 +89,56 @@ public DockerImage(String tag) {
...
             }
+
+            if (output != null && output.length() != 0) {
+                return output.trim();
+            }
+
         }
 
-        throw new IOException("docker didn't leave CID file yet still running. Huh?: "+docker+"\n"+FileUtils.readFileToString(logfile));
+        throw new IOException("docker didn't leave CID file yet still running. Huh?: "+docker+"\n"+IOUtils.toString(p.getInputStream()));
```
jtnord: but you didn;t ask it to leave a CID file...

src/main/java/org/jenkinsci/test/acceptance/docker/DockerContainerHolder.java
abayer: Yeah, I should probably yank the retry.

src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
abayer: And I will reword. =)

src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
```diff
@@ -92,45 +89,56 @@ public DockerImage(String tag) {
...
 
             try {
-                p.exitValue();
-                throw new IOException("docker died unexpectedly: "+docker+"\n"+ FileUtils.readFileToString(logfile));
+                if (p.exitValue() != 0) {
+                    throw new IOException("docker died unexpectedly: " + docker + "\n" + output);
+                }
             } catch (IllegalThreadStateException e) {
-                //Docker is still running okay.
+                // Docker is still launching, so we move on.
```
jtnord: What - why...?
THis should not happen as
1) you have read the entire stream to a String (it should not return until you get EOF?)
2) if this could happen use `waitFor()`

src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
```diff
@@ -92,45 +89,56 @@ public DockerImage(String tag) {
...
         } catch (ReflectiveOperationException e) {
             throw new Error(e);
         }
     }
 
-    private String waitForCid(CommandBuilder docker, File cidFile, File logfile, Process p) throws InterruptedException, IOException {
+    private String waitForCid(CommandBuilder docker, Process p) throws InterruptedException, IOException {
         for (int i = 0; i < 10; i++) {
-            Thread.sleep(1000);
+            Thread.sleep(500);
```
jtnord: So why do you need to sleep - you now read the inputStream fully with `IOUtils.toString` - this should block until EOF is received..

src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
```diff
@@ -92,45 +89,56 @@ public DockerImage(String tag) {
...
+            t.init(cid, logProcess, logfile);
             return t;
         } catch (ReflectiveOperationException e) {
             throw new Error(e);
         }
     }
 
-    private String waitForCid(CommandBuilder docker, File cidFile, File logfile, Process p) throws InterruptedException, IOException {
+    private String waitForCid(CommandBuilder docker, Process p) throws InterruptedException, IOException {
         for (int i = 0; i < 10; i++) {
```
jtnord: There should be no reason to loop anymore.  if you do loop for whatever reason then `output` will likely be garbage

src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
abayer: Good point.

src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
abayer: Killed the sleep to death.

src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
```diff
@@ -92,45 +89,49 @@ public DockerImage(String tag) {
...
+        if (p.exitValue() != 0) {
+            throw new IOException("docker died unexpectedly: " + docker + "\n" + output);
+        }
+
+        if (output != null && output.length() != 0) {
+            return output.trim();
         }
 
-        throw new IOException("docker didn't leave CID file yet still running. Huh?: "+docker+"\n"+FileUtils.readFileToString(logfile));
+        throw new IOException("docker didn't output a container id, yet is still running. Huh?: "+docker+"\n"+IOUtils.toString(p.getInputStream()));
```
jtnord: Yeah - but docker has exited, and you have already read the output...
so `IOUtils.toString()` a second time is likely to throw an IOException as the stream is closed?
Even if not - I doubt you would have any more success this time around...

src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
```diff
@@ -92,45 +89,49 @@ public DockerImage(String tag) {
...
 
-    private String waitForCid(CommandBuilder docker, File cidFile, File logfile, Process p) throws InterruptedException, IOException {
-        for (int i = 0; i < 10; i++) {
-            Thread.sleep(1000);
+    private String waitForCid(CommandBuilder docker, Process p) throws InterruptedException, IOException {
+        p.waitFor();
 
-            String cid = FileUtils.readFileToString(cidFile);
-            if (cid != null && cid.length() != 0) return cid;
+        String output = IOUtils.toString(p.getInputStream());
```
jtnord: So OS buffering may become an issue.
the process may not exit as it has output that has not been read.

Normally the OS would buffer some amount and a CID should not be an issue - but I'm not convinced that doing it like this is a "Good Thing"™

abayer: So what would you recommend?

src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
abayer: That remaining `IOUtils.toString()` was an oversight. Changing it to have a much clearer exception.

abayer:
fwiw, I've now got three full runs without any of the flakiness showing up, so I think this really, really does the trick.

src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
```diff
@@ -65,6 +66,26 @@ public boolean isAvailable() {
...
+        Process psProcess = pBuilder.start();
+        psProcess.waitFor();
```
jtnord: If you have a lot of running containers will this exit as writing the output could block on it being read...

jtnord: Damn mobile view!
There should only be a certain amount of output, but stream buffers are os depenant and writing could block causing the process to never exit.

abayer: So same concern as https://github.com/jenkinsci/acceptance-test-harness/pull/138#discussion_r68136355? Lemme know what the right way to do this is and I'll happily do so. =)

src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
jtnord: I belei e you can just use ioutils.tostring without the waitfor...
The input stream sould return -1 when the process terminates at which point you have read everything and lnow that the process has ended, so the waitfor should be un needed.
There is a few shoulds in there but i think i am right :)

abayer: Ok, I'll give it a shot.

jtnord:
Lgtm, lets see if you get stable test runs!

:bee:

abayer:
Queuing up another full run featuring this.

abayer:
Wellll, that broke everything - getting rid of the waitFor, I mean. Reverting.

abayer:
More specifically, trying to check the exitValue without first having called waitFor == doooooooom. I think we can chance the risk of buffer issues.

jtnord:
What error, that implies something funky ie tostring returns before the eof...

abayer:
```
16:29:42 unableToConnectWrongCredentials(plugins.SshSlavesPluginTest)  Time elapsed: 60.836 sec  <<< ERROR!
16:29:42 java.lang.IllegalThreadStateException: process hasn't exited
16:29:42 	at java.lang.UNIXProcess.exitValue(UNIXProcess.java:421)
16:29:42 	at org.jenkinsci.test.acceptance.docker.DockerImage.waitForCid(DockerImage.java:124)
16:29:42 	at org.jenkinsci.test.acceptance.docker.DockerImage.start(DockerImage.java:101)
16:29:42 	at org.jenkinsci.test.acceptance.docker.DockerImage.access$500(DockerImage.java:19)
16:29:42 	at org.jenkinsci.test.acceptance.docker.DockerImage$Starter.start(DockerImage.java:213)
16:29:42 	at org.jenkinsci.test.acceptance.docker.DockerContainerHolder.get(DockerContainerHolder.java:51)
16:29:42 	at plugins.SshSlavesPluginTest.setUp(SshSlavesPluginTest.java:57)
```
i.e.

jtnord:
Ioutils does not read intil eof... Ffs...
Roll your own?

abayer:
I'd feel safer just saying "fuck it!" and letting `waitFor()` do the trick for me, if that's ok.

jtnord:
Ill test on windows tomorrow, as i assume you tested just mac locally and linux on the ci server?

abayer:
Yeah, since I'm not hugely concerned about supporting Docker on Windows beta. =)

abayer:
That and I don't have a Windows box.

jtnord:
Eh. Docker on windows is not beta...
You are mistaking native windows docker and the docker cli, talking to a remote docker hosts e.g. docker toolbox.
It should not make a difference which you use it's the os process buffering that is ths potential issue.
And you should care about windows as that platform brings out all the issies caused ny bad coding :)

rsandell:
:bee: depending on @jtnord 's final say.

olivergondza:
To rephrase the problem, CID might not be written in case ATH have it opened (does it in loop) at the time docker is writing it.

How about waiting for `docker run` co complete and then read the cidfile? There is no polling needed, no chance for a race like this and we will catch the hypothetical risk resuming the execution after CID was written but `docker run` failed (compared to the original implementation).

jtnord:
So I ran the PR with just `LdapPluginTest` and the docker bits seem to work (not sure if a container dumps out a lot of info that we would have an issue later on - so beware if we start seeing hanging tests this is it!).

The tests fail spectacularly (so much for my hard work)... Firefox [crashes at startup](https://github.com/SeleniumHQ/selenium/issues/2110)  and the Selenium barfs out the test with chrome

```
org.openqa.selenium.WebDriverException: unknown error: keys should be a string
<snip>
        at org.openqa.selenium.remote.RemoteWebElement.sendKeys(RemoteWebElement.java:122)
<snip>
        at org.jenkinsci.test.acceptance.po.Control.set(Control.java:128)
        at org.jenkinsci.test.acceptance.po.LdapSecurityRealm.configure(LdapSecurityRealm.java:69)
```

But the docker stuff appears to work for at least the LDAP test (no point trying the others as I would have to inspect all of the output :-(  rather than looking for test that fail or hang)

so :bee: and :cry:

jtnord:
FWIW [`IOUtils,toString()`](http://svn.apache.org/viewvc/commons/proper/io/tags/2.4/src/main/java/org/apache/commons/io/IOUtils.java?revision=1351813&view=markup#l1998) does read until the EOF, so the fact that the stream was closed but the process had not exited is slightly surprising - but I wonder if this is just a race condition in that the streams are closed before the process has fully exited, then the return val is set....
So `IOUtils.toString() && p.waitfor() && p.exitValue()` could be correct thing to do :-/

Digging into this seems I am possibly correct here...

https://www.securecoding.cert.org/confluence/display/java/FIO07-J.+Do+not+let+external+processes+block+on+IO+buffers

src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
```diff
@@ -65,6 +66,26 @@ public boolean isAvailable() {
...
+        if (psProcess.exitValue() == 0) {
```
jtnord: Well I did not know it but `waitFor` returns the exit status - so this can be simplified.

abayer:
@olivergondza Not entirely sure what you mean - the important thing isn't the cidfile specifically, it's getting the container ID and getting the container run log, and this approach does seem to be a more stable way to do that...

src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
abayer: Updated to just use `waitFor` in both places now.

src/main/java/org/jenkinsci/test/acceptance/docker/Docker.java
```diff
@@ -65,6 +66,26 @@ public boolean isAvailable() {
...
+     * Checks if a given container is currently running.
+     *
+     * @param container Container id
+     * @return false if the container is not running, true otherwise
+     */
+    public boolean isContainerRunning(String container) throws IOException, InterruptedException {
+        ProcessBuilder pBuilder = cmd("ps").add("-q", "--filter", "\"id=" + container + '"').build();
+        Process psProcess = pBuilder.start();
+
+        if (psProcess.waitFor() == 0) {
```
jtnord: need to read the input before you waitFor()

src/main/java/org/jenkinsci/test/acceptance/docker/DockerImage.java
```diff
@@ -92,45 +89,49 @@ public DockerImage(String tag) {
...
         } catch (ReflectiveOperationException e) {
             throw new Error(e);
         }
     }
 
-    private String waitForCid(CommandBuilder docker, File cidFile, File logfile, Process p) throws InterruptedException, IOException {
-        for (int i = 0; i < 10; i++) {
-            Thread.sleep(1000);
+    private String waitForCid(CommandBuilder docker, Process p) throws InterruptedException, IOException {
+        int exitVal = p.waitFor();
```
jtnord: need to read the input before you waitfor

jtnord:
@abayer - you should read the output from the command before you call `waitFor()`

Did my PR to your branch not work?

abayer:
Oops, missed that. Rolling back my change and merging in yours.

jtnord:
:bee: - but I feel dirty as I wrote some of the code

abayer:
It is dirty. Life is dirty. =)

armfergom:
🐝

abayer:
Ok, I'm merging!

PR #139, opened 2016-06-22
[JENKINS-36046] Correct parent path for credentials. 

armfergom:
Supersedes #133 

- [x] The initial reason to create #133 was that I thought the `path` for  controls in `FileCredentials` and `StringCredentials` were incorrect, but that is not the case. The problem was that after #123 , when adding a new credential the parent `path` with which it was created was empty. This PR fixes this and there is now no need to change the `FileCredentilas` and `StringCredentials` controls. 

- [x] Also, I have noticed that `selectDropdownMenuAlt` method (also introduced in #123) is making tests to be flaky. It performs several redundant clicks in the credentials type select that should not be performed and cause unexpected behaviour.

@olivergondza I have closed #133 since it is no longer relevant.

@reviewbybees esp @amuniz @recena @kwhetstone


abayer:
Looks good to me. :bee:

andresrc:
:bee:

PR #140, opened 2016-06-22
[FIXED JENKINS-35443] Look for form action, not title.

abayer:
https://issues.jenkins-ci.org/browse/JENKINS-35443

The title for either addUser or signup is always "Sign up" in 1.x,
at least in 1.651, but in 2.x, the title varies depending on whether
you're hitting the addUser endpoint or the signup endpoint. Since we
want the addUser endpoint, looking for a title of "Sign up" breaks on
2.x, while looking for a title of "Create user" (the one we get on the
right endpoint on 2.x) fails on 1.x.

So - instead, verify we're on the right page by looking for
"createAccountByAdmin" as the form action, since that's consistent and accurate.

### DIFF
--- a/src/main/groovy/org/jenkinsci/test/acceptance/po/users/AddUserPage.groovy
+++ b/src/main/groovy/org/jenkinsci/test/acceptance/po/users/AddUserPage.groovy
@@ -11,7 +11,7 @@ import org.kohsuke.randname.RandomNameGenerator
 class AddUserPage extends Page {
 
     static url = "securityRealm/addUser"
-    static at = { title == "Create User [Jenkins]" }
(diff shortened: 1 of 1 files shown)


olivergondza:
The rational behind the change makes sense to me.

andresrc:
:bee:

kwhetstone:
:bee: Looks fine to me.  This will definitely is a good example of something that would be caught by constantly running these tests.

abayer:
What this really took was me getting enough understanding of how the Java Selenium tests worked to be able to figure out from there how the Groovy Geb tests worked. =) Anyway, thanks!

PR #141, opened 2016-06-23
Jsch had issues connecting in some cases - fix sshd_config

abayer:
Adding the additional KexAlgorithms and MACs gets past an error with
protocol negotiation in some environments.

cc @reviewbybees


andresrc:
:bee:

armfergom:
🐝

PR #142, opened 2016-06-23
Don't visit Jenkins home page after scheduleBuild

Vlatombe:
Since we don't visit the build page anymore once the build is started,
it is not useful to go to the Jenkins home page. Especially since the
method getJenkins() is not fully implemented (doesn't work if there are
several Jenkins instance in injector)


Vlatombe:
@olivergondza does it look okay to you?

Vlatombe:
thanks!

PR #145, opened 2016-06-27
Make minimum browser window size configurable.

abayer:
Currently via environment variables.

We've had a lot of problems with scrolling to check
matrix/credentials/etc fields, which are...annoying. And fairly
pointless to work on, frankly. I'd strongly advocate for instead
allowing configurability of window geometry - I'd've thought that the
maximize call would do the trick on its own with a larger Xvnc
geometry, but it didn't, so I've made the browser height and width
configurable via environment variables.


olivergondza:
How about setting minimum screen resolution per ATH (or Jenkins itself) higher not to waste peoples time? If we are running into problems with such essentials use-cases, odds are more people is unhappy.

abayer:
Yeah, the more I'm thinking about it, the more a bigger resolution in general makes sense. Closing this, doing a new PR with a bigger window.

PR #146, opened 2016-06-27
Make minimum browser window size larger.

abayer:
We've had a lot of problems with scrolling to check
matrix/credentials/etc fields, which are...annoying. And fairly
pointless to work on, frankly. I'd strongly advocate for instead
allowing configurability of window geometry - I'd've thought that the
maximize call would do the trick on its own with a larger Xvnc
geometry, but it didn't - in fact, it seems to do absolutely *nothing*
when run inside Xvnc.

So instead, let's just suck it up and bump the minimum window size to
an admittedly silly-large 1680x1050. Selenium does fine with detecting
visibility of parts of the window that are off the screen, so a
smaller screen than 1680x1050 causes no problems.

I'd be open to a smaller width too, but definitely bigger than
1280x960, the previous minimum.

This replaces #145.

As always, cc @reviewbybee and @olivergondza.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/FallbackConfig.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/FallbackConfig.java
@@ -166,12 +166,11 @@ public class FallbackConfig extends AbstractModule {
     public WebDriver createWebDriver(TestCleaner cleaner, TestName testName, ElasticTime time) throws IOException {
         WebDriver base = createWebDriver(testName);
 
-        // Make sue the window have minimal resolution set, even when out of the visible screen. Try maximizing first so
-        // it has a chance to fit the screen nicely if big enough.
-        base.manage().window().maximize();
+        // Make sue the window have minimal resolution set, even when out of the visible screen.
+        // Note - not maximizing here any more because that doesn't do anything.
         Dimension oldSize = base.manage().window().getSize();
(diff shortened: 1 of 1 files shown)


abayer:
Some open JIRAs this seems to resolve:

- [JENKINS-34690](https://issues.jenkins-ci.org/browse/JENKINS-34690)
- [JENKINS-36042](https://issues.jenkins-ci.org/browse/JENKINS-36042)
- [JENKINS-34411](https://issues.jenkins-ci.org/browse/JENKINS-34411)

recena:
I'm not sure about this proposal. Today we need a size, tomorrow another one, etc...

I'm investigating a different approach based on remove all scrooll bars from containers in order to get all element visible. Let me try it.

abayer:
In the case of authorization, there aren't any scroll bars on the box - it's just a ridiculously wide box. =)

abayer:
Another one that works with this but doesn't with master - [JENKINS-31738](https://issues.jenkins-ci.org/browse/JENKINS-31738) which...makes no sense to me at first glance, but there you are.

abayer:
whoops, cc @reviewbybees - typo!

mafraba:
LGTM at least until we find a definitive solution 🐝

andresrc:
:bee:

kwhetstone:
:bee: with objections; hopefully we'll find something that can keep us ahead of the UI eventually.

PR #148, opened 2016-07-01
Properly wait for dynamically generated blocks to appear in UI

olivergondza:
Converted some of the places, lot more left to do. Let's see if CI build spot some problem.


src/main/java/org/jenkinsci/test/acceptance/plugins/credentials/CredentialsPage.java
```diff
@@ -30,7 +27,7 @@ public CredentialsPage(Jenkins j, String domainName) {
         addButton.selectDropdownMenuAlt(type);
         String path = find(by.name("credentials")).getAttribute("path");
 
-        return newInstance(type, this, path);
+        return newInstance(type, this, "/credentials");
```
raul-arabaolaza: Should we use the previous path variable instead of a hardcoded one??

raul-arabaolaza:
I really really like this PR. Can I help with something??

olivergondza:
@raul-arabaolaza, I am in a phase of polishing the failures detected by PR builder. I do not expect any drastic changes - it just gets slow waiting for the results. Reviews and testing appreciated.

olivergondza:
There do not seem to be any failures caused by this PR. I will merge it after #207, that has a potential to resolve a lot of the failing tests.

PR #151, opened 2016-07-05
[JENKINS-34614] Fix for ftpd container connectivity problems

mafraba:
Fix for [JENKINS-34614](https://issues.jenkins-ci.org/browse/JENKINS-34614).

Several factors were preventing the connection to the ftp server:

1. Due to how FTP protocol works ([details here](http://www.slacksite.com/other/ftp.html)) we need an exact match of container and host ports. If anyone knows another solution please comment.
2. We needed to force the 'passive' ip/host in the ftp server.

My proposed solution to address (1) is to add a new boolean property `matchHostPorts` in the `DockerFixture` annotation, which when true will force an exact port mapping between docker container and host.

Also modified the `FtpdContainer`'s `Dockerfile` to fix the ubuntu version, and to address (2).

@reviewbybees esp @abayer, @olivergondza

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/docker/DockerContainerHolder.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/docker/DockerContainerHolder.java
@@ -51,7 +51,9 @@ public class DockerContainerHolder<T extends DockerContainer> implements Provide
             File runlog = diag.touch("docker-" + fixture.getSimpleName() + ".run.log");
             try {
                 Starter<T> containerStarter = docker.build(fixture, buildlog).start(fixture);
-                if (portOffset != null) containerStarter.withPortOffset(portOffset);
+                if (portOffset != null) {
+                    containerStarter.withPortOffset(portOffset);
+                }
                 container = containerStarter.withLog(runlog).start();
             } catch (InterruptedException | IOException e) {
                 throw new Error("Failed to start container - " + fixture.getName(), e);
--- a/src/main/java/org/jenkinsci/test/acceptance/docker/DockerContainerHolder.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/docker/DockerContainerHolder.java
@@ -11,6 +11,8 @@ import javax.inject.Named;
 import javax.inject.Provider;
 import java.io.File;
 import java.io.IOException;
+import org.jenkinsci.test.acceptance.docker.DockerImage.Starter;
+import org.jenkinsci.test.acceptance.docker.fixtures.FtpdContainer;
 
 /**
  * Inject this object to automate the cleanup of a running container at the end of the test case.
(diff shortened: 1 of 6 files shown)


src/main/java/org/jenkinsci/test/acceptance/docker/DockerContainerHolder.java
```diff
@@ -36,7 +38,7 @@
      */
     @Inject(optional = true)
     @Named("dockerPortOffset")
-    private int portOffset = 0;
+    private Integer portOffset;
```
recena: Why not `int`?

src/main/java/org/jenkinsci/test/acceptance/docker/DockerContainerHolder.java
```diff
@@ -48,7 +50,9 @@ public synchronized T get() {
             File buildlog = diag.touch("docker-" + fixture.getSimpleName() + ".build.log");
             File runlog = diag.touch("docker-" + fixture.getSimpleName() + ".run.log");
             try {
-                container = docker.build(fixture, buildlog).start(fixture).withPortOffset(portOffset).withLog(runlog).start();
+                Starter<T> containerStarter = docker.build(fixture, buildlog).start(fixture);
+                if (portOffset != null) containerStarter.withPortOffset(portOffset);
```
recena: Better

```
if (portOffset != null) {
    containerStarter.withPortOffset(portOffset);
}
```

recena:
LGTM, I assume you tried it 😄

src/main/java/org/jenkinsci/test/acceptance/docker/DockerContainerHolder.java
mafraba: This way the semantic is clearer: we can use no port offset (null, which is default vale), or we can set an offset of 0, which means exact mapping. 

I leverage this to implement the `matchHostPorts` property in the docker fixture.

recena: @mafraba Indeed `port 0` means:

```
/* 
 * Obtain a reference to a local port for the given sock, if snum is zero it means select any available local port.
 */
```

Negative ports are not valid. Maybe you can use `-1`?

mafraba:
Strange that I didn't get those errors while testing. Plus, they're the ports already in use in the tests, I didn't change them. Did these tests work at some point in the past?

Will do anyway. Thanks.

abayer:
I have no idea if the tests ever worked. =)

abayer:
Will test latest commit later today.

abayer:
:bee: All 17 tests passed with this change.

src/main/java/org/jenkinsci/test/acceptance/docker/DockerContainerHolder.java
recena: From Linux kernel.

You don't need a `Integer` here.

recena:
🐝 since this has been discussed offline

mafraba:
@reviewbybees done

PR #152, opened 2016-07-06
[JENKINS-36469] CapybaraPortingLayerImpl.clickButton() has problems with vertical scroll and sticky buttons

raul-arabaolaza:
[JENKINS-36469](https://issues.jenkins-ci.org/browse/JENKINS-36469)

In the case the button is sticky this PR does not scroll before clicking it, in other case the functionality is the existing one

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
@@ -22,6 +22,7 @@ import org.openqa.selenium.WebElement;
 
 import com.google.common.base.Joiner;
 import com.google.inject.Injector;
+import org.openqa.selenium.interactions.Actions;
 
 import static java.util.Arrays.*;
 
@@ -82,7 +83,23 @@ public class CapybaraPortingLayerImpl implements CapybaraPortingLayer {
     @Override
     public void clickButton(String text) {
         WebElement e = find(by.button(text));
-        e.click();
+        /*
+         * YUI sticky buttons present some problems when scroll to them, also if you have a sticky button
+         * you should not need to scroll to use them
+         */
+        boolean isStickyButton = false;
+        WebElement stickyContainer = getElement(by.id("bottom-sticker"));
+        if (stickyContainer != null) {
+            JavascriptExecutor je = (JavascriptExecutor)driver;
+            isStickyButton = (boolean)je.executeScript("return arguments[0].contains(arguments[1])", stickyContainer, e);
+        }
+        if (isStickyButton) {
+            Actions builder = new Actions(driver);
+            builder.moveToElement(e).click(e);
+            builder.perform();
+        } else {
+            e.click();
+        }
     }
 
     /**


mafraba:
🐝 

Anyway, as a suggestion: wouldn't it be a good idea to put this in the `Scroller` class instead? That way, all scrolling-related weirdness would be encapsulated there.

raul-arabaolaza:
@mafraba {{Scroller}} affects the scroll of all clickable elements, this issue is only related to clicking buttons, so I believe the "sticky button logic" logic has not to be applied to href, radios and other input clicks. And so the right place is not the very low level {{Scroller}} class but the high level clickButton method. But not hard feelings, so if any other thinks that is a good idea I will change the PR

raul-arabaolaza:
@reviewbybees done

andresrc:
:bee:

src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
```diff
@@ -82,7 +83,23 @@ protected final WebDriver visit(URL url) {
...
-        e.click();
+        /*
+         * YUI sticky buttons present some problems when scroll to them, also if you have a sticky button
+         * you should not need to scroll to use them
+         */
+        boolean isStickyButton = false;
+        WebElement stickyContainer = getElement(by.id("bottom-sticker"));
+        if (stickyContainer != null) {
+            JavascriptExecutor je = (JavascriptExecutor)driver;
+            isStickyButton = (boolean)je.executeScript("return arguments[0].contains(arguments[1])", stickyContainer, e);
```
olivergondza: `CapybaraPortingLayer#executeScript(String, Object...)`

src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
```diff
@@ -82,7 +83,23 @@ protected final WebDriver visit(URL url) {
...
+        }
+        if (isStickyButton) {
+            Actions builder = new Actions(driver);
+            builder.moveToElement(e).click(e);
```
olivergondza: Does this works better than `Scroller` JS implementation? Will it hurt if we use it all the time?

src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
raul-arabaolaza: @olivergondza Thanks!! I will make a Follow up PR to implement this suggestion

src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
raul-arabaolaza: @olivergondza Yes on sticky buttons, on other elements that actually need scrolling I have not tested, but I have created [JENKINS-36551](https://issues.jenkins-ci.org/browse/JENKINS-36551) and I will make a WIP PR to check how does the {{Scroller}} class performs using this method instead of scrollerjs

PR #153, opened 2016-07-06
[JENKINS-31738] Non-deterministic behavior in PrioritySorterPluginTest#match_jobs_by_view

raul-arabaolaza:
[JENKINS-31738](https://issues.jenkins-ci.org/browse/JENKINS-31738) 

Downstream of #152 

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Node.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Node.java
@@ -44,6 +44,8 @@ public abstract class Node extends ContainerPageObject {
     public abstract String getName();
 
     public void setExecutors(int n) {
+        // On high Res screens the UI redraws, so best wait for numExecutors to be present
+        waitFor(by.path("/numExecutors"));
         control("/numExecutors").set(n);
     }
 
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Slave.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Slave.java
@@ -70,6 +70,8 @@ public class Slave extends Node {
         return new Matcher<Slave>("slave run build in order: %s", Joiner.on(' ').join(jobs)) {
             @Override public boolean matchesSafely(Slave slave) {
                 slave.visit("builds");
+                //Jobs table may take a little to be populated, give it some time
+                slave.elasticSleep(2000);
                 String list = slave.find(by.id("projectStatus")).getText();
 
                 StringBuilder sb = new StringBuilder(".*");


recena:
🐝

raul-arabaolaza:
Seems there is a problem when executed on CI.

raul-arabaolaza:
Rebased to master since #152 has been merged

src/main/java/org/jenkinsci/test/acceptance/po/Node.java
```diff
@@ -44,6 +44,8 @@ protected Node(Injector i, URL url) {
     public abstract String getName();
 
     public void setExecutors(int n) {
+        // On high Res screens the UI redraws, so best wait for numExecutors to be present
+        waitFor(by.path("/numExecutors"));
```
olivergondza: I would like to avoid doing this for every control we interact with. How comes it does not blow up on the next line?

raul-arabaolaza: @olivergondza Not for every control, most of the time when the containing page is loaded the control is ready to use and no need for wait exists, but in this concrete case the page redraws itself after load to adjust to size, with the result that sometimes 'control("/numExecutors").set(n)' was failing due to 'NoSuchElement' exceptions. Using ẃaitFor' was only a way to give the page time to redraw.

olivergondza: But this can affect any control on any config form, no? Also, are we sure redrawing is the cause - does it remove/hide the element temporarily?

raul-arabaolaza: I can't be sure about the rest of the pages, I have not seen any fails because of redrawing. But in the case of this test, yes. Redrawing were hiding the elements temporarily, I have seen waits (in the form of elasticSleep) on other tests to ensure JS generation is finished before interacting with the page, but personally I prefer to use `waitFor` instead of `elasticSleep` to not delay the tests more than needed. Also is more than probably that this kind of fail happends only on HD displays (I was unable to get any fail until my new laptop arrived) making it not very common and undetected by the CI server, but a problem is a problem and should be fixed.

raul-arabaolaza:
After merging upstream PR the problems continue, so for the moment this PR is WIP

raul-arabaolaza:
So, after the last commit the Test has worked for 2 consecutive times on CI, making it as available for review

rsandell:
:bee:

recena:
🐝

raul-arabaolaza:
@reviewbybees done

PR #155, opened 2016-07-07
[JENKINS-35644] Fix JiraPluginTest

mafraba:
[JENKINS-35644](https://issues.jenkins-ci.org/browse/JENKINS-35644)

Fix for JiraPluginTest. The problem is that the fixture was using always the latest JIRA version available, but from 7.0 the JIRA SOAP API was gone.

The test set up was trying to interact with the JIRA through this missing API, hence the failure. I've pinned down the docker fixture to 6.3, and the test now succeeds.

@reviewbybees

### DIFF
--- a/src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/JiraContainer/Dockerfile
+++ b/src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/JiraContainer/Dockerfile
@@ -2,8 +2,14 @@
 # Runs JIRA
 #
 #    The initial password is 'admin:admin'
-# 
+#
 FROM ubuntu:xenial
+
+# Pin JIRA version to make the tests more predictable and less fragile
+# In particular, pinned to 6.X because from 7.X the SOAP API is gone, and it's
+# used in the tests. Upgrading the tests to use the REST API is not trivial
+# since some operations are not yet available in its java client (e.g. project creation)
+ENV JIRA_VERSION 6.3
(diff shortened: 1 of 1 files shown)


armfergom:
@mafraba good catch!! However, I think the fix should be to make the `JiraContainer` class to use the REST API instead of the [deprecated](https://wiki.jenkins-ci.org/display/JENKINS/JIRA+Plugin#JIRAPlugin-UsingJIRARPCSOAP%28deprecated%29) SOAP.

mafraba:
I agree @armfergom , and that was my initial approach (see branch [JENKINS-35644-WIP-REST](https://github.com/mafraba/acceptance-test-harness/tree/JENKINS-35644-WIP-REST)).

But there are some operations not available yet in the REST API, or at least in the java client. One of them is the creation of a project, which we need in the test. That makes migrating to the new API not-so-trivial. 

Definitely doable, but given that the test still serves its purpose without upgrading to the last JIRA version, I think we can use this solution until the jira rest java client includes the missing operations.

If at some point the Jira plugin does not work anymore against old JIRA versions:

1. We will notice because this test will fail
2. We will need to update the test (and hopefully it will be trivial if the jira rest client is completed by then)

Anyway, if still there is consensus that we should use the REST API from the test, I'm more than happy to spend some more time on this issue and take that approach.

armfergom:
fair enough, but it'd be great to revisit this when the functionality is available in the REST API.

🐝

PR #156, opened 2016-07-08
[JENKINS-31060] - AntPluginTest does not work if ANT_HOME set

mafraba:
[JENKINS-31060](https://issues.jenkins-ci.org/browse/JENKINS-31060)

Proposal to bring back the test previously removed, with some modifications to make it work.

The original error was due mainly to:

1. an inconsistent setup of PATH and ANT_HOME env variables,
2. how ant is "too smart" when looking for an "ANT_HOME" variable
3. how we are also being smart enough at setting up a fake home for tool installations (see [ToolInstallation.java](https://github.com/jenkinsci/acceptance-test-harness/blob/c1512e4da9dc3b4dd701988dae103eaec405ec25/src/main/java/org/jenkinsci/test/acceptance/po/ToolInstallation.java#L123)).

In this test "the user" is entering the ANT_HOME (fake home) which jenkins must use for the build. I think that if the test uses this exact same ant script to set the expected version, the test both makes more sense and will be tolerant to the wrong set up which caused the issue.

@reviewbybees , esp @abayer

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/ant/AntInstallation.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/ant/AntInstallation.java
@@ -39,7 +39,9 @@ public class AntInstallation extends ToolInstallation {
         super(context, path);
     }
 
-    public void useNative() {
-        installedIn(fakeHome("ant", "ANT_HOME"));
+    public String useNative() {
+        String antHome = fakeHome("ant", "ANT_HOME");
+        installedIn(antHome);
+        return antHome;
     }
 }
--- a/src/test/java/plugins/AntPluginTest.java
+++ b/src/test/java/plugins/AntPluginTest.java
@@ -63,6 +63,25 @@ public class AntPluginTest extends AbstractJUnitTest {
         );
     }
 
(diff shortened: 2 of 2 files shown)


olivergondza:
:+1:

recena:
🐝

src/main/java/org/jenkinsci/test/acceptance/plugins/ant/AntInstallation.java
```diff
@@ -39,7 +39,9 @@ public AntInstallation(Jenkins context, String path) {
         super(context, path);
     }
 
-    public void useNative() {
-        installedIn(fakeHome("ant", "ANT_HOME"));
+    public String useNative() {
```
oleg-nenashev: 🐜 such compatibility breakage here is fine, right?

recena: @oleg-nenashev This is something install to the test scope. I don't see the problem whether it was checked.

mafraba:
Follow up PR: https://github.com/jenkinsci/acceptance-test-harness/pull/158

PR #157, opened 2016-07-08
Add a Jenkinsfile for parallelized ATH runs via Pipeline

abayer:
Note - this depends on https://github.com/jenkinsci/jenkins-project-pipeline-lib/pull/1.

This is descended from how we've been doing ATH runs internally at CloudBees - loading most of the logic from Pipeline libraries, splitting the ATH into a number of separate parallel blocks using the Parallel Test Executor plugin, and then pulling all the results back together in the end. 

The plan is to start using this for the non-PR-triggered ATH builds on jenkins.ci.cloudbees.com, with the PR builds moved out to a separate, still linear job for now. Until [JENKINS-33274](https://issues.jenkins-ci.org/browse/JENKINS-33274) is resolved, multibranch jobs will have to run all the tests serially for the first build on each branch or PR anyway, but the main reason is that we'll simply exhaust our executor resources on jenkins.ci.cloudbees.com with PR builds. Down the road, once we've got [JENKINS-30412](https://issues.jenkins-ci.org/browse/JENKINS-30412), we should hopefully be able to get some logic in place to figure out what tests have changed/are relevant to a given PR so that we can avoid having to run _all_ the tests for every PR.

cc @jenkinsci/code-reviewers, @reviewbybees, @olivergondza, @jtnord

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,50 @@
+/*
+ * This Jenkinsfile is intended to run on https://jenkins.ci.cloudbees.com and may fail anywhere else.
+ * It makes assumptions about plugins being installed, labels mapping to nodes that can build what is needed, etc.
+ *
+ * The required label is "hi-speed" */
+
+// Only keep the 50 most recent builds.
+properties([[$class: 'jenkins.model.BuildDiscarderProperty', strategy: [$class: 'LogRotator',
+                                                                        numToKeepStr: '50']]])
+
+loadParameterizedLibs { functions, ath ->
+
+    parallel 'jenkins-war': { functions.stashJenkinsWar(war, athLabel(), true) },
+        'get-ath': { stashAth(functions) },
+        failFast: true
+
+
+}
+
+def loadParameterizedLibs(def body) {
+    String functionsRepo = FUNCTIONS_REPO
+    String functionsBranch = FUNCTIONS_BRANCH
+    String label = FUNCTIONS_LABEL
+    loadLibs(label, functionsRepo, functionsBranch, body)
+}
+
+def loadLibs(String label = null,
+             String functionsRepo = "jenkinsci/jenkins-project-pipeline-lib",
+             String functionsBranch = "master",
+             def body) {
+    if (label == null) {
+        label = athLabel()
+    }
+
+    if (!functionsRepo.endsWith(".git")) {
+        functionsRepo = functionsRepo + ".git"
+    }
+    
+    def functions
+    def ath
+
+    node(label) {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,110 @@
...
+    }
+    
+    def functions
+    def ath
+
+    node(label) {
+        git changelog: false, poll: false, url: "git://github.com/" + functionsRepo, branch: functionsBranch
+
+        functions = load 'lib/functions.groovy'
+        ath = load 'lib/ath.groovy'
```
olivergondza: Is there any value in sharing `ath.groovy`? I doubt anybody would use it without ATH sources.

abayer: Main reasoning is so we can eventually branch the ATH without also forking the build tooling, if that makes any sense.

Jenkinsfile
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * The required label is "hi-speed" */
+
+// TODO: Figure out how to run just changed/affected-by-changes tests when running from a PR.
+// TODO: Get https://issues.jenkins-ci.org/browse/JENKINS-33274 fixed and configured so that we don't have to run
+//   serially for the first run on a branch.
+
+timestamps {
+    // Only keep the 50 most recent builds.
+    properties([[$class: 'jenkins.model.BuildDiscarderProperty', strategy: [$class: 'LogRotator',
```
jglick: Should be updated to new symbols.

Jenkinsfile
```diff
@@ -0,0 +1,110 @@
...
+}
+
+/**
+ * Loads the libraries using known parameters for repo/branch/node if they're specified, and then makes the libraries
+ *   available in the execution of the body.
+ *
+ * @param body Closure to execute.
+ */
+def loadLibs(def body) {
+    String functionsRepo = defaultValueIfNotSet("FUNCTIONS_REPO", "jenkinsci/jenkins-project-pipeline-lib")
```
jglick: Note that run-time variation of a shared library is not currently supported in JENKINS-31155; would require a `library` step, which would not work with static typing. At any rate I do not see where this variable even could be defined.

jglick: Anyway the _Replay_ function lets you play with library changes more easily than what I am guessing this hook was for.

stephenc:
@abayer what is the status of this?

abayer:
I need to resurrect this. Will try to do so next week.

abayer:
Closing this.

jglick:
Sad, no plans for this?

olivergondza:
Once I will get the docker tests running in https://github.com/jenkinsci/acceptance-test-harness/pull/238, parallel test execution is the next thing to look at.

PR #161, opened 2016-07-20
[JENKINS-36818] Add the ability to delete a project

raul-arabaolaza:
[JENKINS-36818](https://issues.jenkins-ci.org/browse/JENKINS-36818) This could be useful to implement some tests

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Job.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Job.java
@@ -27,6 +27,7 @@ import org.jenkinsci.test.acceptance.controller.LocalController;
 import org.jenkinsci.test.acceptance.junit.Resource;
 import org.junit.internal.AssumptionViolatedException;
 import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.ui.ExpectedConditions;
 import org.zeroturnaround.zip.ZipUtil;
 
 import com.google.inject.Injector;
@@ -415,8 +416,6 @@ public class Job extends TopLevelItem {
     public void delete() {
         this.open();
         clickLink("Delete Project");
-        // We need to wait a little for the alert to appear
-        elasticSleep(500);
-        confirmAlert();
+        confirmAlert(2);
     }
 }
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Job.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Job.java
@@ -408,4 +408,15 @@ public class Job extends TopLevelItem {
 
         return links;
     }
+
+    /**
+     * Deletes the current job
+     */
+    public void delete() {
+        this.open();
+        clickLink("Delete Project");
+        // We need to wait a little for the alert to appear
+        elasticSleep(500);
+        confirmAlert();
+    }
 }
--- a/src/test/java/core/DeleteJobTest.java
+++ b/src/test/java/core/DeleteJobTest.java
@@ -1,0 +1,33 @@
+package core;
+
+import org.jenkinsci.test.acceptance.junit.AbstractJUnitTest;
+import org.jenkinsci.test.acceptance.po.FreeStyleJob;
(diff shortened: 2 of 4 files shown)


raul-arabaolaza:
@reviewbybees

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
```diff
@@ -408,4 +408,15 @@ public ScmPolling pollScm() {
...
     }
+
+    /**
+     * Deletes the current job
+     */
+    public void delete() {
+        this.open();
+        clickLink("Delete Project");
+        // We need to wait a little for the alert to appear
+        elasticSleep(500);
```
andresrc: Given that alert [throws](http://seleniumhq.github.io/selenium/docs/api/java/org/openqa/selenium/WebDriver.TargetLocator.html#alert--) an exception if there's no alert, could we go for a "wait until" (with timeout) kind of waiting.

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
```diff
@@ -408,4 +408,15 @@ public ScmPolling pollScm() {
 
         return links;
     }
+
+    /**
+     * Deletes the current job
+     */
+    public void delete() {
+        this.open();
+        clickLink("Delete Project");
```
andresrc: May this label change for different kind of jobs?

kwhetstone: Or it could change in the future.  Maybe parametertize it out or override with a delete(String label) method?

olivergondza: It is responsibility of given job page object to know that - I see no reason to make the test responsible for choosing the correct label. I am fine leaving it as it is and override/rework when needed.

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
olivergondza: We should, IIRC selenium have a waiting condition for alert to pop up.

src/test/java/core/DeleteJobTest.java
```diff
@@ -0,0 +1,33 @@
...
+     * Then the job details page should not exist
+     */
+    @Test
+    public void delete_a_simple_job() {
+        FreeStyleJob j = jenkins.jobs.create(FreeStyleJob.class, "simple-job");
+        assertTrue(j.open().getTitle().contains(j.name));
+
+        j.delete();
+
+        assertFalse(j.open().getTitle().contains(j.name));
```
olivergondza: I prefer asserting the page object does not exists (`Matchers#pageObjectExists`).

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
raul-arabaolaza: @andresrc @kwhetstone I agree with @olivergondza here, I was in a similar situation where a `FreeStyleMultiBranchProject` changed the usual UI and `Job.addShellStep()' were not working, overriding it in  `FreeStyleMultiBranchProject` was very easy

src/test/java/core/DeleteJobTest.java
raul-arabaolaza: @olivergondza Changed, thanks

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
raul-arabaolaza: @andresrc @olivergondza Agree, Selenium provides `ExpectedConditions.isAlertPresent()` which I can use with `WebDriverWait`

raul-arabaolaza:
@andresrc @kwhetstone  @olivergondza I have implemented some changes. Could you please check them??

olivergondza:
:+1:

kwhetstone:
:bee: looks good to me

PR #162, opened 2016-07-28
[JENKINS-36928] Job PO does not accurately reflects the real job underlying state

raul-arabaolaza:
[JENKINS-36928](https://issues.jenkins-ci.org/browse/JENKINS-36928)

Instead of making a Job always aware of the Build Parameters I have decided to add the ability for the job to load them on demand to minimize impact on existing code. This way only tests that need that need to be modified and the general contract of Job is simply extended instead of modified. So with this PR a Job can load it's existing Build Parameters as follow:

``` @Test
    public void copy_a_simple_job_loads_build_params() {
        FreeStyleJob j = jenkins.jobs.create(FreeStyleJob.class, "simple-job");
        j.configure();
        j.addParameter(StringParameter.class).setName("Param1").setDefault("");
        j.addParameter(StringParameter.class).setName("Param2").setDefault("");
        j.save();
        jenkins.jobs.copy(j, "simple-job-copy");
        assertThat(driver, hasContent("simple-job-copy"));

        FreeStyleJob k = jenkins.jobs.get(FreeStyleJob.class, "simple-job-copy");
        k.loadExistingBuildParameters(StringParameter.class);
        j.visit("config.xml");
        String jxml = driver.getPageSource();

        k.visit("config.xml");
        String kxml = driver.getPageSource();

        assertThat(jxml, is(kxml));
        assertTrue(k.getParameters().size() == j.getParameters().size());
    }
```

Any suggestion/complaint about this approach is more than welcome, especially in the case of specializing the parameter load for custom Job types.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Parameter.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Parameter.java
@@ -2,20 +2,30 @@ package org.jenkinsci.test.acceptance.po;
 
 import org.openqa.selenium.By;
 
+import java.util.LinkedList;
+import java.util.List;
+
 /**
  * Parameter for builds.
  * <p/>
- * Use {@link Describable} annotation to register an implementation.
+ * Use {@link Describable} annotation to register an implementation. The value of {@link Describable} must be the
+ * corresponding caption of the add parameter menu item AND the corresponding {@link hudson.model.ParameterDefinition}
+ * class. For example {"String Parameter", "hudson.model.StringParameterDefinition"}
  *
  * @author Kohsuke Kawaguchi
  */
 public abstract class Parameter extends PageAreaImpl {
+    private static List<Class<? extends Parameter>> subtypes = new LinkedList<>();
+
     public final Job job;
     private String name;
 
     public Parameter(Job job, String path) {
         super(job, path);
         this.job = job;
+        if (!Parameter.subtypes.contains(this.getClass())) {
+            Parameter.subtypes.add(this.getClass());
+        }
     }
 
     public String getName() {
@@ -63,5 +73,9 @@ public abstract class Parameter extends PageAreaImpl {
         String np = driver.findElement(by.xpath("//input[@name='name' and @value='%s']", name)).getAttribute("path");
         String path = np.replaceAll("/name$", "") + "/" + rel;
         return by.path(path);
+    }
+
+    public static List<Class<? extends Parameter>> all() {
+        return Parameter.subtypes;
     }
 }
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Parameter.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Parameter.java
@@ -28,6 +28,13 @@ public abstract class Parameter extends PageAreaImpl {
         return this;
     }
 
+    /**
+     * Sets this objects name property without modifying the UI
+     */
+    void setNameProperty(String name) {
+        this.name = name;
+    }
+
     public Parameter setDescription(String v) {
         control("description").set(v);
         return this;
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Parameter.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Parameter.java
@@ -14,6 +14,7 @@ import java.util.List;
  *
  * @author Kohsuke Kawaguchi
  */
+//TODO remove the caption value from @Describable once the addParameter method can ignore it
 public abstract class Parameter extends PageAreaImpl {
     private static List<Class<? extends Parameter>> subtypes = new LinkedList<>();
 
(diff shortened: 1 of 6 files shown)


olivergondza:
@raul-arabaolaza, Why don't we just issue another json request from `getParameters()`. This seems to introduce completely new patterns and expose some nasty details to clients.

raul-arabaolaza:
@olivergondza That was my first idea, unfortunately (or logically) the data stored in the job configuration includes the build parameters as instances of `hudson.model.ParameterDefinition` (see below)
```
{
          "defaultParameterValue" : {
            "name" : "nanana",
            "value" : "1"
          },
          "description" : "ewfwefwf",
          "name" : "nanana",
          "type" : "StringParameterDefinition"
        }
```
And the ATH uses as parameter abstraction instances of class `org.jenkinsci.test.acceptance.po.Parameter` which as `@Describable` value use the UI label of the select parameter combo in the UI. So the problem is that I have not found any solid (apart on relying on name conventions which is too weak for my tastes) way to relate `hudson.model.ParameterDefinition` classes like `StringParameterDefinition` with the `org.jenkins.test.acceptance.po.Parameter` ones used by the ATH without making a big change on the existing classes contract. Also you can not use any info from the UI because all descriptorId for parameters referer to  `hudson.model.ParameterDefinition` classes.

IMHO the best way would be to include in the `@Describable` value of the `org.jenkins.test.acceptance.po.Parameter` subclasses used by the ATH the corresponding `hudson.model.ParameterDefinition` class. But that implies a change in the contract of `org.jenkins.test.acceptance.po.Parameter` with, potentially, a lot of repercussions in the entire ATH and honestly I believe I am not in position to request such a change by myself.

So this approach is strange and very far from perfect, but I believe it has minimum impact on existing code and could potentially be easily replaced if a better solution is proposed and approved

olivergondza:
> IMHO the best way would be to include in the @Describable value of the org.jenkins.test.acceptance.po.Parameter subclasses used by the ATH the corresponding hudson.model.ParameterDefinition class. But that implies a change in the contract of org.jenkins.test.acceptance.po.Parameter with, potentially, a lot of repercussions in the entire ATH and honestly I believe I am not in position to request such a change by myself.

IIRC Jesse have done similar change in the past for one of the page objects and he seem to feal strongly to prefer Jenkins class names over textual descriptions. I lean towards the same direction after considering the problem you have described. So if this is the only reason not to do what is straightforward, we should go and change `@Describable`semantics for parameters.

raul-arabaolaza:
@olivergondza Great! If you guys think is a good idea I am heartily on it. One last question before changing it, even with the change in `@Describable` semantics I believe I still need classpath scanning or to know the available  `org.jenkins.test.acceptance.po.Parameter` classes, I am thinking in adding an `all()` method to `org.jenkins.test.acceptance.po.Parameter` class and to change the constructor for all subclasses to register themselves. Sounds good to you??

PS. Thanks for your time

raul-arabaolaza:
Never mind, for a Job in the ATH to have Build params you nee to invoke the addParam method and instantiate the desired Parameter that means a static list and some code in the constructor suffice...

raul-arabaolaza:
@olivergondza Changes done, although I have not been able to only use the JSON data because to create `org.jenkins.test.acceptance.po.Parameter` objects I need the path from the UI, but the code is much simpler now and no strange patterns or internals are shown to the user.

@reviewbybees

src/main/java/org/jenkinsci/test/acceptance/po/Parameter.java
```diff
@@ -2,20 +2,30 @@
...
+import java.util.LinkedList;
+import java.util.List;
+
 /**
  * Parameter for builds.
  * <p/>
- * Use {@link Describable} annotation to register an implementation.
+ * Use {@link Describable} annotation to register an implementation. The value of {@link Describable} must be the
+ * corresponding caption of the add parameter menu item AND the corresponding {@link hudson.model.ParameterDefinition}
+ * class. For example {"String Parameter", "hudson.model.StringParameterDefinition"}
```
olivergondza: You have preserved the text description to kep it working in `Job#addParameter()`, right? Please add TODO get rid of that.

raul-arabaolaza: Done, should I fill an Issue on jenkins core to request the menu items in add parameter to have a descriptorId??

raul-arabaolaza:
@reviewbybees

raul-arabaolaza:
@olivergondza A quick ack so it can be merged please??

olivergondza:
I wonder how other people feel about this. There are things I do not like (extracting PO from code, 2 arguments for `@Describable`). Though, I do not see a better way to implement this, so I lean towards to merge this.

raul-arabaolaza:
@olivergondza I agree with you, this is just what we can do for the moment.

For discussion sake IMHO PO should be a way to send commands to Jenkins (create a job, schedule a build, etc) and in that sense they must be unidirectional, but the problem is that sometimes we need to query the jenkins state to properly implement that commands and for the moment the only thing that exists in the ATH are the PO itselves... maybe we need another set of abstractions apart from PO to query the jenkins state and use them when necessary in our PO (for example when scheduling a job i need to query the job for possible build parameters)

raul-arabaolaza:
@olivergondza Any news for this? Is blocking me on other sides and it would be great to know if I can merge it or need to go with a workaround for the moment and somehow "ease" this problem with some PR to jenkins itself to remove the need of the "strange" things of this PR

raul-arabaolaza:
@jglick If you have time, can you give an opinion here? I would like to advance or definitively close this

PR #163, opened 2016-08-02
[JENKINS-37111] Change update center info update

andresrc:
See [JENKINS-37111](https://issues.jenkins-ci.org/browse/JENKINS-37111)

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
@@ -88,10 +88,10 @@ public class PluginManager extends ContainerPageObject {
                 try {
                     button.findElement(by.id("it does not matter"));
                 } catch(StaleElementReferenceException e) {
-                    return true; // we are in another page now
+                    return Boolean.TRUE; // we are in another page now
                 } catch(Exception e) {
                 }
-                return false;
+                return Boolean.FALSE;
             }
         });
         updated = true;
--- a/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
@@ -79,6 +79,7 @@ public class PluginManager extends ContainerPageObject {
      */
     public void checkForUpdates() {
         visit("advanced");
+        final String current = getCurrentUrl();
         // The check now button is a form submit (POST) with a redirect to the same page only if the check is successful.
         // We use the button itself to detect when the page has changed, which happens after the refresh has been done
         // And we check for the presence of the button again
@@ -94,8 +95,9 @@ public class PluginManager extends ContainerPageObject {
                     } catch(StaleElementReferenceException e) {
                         // with this exception we know we've left the original page
                         // we look for an element in the page to check for success
-                        find(by.button("Check now"));
-                        return true;
+                        if (current.equals(getCurrentUrl())) {
+                            return true;
+                        }
                     }
                 } catch(Exception e) {
                 }
--- a/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
@@ -14,6 +14,7 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Logger;
 
+import com.google.common.base.Predicate;
 import org.apache.commons.lang3.StringUtils;
 import org.eclipse.aether.resolution.ArtifactResolutionException;
 import org.jenkinsci.test.acceptance.junit.WithPlugins;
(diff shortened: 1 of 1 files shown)


mafraba:
🐝

amuniz:
:bee:

src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
```diff
@@ -76,8 +76,8 @@ public PluginManager(Jenkins jenkins) {
      * Force update the plugin update center metadata.
      */
     public void checkForUpdates() {
-        visit("checkUpdates");
-        waitFor(by.xpath("//span[@id='completionMarker' and text()='Done']"));
+        visit("advanced");
+        clickButton("Check now");
```
jtnord: Why do you no longer wait for the check to happen - it may not be instant - and smells like a :bug:

jtnord:
:bug: inline IIUC

andresrc:
Ups, I was going to do the POST request directly (which returns after completion) but changed to using the button afterwards. Fixed.

recena:
:bee:

src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
```diff
@@ -76,8 +77,23 @@ public PluginManager(Jenkins jenkins) {
...
+        waitFor().withTimeout(30, TimeUnit.SECONDS).until(new Callable<Boolean>() {
+            @Override
+            public Boolean call() throws Exception {
+                try {
+                    button.findElement(by.id("it does not matter"));
+                } catch(StaleElementReferenceException e) {
+                    return true; // we are in another page now
+                } catch(Exception e) {
+                }
```
jtnord: does this not need a `Thread.sleep(10)` - I don't want to see a CPU core spinning, or if we are running on a single CPU VM to be fighting with the update center fetch.

src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
```diff
@@ -76,8 +77,23 @@ public PluginManager(Jenkins jenkins) {
...
+        // We use the button itself to detect when the page has changed, which happens after the refresh has been done
+        final WebElement button = find(by.button("Check now"));
+        clickButton("Check now");
+        waitFor().withTimeout(30, TimeUnit.SECONDS).until(new Callable<Boolean>() {
+            @Override
+            public Boolean call() throws Exception {
+                try {
+                    button.findElement(by.id("it does not matter"));
+                } catch(StaleElementReferenceException e) {
+                    return true; // we are in another page now
```
jtnord: `Boolean.TRUE`

src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
```diff
@@ -76,8 +77,23 @@ public PluginManager(Jenkins jenkins) {
...
+                } catch(Exception e) {
+                }
+                return false;
```
jtnord: `Boolean.FALSE`

src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
andresrc: By default the waits have a polling interval of 500 ms.

src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
```diff
@@ -76,8 +77,23 @@ public PluginManager(Jenkins jenkins) {
...
      */
     public void checkForUpdates() {
-        visit("checkUpdates");
-        waitFor(by.xpath("//span[@id='completionMarker' and text()='Done']"));
+        visit("advanced");
+        // The check now button is a POST with a redirect to the same page.
+        // We use the button itself to detect when the page has changed, which happens after the refresh has been done
+        final WebElement button = find(by.button("Check now"));
+        clickButton("Check now");
+        waitFor().withTimeout(30, TimeUnit.SECONDS).until(new Callable<Boolean>() {
```
olivergondza: Note there is an overload where checked exceptions are treated as false. Plus, it gets reported in case the expected event never happen.

src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
```diff
@@ -76,8 +77,23 @@ public PluginManager(Jenkins jenkins) {
...
+            @Override
+            public Boolean call() throws Exception {
+                try {
+                    button.findElement(by.id("it does not matter"));
```
olivergondza: I do not understand this. We need to get navigated to the page after clicking the button _and_ wait for the right text to appear.

andresrc:
@olivergondza I've added an improved wait condition. That button executes a form submit (POST) to [this method](https://github.com/jenkinsci/jenkins/blob/jenkins-2.7.1/core/src/main/java/hudson/PluginManager.java#L1504) that just returns to the calling page in case of success or a plain `FormValidation` in case of error.

So the "wait condition" has been set to "I leave the current page and return to the same one"

olivergondza:
@andresrc Can we check the URL before and after clicking the button to tell we got redirected back?

andresrc:
@olivergondza that could be very fragile, imho, as we can get to the point that we never get "to see" the intermediate URL. Looking for stale components seems to be a recommended method to check that you have left the current page.

olivergondza:
I am not suggesting to check the the intermediate URL. As I understand, for successful check we are redirected back to URL where we ware before clicking the button. This is what we should check/wait for IMO.

PR #165, opened 2016-08-02
Cleanup

olivergondza:
- Unused imports
- Deprecated test
- Useless cucumber scenarios in JUnit comments
  - It has no value and confuses newcomers that waste their time writing those for new tests


recena:
👍  Thanks for this kind of things.

raul-arabaolaza:
If my opinion serves of anything, I really like this PR :+1:

andresrc:
:+1:

src/test/java/core/SlaveTest.java
```diff
@@ -46,9 +45,7 @@ public void setUp() throws ExecutionException, InterruptedException {
         slave = slaveController.install(jenkins).get();
     }
 
-    /**
-     * Scenario: Bring slave offline and then online.
-     */
+    // Bring slave offline and then online.
```
jtnord: use `/** */` so that it gets javadoc formatted and then the IDE can give you the doc,

src/test/java/plugins/JabberPluginTest.java
```diff
@@ -41,7 +40,7 @@
 @WithPlugins("jabber")
 @Category(DockerTest.class)
 @WithDocker
-@Ignore
+@Ignore // TODO bring this back to life or remove from testsuite
```
jtnord: `@Ignore("TODO: bring back...")`

jtnord:
> Useless cucumber scenarios in JUnit comments
It has no value and confuses newcomers that waste their time writing those for new tests

So I disagree with this comment - I actually find the gherkin descriptions a very easy thing to follow to understand what the test is actually **supposed** to be testing.  Quite often a test name will just not quite give you enough information.  Whilst you could look at this as gherkin it actually is really just a description of the pre-conditions and the expected behaviour - and this is 100% invaluable when a test fails. I have fix a few where I was stumped as to if the failure was in the setup or the actual test (ie is it a genuine test failure of the code being tests, or can I change this code to get the test to pass because it is just setup code), when you just have the java code it is completely not clear).

So I'm not rushing to give my +1 on this - to the point I am almost but not quite a thumbs down.

olivergondza:
> So I disagree with this comment - I actually find the gherkin descriptions a very easy thing to follow to understand what the test is actually supposed to be testing. Quite often a test name will just not quite give you enough information.

Agreed. I have not deleted all the test comments, just those that blindly duplicated the code (many of them ware copypasted from selenium-tests).

> I have fix a few where I was stumped as to if the failure was in the setup or the actual test (ie is it a genuine test failure of the code being tests, or can I change this code to get the test to pass because it is just setup code), when you just have the java code it is completely not clear).

I advocate in favor of using comments to express what can not be expressed by the code. I do not want to use a pile of comments that are so easy to get outdated (as gherkin scenarios full of details are) as an excuse to write code that is hard to understand. Let's make the code readable, not hide it between the comments.

jtnord:
:+1:

PR #168, opened 2016-08-06
[FIXED JENKINS-37232] - starting with 2.7 the ATH began to fail during create item

kzantow:
[JENKINS-37232](http://issues.jenkins-ci.org/browse/JENKINS-37232)

Firefox has an issue sending blur events when running the ATH. This simulates the blur event for a specific field that needs it during item creation.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/JobsMixIn.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/JobsMixIn.java
@@ -24,6 +24,7 @@ public class JobsMixIn extends MixIn {
 
         // since 2.7, a bug in Firefox webdriver may prevent the blur event from triggering
         // properly, so we manually execute a blur event here, see: JENKINS-37232
+        // See: https://github.com/seleniumhq/selenium-google-code-issue-archive/issues/7346
         try {
             executeScript(
                 "var obj = document.getElementById('name');"
@@ -31,7 +32,7 @@ public class JobsMixIn extends MixIn {
                 + "ev.initEvent('blur', true, false);"
                 + "obj.dispatchEvent(ev);"
                 + "return true;");
-        } catch(Exception e) {
+        } catch (Exception e) {
             // This should rarely fail on modern browsers,
             // we don't really care if it does since Firefox was 
             // the only one that seemed to exhibit the issue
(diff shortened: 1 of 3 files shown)


varmenise:
🐝

amuniz:
mega-:bee:

recena:
Waiting for the results from the build....

src/main/java/org/jenkinsci/test/acceptance/po/JobsMixIn.java
```diff
@@ -22,6 +22,21 @@ public JobsMixIn(ContainerPageObject context) {
...
+        // since 2.7, a bug in Firefox webdriver may prevent the blur event from triggering
+        // properly, so we manually execute a blur event here, see: JENKINS-37232
+        try {
+            executeScript(
+                "var obj = document.getElementById('name');"
+                + "var ev = document.createEvent('MouseEvents');"
+                + "ev.initEvent('blur', true, false);"
+                + "obj.dispatchEvent(ev);"
+                + "return true;");
+        } catch(Exception e) {
```
recena: `catch (Exception e) {`

recena:
@kzantow Do you have any reference (link) about this bug in the Firefox webdriver?

kzantow:
@recena fixed the formatting and added a link to (one of) the issues reported on this: https://github.com/seleniumhq/selenium-google-code-issue-archive/issues/7346

amuniz:
re-:bee:

oleg-nenashev:
🐝 
@samatdav What will happen when your PR gets merged?

kzantow:
@oleg-nenashev I assume https://github.com/jenkinsci/jenkins/pull/2388 is what you're referring to? There might need to be more changes here, this blur change will probably still be needed if the existing code is still being used to validate the name field.

samatdav:
@kzantow I guess Oleg may refer to https://github.com/jenkinsci/jenkins/pull/2416
@oleg-nenashev I am not sure but it seems like my PR should be fine. Since I do not change the standard behavior of the `blur` event. I only add another event.

recena:
👍

olivergondza:
I wonder what kind of FF problem that is. I experienced problems with events not sent / handled differently when firefox did not have focus. It works as expected in VNC, though. Is this one of those cases?

In that case I would really prefer requiring browser to have focus to run ATH (using dedicated display would be a requirement) over ad-hoc javascript fixups to compensate what is a problem between selenium and target browser (as I understand it).

kzantow:
@olivergondza it would be nice to have consistent behavior for Firefox, but really we just need this to work and people won't necessarily have access to run this through some VNC environment without a lot of setup. It would be irritating to add tests if you can't just run an individual one on your local machine, which wouldn't bode well for actually having tests added. And hopefully we can all agree having more tests is a good thing.

kzantow:
@reviewbybees done

kzantow:
@reviewbybees again :)

mafraba:
:bee:

recena:
👍

PR #169, opened 2016-08-08
Fixed failing docker tests

jcechace:
https://github.com/jenkinsci/acceptance-test-harness/commit/419192fff9c9c77ccab77841243b372b0ccf0fc3#diff-0e01d38750e8f41c10054e6f690694f2R22 Broke a bunch of docker tests by making ```dockerHostResolver``` static.  

This PR only modifies DockerImageTest so that it cleans ups properly after executing. The field should have never been static, however there is quite a few things depending on it now.


olivergondza:
The underlying problem was the mock leaked through the static field to real tests causing: `docker: Error response from daemon: failed to create endpoint stoic_hamilton on network bridge: Error starting userland proxy: listen tcp 192.168.99.100:33137: bind: cannot assign requested address.`

PR #170, opened 2016-08-10
Tweaks to VNC wrapper

jglick:
@reviewbybees & @olivergondza

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -747,7 +747,27 @@
         </plugins>
       </build>
     </profile>
-    <!-- TODO could be laboriously repeated for $TYPE, $DISPLAY, etc. (seems Maven provides no generic way to do this) -->
+    <profile>
+      <id>BROWSER_DISPLAY</id>
+      <activation>
+        <property>
+          <name>BROWSER_DISPLAY</name>
+        </property>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <configuration>
+              <environmentVariables>
+                  <BROWSER_DISPLAY>${BROWSER_DISPLAY}</BROWSER_DISPLAY>
+              </environmentVariables>
+            </configuration>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -747,7 +747,27 @@
         </plugins>
       </build>
     </profile>
-    <!-- TODO could be laboriously repeated for $TYPE, $DISPLAY, etc. (seems Maven provides no generic way to do this) -->
+    <profile>
+      <id>BROWSER_DISPLAY</id>
+      <activation>
+        <property>
+          <name>BROWSER_DISPLAY</name>
```
jglick: So we can add to `~/.m2/settings.xml` something like

```xml
<profile>
    <id>jenkins-acceptance-test-lts-2</id>
    <properties>
        <JENKINS_WAR>…/.m2/repository/org/jenkins-ci/main/jenkins-war/2.7.2/jenkins-war-2.7.2.war</JENKINS_WAR>
        <LOCAL_SNAPSHOTS>true</LOCAL_SNAPSHOTS>
        <INTERACTIVE>true</INTERACTIVE>
        <RECORDER>off</RECORDER>
        <BROWSER_DISPLAY>:42</BROWSER_DISPLAY>
    </properties>
</profile>
```

vnc.sh
```diff
@@ -7,7 +7,8 @@ if [ ! -z "$1" ]; then
   display=":$1"
 fi
 
-vncserver $display > /dev/null
-vncviewer $display > /dev/null &
+vncserver -kill $display
```
jglick: allows the script to be run repeatedly without manually cleaning up

vnc.sh
```diff
@@ -7,7 +7,8 @@ if [ ! -z "$1" ]; then
...
+vncserver -geometry 1750x1250 $display > /dev/null
```
jglick: default geometry was too small to view Firefox for me

vnc.sh
```diff
@@ -7,7 +7,8 @@ if [ ! -z "$1" ]; then
...
+(vncviewer $display || vncviewer localhost$display) > /dev/null &
```
jglick: `vncviewer-java` does not accept just `:42`, requires `localhost:42`

olivergondza:
:+1:

PR #171, opened 2016-08-10
Wait for the agent to come back online before we even try to resume the build

jglick:
Complements https://github.com/jenkinsci/pipeline-input-step-plugin/pull/2 by being more robust against JENKINS-37154-related behavior. Would benefit from https://github.com/jenkinsci/workflow-cps-plugin/pull/40.

@reviewbybees


svanoort:
:bee:

PR #178, opened 2016-08-24
Testing Pipeline + Subversion

jglick:
Was trying to reproduce [JENKINS-37538](https://issues.jenkins-ci.org/browse/JENKINS-37538) and failed. Anyway, it is useful to have a basic smoke test like this in place.

@reviewbybees


abayer:
I like tests! :bee:

PR #179, opened 2016-08-29
Stabilisation

jcechace:
Some minor tweaks helping to stabilise the execution of ATH. 

* A valid path is used as value of "Remote FS dir" for docker containers 
* Ant plugin tests should work correctly when ANT_HOME is not set at all

### DIFF
--- a/src/test/java/plugins/SshSlavesPluginTest.java
+++ b/src/test/java/plugins/SshSlavesPluginTest.java
@@ -48,6 +48,9 @@ import org.junit.experimental.categories.Category;
 @Category(DockerTest.class)
 @WithDocker
 public class SshSlavesPluginTest extends AbstractJUnitTest {
+
+    public static final String REMOTE_FS = "/tmp";
+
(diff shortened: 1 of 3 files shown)


jcechace:
Note that this by no means fixes all instabilities -- there are still things such a git plugin tests which I mystery to me and which I am unable to reproduce outside full jenkins ATH run.

src/test/java/plugins/SshSlavesPluginTest.java
```diff
@@ -123,7 +127,7 @@
         
         
         verify();
-        verifyLog("sh -c \"cd \"/tmp/" + slave.getName() + "\" && java  -jar slave.jar\"");
+        verifyLog("sh -c \"cd \"" + REMOTE_FS + "\" && java  -jar slave.jar\"");
```
olivergondza: We need to keep the slave name in the path to prevent several slaves using the same directory at the same time.

olivergondza: :facepalm: this is not page object fix.

PR #186, opened 2016-09-08
[JENKINS-38058] pass context when creating 'FolderItem' objects

mafraba:
[JENKINS-38058](https://issues.jenkins-ci.org/browse/JENKINS-38058)

Passing context page object to folders so that we can retrieve the real Jenkins instance later on

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/JobsMixIn.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/JobsMixIn.java
@@ -58,7 +58,14 @@ public class JobsMixIn extends MixIn {
     }
 
     public <T extends TopLevelItem> T get(Class<T> type, String name) {
-        return newInstance(type, injector, url("job/%s/", name), name);
+        try {
+            assert getContext() != null;
+            return newInstance(type, getContext(), url("job/%s/", name), name);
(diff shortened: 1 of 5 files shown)


raul-arabaolaza:
:bee:

mafraba:
@reviewbybees done

src/main/java/org/jenkinsci/test/acceptance/po/JobsMixIn.java
```diff
@@ -58,7 +58,14 @@ public Object call() throws Exception {
     }
 
     public <T extends TopLevelItem> T get(Class<T> type, String name) {
-        return newInstance(type, injector, url("job/%s/", name), name);
+        try {
+            assert getContext() != null;
+            return newInstance(type, getContext(), url("job/%s/", name), name);
+        } catch (AssertionError e) {
```
olivergondza: I do not like this as we throw the exception all over the place. We should either fix all the implementations and accept external client will be broken or make the `newInstance` to throw more specific exception to implement this fallback.

mafraba: Fair enough. I propose a different solution with an explicit check for the constructor, please see new commit.

PR #187, opened 2016-09-16
Fix property renaming from localRootPath to nodeMountPoint

alexsomai:
The External Workspace Node property was renamed from `localRootPath` to `nodeMountPoint`.

CC @oleg-nenashev @martinda


martinda:
Thanks Alex, good catch. I had forgotten about the acceptance tests!

alexsomai:
I wonder why the tests are failing, it says something about `UpdateCenter$InstallationFailedException`, although they are passing locally.

PR #188, opened 2016-09-16
Make plugin overriding work when uploadPlugins=false

andresrc:
As described [here|https://github.com/jenkinsci/acceptance-test-harness/blob/master/docs/SUT-VERSIONS.md] a custom plugin version can be provided by setting the `<id>.jpi` environment variable to the path of the custom version.

However, these mechanism does not work when the `uploadPlugins` config option is set, and this PR fixes that.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
@@ -8,6 +8,7 @@ import java.io.File;
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.Iterator;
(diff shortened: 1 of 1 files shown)


recampbell:
🐝

olivergondza:
How do you intend to implement that? Meaning how can UI install/update be used to deploy plugin snapshot builds from by `~/.m2/repository`? My current thinking is to do upload in case there is some overriding, UI install otherwise.

jglick:
I am guessing this is now obsolete. After #379, is there still any problem you see?

olivergondza:
Agreed.

PR #189, opened 2016-09-16
Skip tests that are not related to a set of 'interesting' plugins

andresrc:
This PR includes a new configuration option (see docs/ONLY_FOR_PLUGINS.md) that let's you skip tests that are not related to a set of specific plugins.

For a test to be considered "related" it must have a `WithPlugins` annotation that include any of the specified plugins or any plugin that depends on them.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/junit/JenkinsAcceptanceTestRule.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/junit/JenkinsAcceptanceTestRule.java
@@ -5,6 +5,8 @@ import com.google.inject.Injector;
 
 import org.jenkinsci.test.acceptance.controller.JenkinsController;
 import org.jenkinsci.test.acceptance.guice.World;
+import org.jenkinsci.test.acceptance.po.PluginManager;
+import org.jenkinsci.test.acceptance.update_center.UpdateCenterMetadata;
 import org.junit.internal.AssumptionViolatedException;
 import org.junit.rules.MethodRule;
 import org.junit.rules.TestRule;
@@ -15,6 +17,7 @@ import org.jvnet.hudson.annotation_indexer.Index;
 import org.openqa.selenium.NoSuchElementException;
 import org.openqa.selenium.WebDriver;
 
+import javax.inject.Provider;
 import java.io.IOException;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.AnnotatedElement;
@@ -50,6 +53,8 @@ public class JenkinsAcceptanceTestRule implements MethodRule { // TODO should us
             public void evaluate() throws Throwable {
                 World world = World.get();
                 Injector injector = world.getInjector();
+
+                checkOnlyForInterestingPlugins(injector.getProvider(UpdateCenterMetadata.class).get());
 
                 world.startTestScope(description.getDisplayName());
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/test/acceptance/junit/JenkinsAcceptanceTestRule.java
```diff
@@ -51,6 +54,8 @@ public void evaluate() throws Throwable {
                 World world = World.get();
                 Injector injector = world.getInjector();
 
+                checkOnlyForInterestingPlugins(injector.getProvider(UpdateCenterMetadata.class).get());
```
kohsuke: You can simply do `injector.get(UpdateCenterMetadata.class)` and Guice will be smart enough to know that it needs to use a provider.

kohsuke:
:bee:

recampbell:
🐝

olivergondza:
https://github.com/jenkinsci/acceptance-test-harness/blob/master/single-plugin.groovy is supposed to do the same thing if I understand this correctly.

recampbell:
Yes, I think we want to build on single-plugin.groovy. What @andresrc did here also checked for dependant plugins, but perhaps he can adapt that code to the single-plugin.groovy

andresrc:
Yes, I was not familiar with this, I'll refactor this PR.

scoheb:
Hi,

We have refactored single-plugin.groovy to be more strict in that it requires all plugins defined in @WithPlugins to be contained in the **TEST_ONLY_PLUGINS** variable.

For example, if the JiraPluginTest contains `@WithPlugins = git, jira` AND the environment variable **TEST_ONLY_PLUGINS** is "git" then the JiraPluginTest will be _skipped_.

I guess this might not cover your use case entirely, but since there was a comment on building on single-plugin.groovy, I'd thought I would share this....

```
// Run tests that requires plugins enumerated in TEST_ONLY_PLUGINS variable.
//
// This strategy is more strict in that it requires all plugins defined in @WithPlugins
// to be contained in TEST_ONLY_PLUGINS variable.
// For example, if the JiraPluginTest contains @WithPlugins = git, jira
// AND the environment variable TEST_ONLY_PLUGINS is "git"
// then the JiraPluginTest will be skipped.
//
// 'TEST_ONLY_PLUGINS=git,envinject' run all tests that require git or envinject
// using WithPlugins annotations

import org.junit.runner.Description;
import org.junit.runners.model.Statement;

import org.jenkinsci.test.acceptance.junit.FilterRule.Filter;
import org.jenkinsci.test.acceptance.junit.WithPlugins;
import org.jenkinsci.test.acceptance.po.PluginManager.PluginSpec;

import java.lang.annotation.Annotation;

bind Filter toInstance new FilterImpl();

class FilterImpl extends Filter {
    public String whySkip(Statement base, Description desc) {
        boolean foundAllPlugins = true;
        for (annot in getAnnotations(desc, WithPlugins.class)) {
            if (annot.value().size() == 0) {
                foundAllPlugins = false;
            }
            for (value in annot.value()) {
                if (!testOnlyPlugins().contains(new PluginSpec(value).name)) {
                    foundAllPlugins = false;
                }
            }
        }
        if (foundAllPlugins) {
            return null;
        }
        return "Running only tests for plugins: ${testOnlyPlugins()}";
    }

    private Collection<String> testOnlyPlugins() {
        return (System.getenv("TEST_ONLY_PLUGINS") ?: "").split(",").each { it.trim() };
    }
}
```

recampbell:
@olivergondza @andresrc This would be a great addition, what's the next step in merging this? Huge time saver!

olivergondza:
@recampbell, which one, there are 2 proposals here?

PR #190, opened 2016-09-17
Upgrade to jsch 0.1.54

recampbell:
to avoid com.jcraft.jsch.JSchException: Algorithm negotiation fail

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -402,7 +402,7 @@
     <dependency>
       <groupId>com.jcraft</groupId>
       <artifactId>jsch</artifactId>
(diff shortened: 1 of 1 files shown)


recampbell:
Noting that this fixes the bug I was seeing in the git plugin's ATH run.

Failed with  Algorithm negotiation fail:  https://ci.jenkins.io/job/Plugins/job/git-plugin/job/PR-439/10/

Passed using my version of ATH with this fix: https://ci.jenkins.io/job/Plugins/job/git-plugin/job/PR-439/11/

PR #191, opened 2016-09-17
[WIP] [JENKINS-38307] Stop installing detached plugins

recampbell:
We should just let the war behave as normal instead of hacking the plugins to avoid fixing some WithPlugin lists.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/controller/LocalController.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/controller/LocalController.java
@@ -291,6 +291,7 @@ public abstract class LocalController extends JenkinsController implements LogLi
         if (javaHome != null) {
             env.put("JAVA_HOME", javaHome.getAbsolutePath());
         }
+        env.put("jenkins.install.state", "TEST");
         return env;
     }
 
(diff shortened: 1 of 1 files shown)


recampbell:
Noting that this will likely break a lot of tests, but since I'd rather not cause a brown out by running the ATH locally, I'll let the PR builder handle it.

recena:
@recampbell If we do that, what happens whether we want to run the ATH with previous releases of 2.x? /cc @olivergondza

olivergondza:
@recena, Good point, that needs to be retested.

The change itself did not break much actually.

olivergondza:
The fact we do this seems to be cause of plugin installations problems or an extra work at least. We unbundle outdated credentials plugin, and dynamic deploy of needed plugins fails because credentials plugins is too old. If we quit doing that, the correct plugin version can be installed right away.

My initial concern about pre-2.0 jenkins is not valid as we have started to do this *after* 2.0 - the older version will unbundle the plugins correctly.

I am running tests against 1.609.3 and 2.19.3 before I merge this.

olivergondza:
There seems to be ~5 ATH testcase affected by [JENKINS-37545](https://issues.jenkins-ci.org/browse/JENKINS-37545). Since the bug is and will be present in releases we want to support in ATH for a while and plugin installation is sort of crucial. I lean towards updating the testcases so the detached dependencies are requested explicitly. It will cause no harm to Jenkins 1.X and will make Jenkins 2.X work.

olivergondza:
Continuing in #251.

PR #193, opened 2016-09-29
Delete no-docker variant of LDAP test. The docker one if thriving and…

olivergondza:
… covers more.

The test case was deprecated for a long time and nobody touch that for almost 2 years. Also noone ever reported results during LTS RC testing despite being part of the testplan. Removed from there as well.


scoheb:
+1

PR #200, opened 2016-10-06
[JENKINS-37276] Path atributes can change values if reordered

raul-arabaolaza:
[JENKINS-37276](https://issues.jenkins-ci.org/browse/JENKINS-37276)

Downstream of [PR-5](https://github.com/jenkinsci/form-element-path-plugin/pull/5)

I am not very sure about the `getText().contains` part, but as long as I see I only have four alternatives:

* The `getText().contains` one
* Modify @Describable value in the same way that the proposed by #162 
* Implement any other suggestion
* Let the things like they are because there is a very easy workaround for the bug

Option 2 seems not to have a lot of supporters given the time the PR has been open without activity
I am open to option 3 as soon as I receive a better alternative approach
I am also open to option 4 if anyone considers that this bug is not important or worth the risk

@reviewbybees specially @jglick 
Also @olivergondza opinion would be great here

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/FallbackConfig.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/FallbackConfig.java
@@ -251,7 +251,7 @@ public class FallbackConfig extends AbstractModule {
     @Named("form-element-path.hpi") @Provides
     public File getFormElementsPathFile(RepositorySystem repositorySystem, RepositorySystemSession repositorySystemSession) {
         ArtifactResolverUtil resolverUtil = new ArtifactResolverUtil(repositorySystem, repositorySystemSession);
-        ArtifactResult resolvedArtifact = resolverUtil.resolve(new DefaultArtifact("org.jenkins-ci.plugins", "form-element-path", "hpi", "1.4"));
+        ArtifactResult resolvedArtifact = resolverUtil.resolve(new DefaultArtifact("org.jenkins-ci.plugins", "form-element-path", "hpi", "1.7-alpha-1"));
         return resolvedArtifact.getArtifact().getFile();
     }
 
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Job.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Job.java
@@ -137,7 +137,7 @@ public class Job extends TopLevelItem {
         elasticSleep(1000); // it takes some time until the element is visible
         String path = null;
         // if I have got a Describable use it to get the path
-        if (describableValue != null) {
+        if (describableValue != null && type.equals(PostBuildStep.class)) {
             List<WebElement> newSteps = driver.findElements(by.xpath("//div[@name='%s']", section));
             outer:
             for (WebElement element : newSteps) {
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Job.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Job.java
@@ -137,7 +137,7 @@ public class Job extends TopLevelItem {
         elasticSleep(1000); // it takes some time until the element is visible
         String path = null;
         // if I have got a Describable use it to get the path
-        if (describableValue != null && type.equals(PostBuildStep.class)) {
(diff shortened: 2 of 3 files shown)


olivergondza:
Is this not the same problem as [JENKINS-37276](https://issues.jenkins-ci.org/browse/JENKINS-37276)? When adding publishers it is not added to the end but somewhere in the middle depending on things user can not affect or examine - `Publisher` vs. `Notifier`.

raul-arabaolaza:
@olivergondza You are right, I have marked my issue as duplicated, sorry for that.

However I believe the solution lies in the ATH not assuming any order for publishers instead of making assumptions or forcing publishers to be added in any specific order. Do you believe there is some value on this idea?

olivergondza:
I am not sure I understand. Is there a way for us to force in which order they appear in UI? I do not think so.

raul-arabaolaza:
@olivergondza Sorry, my English is not good enough sometimes. I was trying to say that i also believe that there is no way for us to force any order. Hence I believe the approach in this PR is a valid one.

BTW I have updated the PR to refer the proper issue instead of the duplicated one

olivergondza:
Identifying the added publisher in the full list based on the title in its markup is a bit fragile (but why not). But this does not address the fact, the way path elements are added will change `path`s on other elements as well.

Consider you have added the publisher *A* and its PO is accessing it via path `/publishers`. Then you add *B*. In case the final order will be *A > B*, the first PO will be valid and the second will get `/publishers/[1]` associated. So far so good. But if the resulting order will be "B > A", you will have to try this hard to get the correct path so *B's* PO will point to `/publishers`. But that is where *A's* PO used to point to!

Generally, no previously created publisher POs can be trusted after ne publisher as added as the PATHs they have been using might or might not have been shuffled. That is why I suggest in jira to fix the form attribute assigning plugin never to change paths on elements so they will be in order they have got rendered in and new ones will be in order in which they have got added - not numbered from top to bottom.

Does that make any sense?

raul-arabaolaza:
@olivergondza I think i understand what you mean. 

But even if paths are not changed and remain in the same order that they were added... the fact is that publishers will be ordered on the UI  and the `last` method used by `Job.addStep` will take as path of the newly added publisher the path of the last publisher in the UI, which may or not be the last added. So, even if the path is not changed we need to implement a solution like the one on this PR. 

Another possibility is to try to recalculate and refresh the path for existing publishers (do not know if that's possible) every time we add a new one. Given the usual number of publishers in our tests I do not believe it will result in a performance penalty. Given that now paths may be wrong in any moment I also do not believe it will broke any test not already broken.

So, to resume IIUC I can try to help with the form-element-path-plugin and keep this PR as is or try the refresh approach... What do you think?

PD Thanks for your time I know you are a busy man :)

stephenc:
The publishers *when round-tripped through the config screen* will be in `DescriptorList` order (i.e. reflecting the `@Extension(ordinal)` and some other aspects due to how the descriptor list is handled on the client side)

Programmatically, you can construct an XML or a memory model that is in a different order, but the UI should maintain the `DescriptorList` order due to https://github.com/jenkinsci/jenkins/blob/abf41abe15129acf3552ef2c276f98f68f51b09e/core/src/main/resources/lib/hudson/project/config-publishers2.jelly#L37 (see https://github.com/jenkinsci/jenkins/blob/abf41abe15129acf3552ef2c276f98f68f51b09e/core/src/main/resources/lib/form/hetero-list.jelly#L63-L66 for the documentation)

Not sure how this affects the PR, but it is the logic that needs to be addressed

olivergondza:
@raul-arabaolaza, I agree that the ATH will need to be adapted to fully fix this (provided we will go the `form-element-path-plugin` way).

Allowing page areas to recalculate common prefix path would require changing all attached `Controls`, that are created eagerly, on all existing publishers. We would have to replace a couple of nice final fields with nasty *active object lists* (po > publisher and publisher > control) which I would like to avoid.

raul-arabaolaza:
@olivergondza I have created [PR-5](https://github.com/jenkinsci/form-element-path-plugin/pull/5) on form element path plugin to complement this one. If that PR is fine we can move forward with this

PR #207, opened 2016-10-21
Rely on plugin manager instead of custom dependency resolution by default

olivergondza:
There was a lot of effort put into the plugin upload optimization in ATH but there are still problems with the dependency resolution we provide. Also, the more complicated the algorithms get, the less is what `@WithPlugins` does similar to what user initiated plugin installation would do. Putting aside we do not contribute any coverage for dependency resolution this way.

I am creating this to see how much time we save this way.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
@@ -74,7 +74,7 @@ public class PluginManager extends ContainerPageObject {
         // injection happens in the base class, so for us to differentiate default state vs false state,
         // we need to use Boolean
         if (uploadPlugins==null)
-            uploadPlugins = true;
+            uploadPlugins = false;
     }
 
     /**


olivergondza:
TODO: implement plugin update in case it exists

olivergondza:
The test timed out in 10 hours though it eliminated many plugin installation related problems. It seems to operate more than 10% slower. Perhaps this is a way to go after splitting the load with Pipeline.

olivergondza:
@jenkinsci/code-reviewers, WDYT?

andresrc:
We have been using `uploadPlugins=false` in some runs and it is slower, but provides a more real experience, so imho this change it's worth doing

:+1:

olivergondza:
This will not be all that easy, as `UpdateCenterMetadataDecorator`s depend on this. The options like `artifactid.version`, `artifactid.jpi` or `LOCAL_SNAPSHOTS` would be have no effect silently. So I guess we have to fix the uploading after all.

The question is if we need to keep the UI approach as well.

andresrc:
I think the UI approach must be kept, as doing the things in the same way as the user would do is a must have for some scenarios, but the upload must be fixed, as is needed for those plugin customization options.

I have some ongoing work related to that in https://github.com/jenkinsci/acceptance-test-harness/pull/188 that I have to finish and adapt to this PR.

olivergondza:
Another reason to go this way: [JENKINS-40775](https://issues.jenkins-ci.org/browse/JENKINS-40775)

jglick:
> I think the UI approach must be kept, as doing the things in the same way as the user would do is a must have for some scenarios

Why? We pay a substantial penalty (developers writing tests too) for doing UI-based upload. And uploading plugins via REST `/pluginManager/uploadPlugin` is supported too. We also do some things in the harness via REST. (You could also use the CLI `install-plugin` command but then you need to download the CLI agent which is more of a hassle.)

Cf. https://github.com/jenkinsci/acceptance-test-harness/pull/211#discussion_r119151235.

jglick:
My current proposal is to throw out the current plugin manager code, with all its warts and special cases, and rewrite as follows:

When the test starts, the “real” UC JSON is obtained on demand (using the one-day cache as now). Then the test VM starts a simple HTTP service on a random port which basically serves as a mock UC. Any `LOCAL_JARS` etc. will override metadata in the JSON. All the download URLs are rewritten to come from the mock service too. When a plugin is requested, if it is being overridden with a local copy, that override is served; otherwise we look it up in the local Maven repository by `gav`, downloading on demand, and serve from there.

We would probably need no REST calls at all. UI manipulation would be limited to editing the UC URL field, and then `@WithPlugins` would simply check off boxes in the **Available** tab. We might need to deal with signatures somehow, by disabling them I guess.

WDYT?

jglick:
Just noticed another case where the current system breaks down: it does not grok implicit deps on split plugins. So for example Jenkins will try to update the 4Mb `bouncycastle-api` from the live update center. While it could in principle take advantage of https://github.com/jenkinsci/jenkins/pull/3110 for newer cores, this would just make the logic even more complex.

jglick:
I have also seen that Jenkins is frequently restarted during plugin installation, which makes no sense since as of 2.0 there are no bundled plugins and thus should have been no updates. I think this is due to the hacks in `UpdateCenterMetadata` not interacting well with transitive dependencies computed by Jenkins itself; probably the same bug as in #355.

olivergondza:
@jglick, It seems like compromise between my desire to get rid of the custom dependency solving algorithm and yours of fast plugin installation. I also do not see a simplest way to go.

olivergondza:
Closing as obsolete

PR #210, opened 2016-10-27
[JENKINS-38928] Add a test doing apply then save in a job

Vlatombe:
This test breaks without https://github.com/jenkinsci/form-element-path-plugin/pull/6


src/main/java/org/jenkinsci/test/acceptance/po/ConfigurablePageObject.java
```diff
@@ -98,7 +101,16 @@ public void save() {
 
     public void apply() {
         clickButton("Apply");
-        elasticSleep(1000);
-        assertThat(driver, hasContent("Saved"));
+        waitFor(driver).withTimeout(5, TimeUnit.SECONDS)
```
olivergondza: `waitFor(driver, hasContent("Saved"), 5);` should do the job.

PR #211, opened 2016-10-27
Reduce the number of restarts needed to install plugins

olivergondza:
Ripping of the shared goods from https://github.com/jenkinsci/acceptance-test-harness/pull/207 (Rely on plugin manager instead of custom dependency resolution by default).

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/junit/WithPlugins.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/junit/WithPlugins.java
@@ -7,7 +7,7 @@ import com.google.inject.name.Named;
 import org.jenkinsci.test.acceptance.po.Jenkins;
 import org.jenkinsci.test.acceptance.po.Plugin;
 import org.jenkinsci.test.acceptance.po.PluginManager;
-import org.jenkinsci.test.acceptance.po.PluginManager.PluginSpec;
+import org.jenkinsci.test.acceptance.update_center.PluginSpec;
 import org.jenkinsci.test.acceptance.update_center.UpdateCenterMetadata.UnableToResolveDependencies;
 import org.jenkinsci.test.acceptance.utils.pluginreporter.ExercisedPluginsReporter;
 import org.junit.internal.AssumptionViolatedException;
@@ -26,8 +26,6 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Set;
-import java.util.TreeSet;
 import java.util.logging.Logger;
 
 import static java.lang.annotation.ElementType.*;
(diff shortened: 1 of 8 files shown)


olivergondza:
2 failures on weekly, 0 for stable core. Merging

src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
```diff
@@ -186,101 +197,91 @@ public boolean installPlugins(final String... specs) throws UnableToResolveDepen
...
-                        try {
-                            new UpdateCenter(jenkins).waitForInstallationToComplete(name);
-                        } catch (InstallationFailedException x) {
-                            if (!restartRequired) {
-                                throw x;
-                            }
-                            // JENKINS-19859: else ignore; may be fine after the restart
+                        installPlugin(localFile);
+                    } finally {
+                        if (!localFile.delete()) {
```
jglick: This broke `LOCAL_SNAPSHOTS=true`.

src/main/java/org/jenkinsci/test/acceptance/junit/WithPlugins.java
```diff
@@ -88,57 +90,66 @@ public Statement apply(final Statement base, final Description d) {
...
+                            d.getTestClass().getAnnotation(WithPlugins.class)
+                    );
+
+                    installPlugins(plugins);
+
+                    for (PluginSpec plugin : plugins) {
+                        pluginReporter.log(
+                                d.getClassName() + "." + d.getMethodName(),
+                                plugin.getName(),
+                                plugin.getVersion()
```
jglick: This is wrong. It just spits back out the contents of the `@WithPlugins` annotation, regardless of what Jenkins is actually running.

src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
```diff
@@ -186,101 +197,91 @@ public boolean installPlugins(final String... specs) throws UnableToResolveDepen
...
+                        throw new AssumptionViolatedException(
+                                name + " has version " + availableVersion + " but " + requiredVersion + " was requested");
+                    }
+                    File localFile = newPlugin.resolve(injector, availableVersion);
                     try {
-                        if (new VersionNumber(claimedVersion).compareTo(new VersionNumber(availableVersion)) > 0) {
-                            throw new AssumptionViolatedException(
-                                    name + " has version " + availableVersion + " but " + claimedVersion + " was requested");
-                        }
-                        newPlugin.uploadTo(jenkins, injector, availableVersion);
```
jglick: The old method was fast—a direct POST. Now we are going through the web driver to visit `/pluginManager/advanced` and upload via HTML, which is a lot slower. Is there no middle ground?

olivergondza: Preferred this to get closer to real user interaction. Since this is not the real meat of this change, I do not feel strongly about that.

PR #212, opened 2016-10-28
[JENKINS-39354] Improve CreateSlaveTest robustness

varyvol:

@reviewbybees

### DIFF
--- a/single-plugin.groovy
+++ b/single-plugin.groovy
@@ -8,7 +8,7 @@ import org.junit.runners.model.Statement;
 
 import org.jenkinsci.test.acceptance.junit.FilterRule.Filter;
 import org.jenkinsci.test.acceptance.junit.WithPlugins;
-import org.jenkinsci.test.acceptance.po.PluginManager.PluginSpec;
+import org.jenkinsci.test.acceptance.update_center.PluginSpec;
 
 import java.lang.annotation.Annotation;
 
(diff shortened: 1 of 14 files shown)


recena:
🐝

olivergondza:
What is the original problem? How does this fix it? Personally, I prefer interaction with Page Object/Area (`SshPrivateKeyCredential`) over numerous overloads that forwards POs attributes (this is what the PR seems to do).

varyvol:
The problem is the test was not waiting for the new credential to appear after clicking the add button through SshPrivateKeyCredential, which was was making the test fail randomly.

SshSlaveLauncher#pwdCredentials and SshSlaveLauncher#keyCredentials already:
* Put the data in the dialog.
* Click in add button.
* Wait for the new credential to appear.

I suspect for this test the logic was placed in the test class because there was no method to set a key credential including the description, and that's why I created one (in  SshSlaveLauncher in order not to duplicate the logic already available there).

olivergondza:
I presume the problem is in `addCredential` that does not wait for the Page Area to be rendered. It should be sufficient to fix that one method but hard to tell without stacktrace. There is a pending PR to address this long term struggle https://github.com/jenkinsci/acceptance-test-harness/pull/148, though it does not include this method.

varyvol:
Not exactly, at least for this particular test. It's more like the opposite: all the data has been introduced in the Page Area and then we click a button to close it and go on with the following actions. The problem is such next action sometimes is performed before the dialog has been actually closed.

olivergondza:
Oh, so it is `SshCredentialDialog#add()` returning prematurely, then.

varyvol:
That's right @olivergondza

olivergondza:
A variant of `waitForCredentialVisible` should be moved to `SshCredentialDialog#add` so other pages that add credentials can benefit from this. Then there is no need to encapsulate the interaction with the page object itself.

varyvol:
@olivergondza I have added such wait in the #add method.

The only thing I don't like is that there is logic in the test that is present also in SshSlaveLauncher.

PR #213, opened 2016-10-31
[JENKINS-39354] Improve CreateSlaveTest robustness

varyvol:

Following up https://github.com/jenkinsci/acceptance-test-harness/pull/212

@reviewbybees and @olivergondza

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_credentials/SshCredentialDialog.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_credentials/SshCredentialDialog.java
@@ -4,6 +4,12 @@ import org.jenkinsci.test.acceptance.plugins.credentials.Credential;
 import org.jenkinsci.test.acceptance.po.Control;
 import org.jenkinsci.test.acceptance.po.PageAreaImpl;
 import org.jenkinsci.test.acceptance.po.PageObject;
+import org.openqa.selenium.By;
+import org.openqa.selenium.NoSuchElementException;
+import org.openqa.selenium.StaleElementReferenceException;
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_credentials/SshCredentialDialog.java
```diff
@@ -36,6 +42,19 @@ protected void resolve(String caption) {
...
+        final By addSubmitButton = by.id("credentials-add-submit-button");
+
+        find(addSubmitButton).click();
+
+        waitFor().withTimeout(5, TimeUnit.SECONDS).until(new Callable<Boolean>() {
+            @Override
+            public Boolean call() throws Exception {
+                try {
+                    return !find(addSubmitButton).isDisplayed();
+                } catch (final NoSuchElementException | StaleElementReferenceException ex) {
```
olivergondza: It is better to pass these to `Wait#ignoring` as it will be attached to `TimeoutException` for easier problem analysis.

olivergondza:
It looks good to me otherwise.

src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_credentials/SshCredentialDialog.java
varyvol: I think it is not valid for this situation. Ignoring such exceptions will always make the test fail due to timeout, as the button is not there (which is actually expected) and therefore a timeout will be thrown.

I need the exceptions to be thrown to detect the button has already been hidden and then return true.

olivergondza: Right, it is a signal of success, not a failure. Nevermind.

raul-arabaolaza:
:bee:

PR #216, opened 2016-11-04
[JENKINS-39500] ExternalWorkspaceManagerPluginTest might fail due to plugins installation order

varyvol:
[JENKINS-39500](https://issues.jenkins-ci.org/browse/JENKINS-39500)

@reviewbybees

### DIFF
--- a/src/test/java/plugins/ExternalWorkspaceManagerPluginTest.java
+++ b/src/test/java/plugins/ExternalWorkspaceManagerPluginTest.java
@@ -46,6 +46,8 @@ public class ExternalWorkspaceManagerPluginTest extends AbstractJUnitTest {
 
     @Before
     public void setUp() throws Exception {
+        jenkins.restart();
+
         MavenInstallation.installMaven(jenkins, "M3", "3.1.0");
 
         setUpGlobalConfig();
--- a/src/test/java/plugins/ExternalWorkspaceManagerPluginTest.java
+++ b/src/test/java/plugins/ExternalWorkspaceManagerPluginTest.java
@@ -15,6 +15,7 @@ import org.junit.ClassRule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
 import org.openqa.selenium.By;
+import org.openqa.selenium.NoSuchElementException;
 
 import java.util.concurrent.ExecutionException;
 
@@ -46,8 +47,6 @@ public class ExternalWorkspaceManagerPluginTest extends AbstractJUnitTest {
 
     @Before
     public void setUp() throws Exception {
-        jenkins.restart();
-
         MavenInstallation.installMaven(jenkins, "M3", "3.1.0");
 
         setUpGlobalConfig();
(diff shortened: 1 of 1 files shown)


olivergondza:
The order in which plugins are installed should not matter. How does the problem manifests?

I see some failures in [1] but that recovers after restart during plugin installation - dynamic deploy failed in a nasty way but installation worked after restart.

[1] https://jenkins.ci.cloudbees.com/job/core/job/acceptance-test-harness/2897/testReport/junit/plugins/ExternalWorkspaceManagerPluginTest/shareWorkspaceTwoJobsTwoNodes/

varyvol:
We are watching that for some WARs the restart does not happen even though the dynamic deploy fails. That, as a result, makes the functionality not available, specifically some configurations are not available in the configuration page (External Workspace Definitions).

olivergondza:
https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java#L261 is supposed to get the Jenkins restarted if needed and verify the plugins are installed afterwards. If this is broken, this is where it needs to be fixed. Are some plugins not installed? Or not fully functional after restart without exception presented?

varyvol:
No restart happens as the installation doesn't throw any error on the screen (it does in the logs as you can check in the first link you provided here). The plugin appears as installed but is not fully functional.

olivergondza:
I believe it is restarted:
```
Restarting Jenkins to finish plugin installation
master57633|Nov 03, 2016 8:52:04 PM jenkins.model.Jenkins$25 run
master57633|SEVERE: Restarting VM as requested by anonymous
master57633|Nov 03, 2016 8:52:04 PM jenkins.model.Jenkins cleanUp
```

https://jenkins.ci.cloudbees.com/job/core/job/acceptance-test-harness/2897/testReport/junit/plugins/ExternalWorkspaceManagerPluginTest/shareWorkspaceTwoJobsTwoNodes/

varyvol:
Yes, in that execution it gets restarted and actually the disk pool is configured and the test is executed. But it's not always like that.

andresrc:
@olivergondza when running against 2.19.1 core, there are issues that seems related to [JENKINS-21486](https://issues.jenkins-ci.org/browse/JENKINS-21486) which are the ones that trigger a restart.

In our internal WAR builds, as there are some "baseline" plugins installed those issues does not happen, so no restart is triggered. The issue then seems to be that, given that there are optional dependencies involved, the plugins are installed but some needed extensions (relying on those optional dependencies) may not be working. After the restart all extensions are correctly loaded.

With the changes in https://github.com/jenkinsci/acceptance-test-harness/pull/211, plugin order should not matter (assuming `uploadPlugins=false` which is our case) as it should be calculated by core.

olivergondza:
Please provide an example of failing test you attempt to fix, then.

varyvol:
@olivergondza I have made the restart conditional, based on the presence of the element. This way we avoid unnecessary restarts when the installation finished successfully.

olivergondza:
I really believe this needs to be fixed in plugin installation logic. Also, restarting in page area is not a good idea as whoever manipulates it can have unsaved changes in other controls/areas on that page.

varyvol:
This is what we are watching. The installation is reported to be successful but some components are not able to load.

screenshot1
screenshot2

olivergondza:
So all the plugins are loaded (that is why ATH consider it to be successful install) but not all of their extension points are functional. I suppose this might be because of installing *External Workspace Manager* sooner that the *Pipeline: Supporting APIs* that declare the `RunWrapper`.

As I tested for the dependency resolution algorithm the plugins are added in dependency-first manner. This is probably caused by `WithPlugins` passing the 2 plugins *into* that algorithm in dependency-last order. I am working on a fix.

olivergondza:
I have been digging into this for a while and I conclude the guice warning is a bug in `external-workspace-manager` plugin: It declares `workflow-support` as optional dependency but when loaded without it it fails in a way Jenkins do not even detect deploy has failed and claim plugin is loaded OK but the functionality is limited.

varyvol:
@olivergondza maybe for this specific case it is a bug, but I think the underlying scenario is still there and is perfectly valid.

A plugin is declaring a dependency as optional, so when installed it is fine not having that dependency and therefore not offering the related functionality. However as we are running a test, if we request the plugin to be installed through @WithPlugins, we want to have the whole set of functionalities. 

For example for PMDPluginTest (which originated https://github.com/jenkinsci/acceptance-test-harness/pull/217), where PMD is installed before dashboard-view and that optional functionality is not offered (which does need to be a bug but something desired) making the test to fail.

olivergondza:
It seems that ATH dependency resolution can go a step further and make sure optional dependencies are installed before the dependent plugins. @varyvol, is that correct.

varyvol:
@olivergondza I think that if an optional plugin is contained among the @WithPlugins ones, it should be installed so that the whole functionality is obtained. I think there is no need to install all the optional ones but just the ones that are being expressly pointed in the test. @andresrc

varyvol:
@olivergondza did you have the opportunity to work on a fix for this? Can I be of some help to get this sorted?

olivergondza:
I have not worked any further on this, sorry.

varyvol:
@olivergondza no worries at all. I was spending some time on this, so I am gonna close this one and will raise a ticket and its subsequent PR.

PR #217, opened 2016-11-04
[JENKINS-39504] PmdPluginTest might fail due to plugins installation order

varyvol:
[JENKINS-39504](https://issues.jenkins-ci.org/browse/JENKINS-39504)

In some WARs, the plugins are being installed in different order and the functionality is not available without a restart.

@reviewbybees


olivergondza:
Same as https://github.com/jenkinsci/acceptance-test-harness/pull/216#issuecomment-259499741.

uhafner:
So this one can be closed now? And the fix could be done during plugin loading?

varyvol:
@uhafner we think so. I am closing it.

PR #218, opened 2016-11-08
[JENKINS-39354] CreateSlaveTest robustness

varyvol:
- Perform the click through an action and moving previously to the element.
- Increase the timeout to be sure we wait for enough time.

@reviewbybees @olivergondza

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_credentials/SshCredentialDialog.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_credentials/SshCredentialDialog.java
@@ -7,6 +7,8 @@ import org.jenkinsci.test.acceptance.po.PageObject;
 import org.openqa.selenium.By;
 import org.openqa.selenium.NoSuchElementException;
 import org.openqa.selenium.StaleElementReferenceException;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.interactions.Actions;
 
 import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
@@ -44,9 +46,9 @@ public class SshCredentialDialog extends PageAreaImpl {
     public void add() {
         final By addSubmitButton = by.id("credentials-add-submit-button");
 
-        find(addSubmitButton).click();
+        this.findAndPerformClick(addSubmitButton);
 
-        waitFor().withTimeout(5, TimeUnit.SECONDS).until(new Callable<Boolean>() {
+        waitFor().withTimeout(10, TimeUnit.SECONDS).until(new Callable<Boolean>() {
             @Override
             public Boolean call() throws Exception {
                 try {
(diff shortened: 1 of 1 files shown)


olivergondza:
Please attach a failure log. It seems stable on jenkins.ci.cloudbees.com:

https://jenkins.ci.cloudbees.com/job/core/job/acceptance-test-harness-stable/lastCompletedBuild/testReport/core/CreateSlaveTest/history/
https://jenkins.ci.cloudbees.com/job/core/job/acceptance-test-harness/lastCompletedBuild/testReport/core/CreateSlaveTest/history/

varyvol:
Error Message

Timed out after 5 seconds waiting for org.jenkinsci.test.acceptance.plugins.ssh_credentials.SshCredentialDialog$2@31de8099

Stacktrace

org.openqa.selenium.TimeoutException: Timed out after 5 seconds waiting for org.jenkinsci.test.acceptance.plugins.ssh_credentials.SshCredentialDialog$2@31de8099
	at org.openqa.selenium.support.ui.FluentWait.timeoutException(FluentWait.java:292)
	at org.jenkinsci.test.acceptance.junit.Wait.timeoutException(Wait.java:143)
	at org.openqa.selenium.support.ui.FluentWait.until(FluentWait.java:261)
	at org.jenkinsci.test.acceptance.junit.Wait.until(Wait.java:74)
	at org.jenkinsci.test.acceptance.plugins.ssh_credentials.SshCredentialDialog.add(SshCredentialDialog.java:49)
	at core.CreateSlaveTest.newSlave(CreateSlaveTest.java:62)

armfergom:
:bee:

raul-arabaolaza:
@varyvol Do we need both the timeout increase and the click change? If the problem is related to the button not being clicked and the use of actions solves it there is no reason to have such big timeout
 
Maybe is time to give some love to [JENKINS-36551](https://issues.jenkins-ci.org/browse/JENKINS-36551) to get rid of all this hacks for the `click()` method

varyvol:
@raul-arabaolaza it is unclear the problem is the button not being clicked, it was more kind of a theory. This could avoid it and also increasing the timeout will help in case the problem was the system was under heavy load and needed more time to execute the action.

raul-arabaolaza:
@varyvol I believe we should make the changes to fix the problem and only those changes, it seems easy to try only with the timeout increase to make sure. Honestly a 5 sec timeout for a modal to close should be more than enough given that in local or jenkins.ci there are no problems

varyvol:
@raul-arabaolaza yes, I am not a believer in this being a timeout problem. So I am gonna revert it and let's check if using Actions solves the issue.

varyvol:
@olivergondza are you fine with this? It shouldn't affect already working scenarios and it might solve some where scrolling could provoke some issues.

olivergondza:
>  It shouldn't affect already working scenarios and it might solve some where scrolling could provoke some issues.

*Might*? As I said, upstream is not able to reproduce the problem you are trying to fix. So the only way to proceed I can see is you confirming it improves the situation in special conditions where you run the tests.

Have a look at the patch from a perspective of someone who spot it 2 years from now. Why are we doing something completely different in `SshCredentialDialog` then in every other interaction? Do we need to do this everywhere? Can I safely remove/refactor that if it does not break any tests? I am not able to tell.

>  I am not a believer in this being a timeout problem. So I am gonna revert it and let's check if using Actions solves the issue.

What is the result?

varyvol:
@olivergondza yes, it results. 

I have created a periodically launched job which launches the test in parallel, with the same WAR but using branch master and this branch.

We get failures in master but none on this branch.

varyvol:
@olivergondza are you fine with this being merged?

PR #223, opened 2016-11-15
[JENKINS-39743] Create test for the javadoc retention for each build

varyvol:
[JENKINS-39743](https://issues.jenkins-ci.org/browse/JENKINS-39743)

@reviewbybees

### DIFF
--- a/src/test/java/plugins/JavadocPluginTest.java
+++ b/src/test/java/plugins/JavadocPluginTest.java
@@ -8,6 +8,7 @@ import org.jenkinsci.test.acceptance.po.FreeStyleJob;
 import org.jenkinsci.test.acceptance.po.Job;
 import org.jenkinsci.test.acceptance.po.MatrixConfiguration;
 import org.jenkinsci.test.acceptance.po.MatrixProject;
+import org.junit.Assert;
 import org.junit.Test;
 
 import static org.hamcrest.MatcherAssert.assertThat;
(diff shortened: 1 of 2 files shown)


src/test/java/plugins/JavadocPluginTest.java
```diff
@@ -35,11 +39,30 @@ public void publish_javadoc_from_matrix_job() {
...
+    public void validate_javadoc_retention() {
+        FreeStyleJob job = jenkins.jobs.create();
+        setup(job);
+        Build b = job.startBuild().shouldSucceed();
+
+        assertThat(job, hasAction(JAVADOC_ACTION));
+        try {
+            assertThat(b, hasAction(JAVADOC_ACTION));
+            Assert.fail("Build #1 should not have Javadoc action");
+        } catch (AssertionError ex) {}
```
mafraba: Suggestion. Could this whole try block simply be replaced for the following? 

`assertThat("Build #1 should not have Javadoc action", b, not(hasAction(JAVADOC_ACTION)))`

olivergondza: It should, that is the reason we prefer hamcrest.

varyvol: I did not know that not. Thanks for the suggestion, changing it.

olivergondza:
:+1:

mafraba:
🐝

PR #224, opened 2016-11-15
Initial hello world ATH test for Declarative Pipeline

abayer:
More tests are coming! But here's a start.

cc @reviewbybees esp @rsandell

### DIFF
--- a/src/test/java/plugins/DeclarativePipelineTest.java
+++ b/src/test/java/plugins/DeclarativePipelineTest.java
@@ -1,7 +1,5 @@
 package plugins;
 
-import org.jenkinsci.test.acceptance.docker.DockerContainerHolder;
-import org.jenkinsci.test.acceptance.docker.fixtures.GitContainer;
 import org.jenkinsci.test.acceptance.junit.AbstractJUnitTest;
 import org.jenkinsci.test.acceptance.junit.Since;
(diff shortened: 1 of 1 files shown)


batmat:
@abayer it seems correct from the content, but shouldn't it be 3 tests? Or did you do that on purpose for performance reason or reinstating a same somewhat similar Jenkins instance each time?

src/test/java/plugins/DeclarativePipelineTest.java
```diff
@@ -0,0 +1,147 @@
...
+                "}\n"
+        );
+
+        missingToolVersionJob.sandbox.check();
+        missingToolVersionJob.save();
+
+        Build missingToolVersionBuild = missingToolVersionJob.startBuild().shouldFail();
+        String missingToolVersionConsole = missingToolVersionBuild.getConsole();
+        assertThat(missingToolVersionConsole, containsRegexp("Tool type 'maven' does not have an install of 'some-other-version' configured"));
+        assertThat(missingToolVersionConsole, not(containsRegexp("FOO is BAR")));
```
batmat: NIT: why regexp and not `containsString`?

batmat:
So apparently something fails, but the build got wiped out, can someone retrigger a PR build to see what it is, and if still present? @abayer ?

PR #225, opened 2016-11-17
[JENKINS-39828] Ignore test if plugin cannot be updated through the UI

varyvol:
[JENKINS-39828](https://issues.jenkins-ci.org/browse/JENKINS-39828)

@reviewbybees @andresrc @olivergondza


mafraba:
🐝  looks good

raul-arabaolaza:
LGTM :bee:

olivergondza:
I presume this can happen in 2 cases: Requiring unreleased plugin version or using Jenkins version too old for the required plugin version. It makes sense to me.

recena:
:bee:

andresrc:
Yes, and those cases are more common when using `uploadPlugins=false`

olivergondza:
Actually, I am not sure about this. *Using Jenkins version too old for the required plugin version* should show a warning there but should still be presented. I am afraid this will hide real problems when UI changes.

@andresrc, This can be triggered if and only if `uploadPlugins=false`.

andresrc:
@olivergondza makes sense

varyvol:
@olivergondza this only happens when uploadPlugins=false, as only then the private method tickPluginToInstall() is called.

PR #227, opened 2016-11-21
[JENKINS-39828] Ignore test if plugin cannot be updated through the UI [new]

varyvol:
Downstream on [PR #225](https://github.com/jenkinsci/acceptance-test-harness/pull/225)

[JENKINS-39828](https://issues.jenkins-ci.org/browse/JENKINS-39828)

@reviewbybees @andresrc @olivergondza

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
@@ -25,6 +25,7 @@ import org.jenkinsci.test.acceptance.update_center.PluginSpec;
 import org.jenkinsci.test.acceptance.update_center.UpdateCenterMetadata;
 import org.jenkinsci.test.acceptance.update_center.UpdateCenterMetadata.UnableToResolveDependencies;
 import org.junit.internal.AssumptionViolatedException;
+import org.openqa.selenium.NoSuchElementException;
 import org.openqa.selenium.StaleElementReferenceException;
 import org.openqa.selenium.TimeoutException;
 import org.openqa.selenium.WebElement;
@@ -232,7 +233,7 @@ public class PluginManager extends ContainerPageObject {
             for (final PluginSpec n : specs) {
                 switch (installationStatus(n)) {
                     case NOT_INSTALLED:
-                        tickPluginToInstall(n);
+                        tickPluginToInstall(n, false);
                     break;
                     case OUTDATED:
                         update.add(n);
@@ -251,7 +252,7 @@ public class PluginManager extends ContainerPageObject {
             if (!update.isEmpty()) {
                 visit(""); // Updates tab
                 for (PluginSpec n : update) {
-                    tickPluginToInstall(n);
+                    tickPluginToInstall(n, true);
                 }
                 clickButton("Download now and install after restart");
             }
@@ -263,9 +264,21 @@ public class PluginManager extends ContainerPageObject {
         return false;
     }
 
-    private void tickPluginToInstall(PluginSpec spec) {
+    private void tickPluginToInstall(PluginSpec spec, boolean updating) {
         String name = spec.getName();
-        check(find(by.xpath("//input[starts-with(@name,'plugin.%s.')]", name)));
+        WebElement pluginInstallationTick = null;
+        try {
+            pluginInstallationTick = find(by.xpath("//input[starts-with(@name,'plugin.%s.')]", name));
+        } catch (NoSuchElementException ex) {
+            if (updating) {
+                throw new AssumptionViolatedException(String.format(
+                    "Plugin %s needs to be updated but does not appear in 'updates' page",
+                    name
+                ));
+            }
+            throw ex;
+        }
+        check(pluginInstallationTick);
         final VersionNumber requiredVersion = spec.getVersionNumber();
         if (requiredVersion != null) {
             final VersionNumber availableVersion = getAvailableVersionForPlugin(name);


olivergondza:
The implementation seems the same as #225 I reverted. What is the real world use-case for this?

varyvol:
The ones specified in the previous PR, but now with confirmation that this is only happening when uploadPlugins=false.

andresrc:
@olivergondza this PR is proposing https://github.com/jenkinsci/acceptance-test-harness/pull/225 again, aa it has been checked that it only applies when `uploadPlugins=false`, as discussed there.

olivergondza:
Well, all affected code is called iff `uploadPlugins=false` - no question about that. What I object against is catching `NoSuchElementException` (that can be a symptom of a regression) and by that assuming something as unusual as an attempt to update plugin while we are using the latest version available in connected UC (what will cause the test to be silently skipped masking the potential bug). Or at least that is my speculation of what this should be for. Perhaps there is a safer way to implement that but I can not really comment until I understand what is the end user motivation for this feature.

varyvol:
The two use cases would be:

- An older core version, so that the plugin does not appear as its baseline is not satisfied.
- An UC that does not offer that version for the plugin.

If you feel like it would be safer and as those cases are related with an specific configuration for the tested environment, what do you think about making this behaviour optional depending on the value of a configuration parameter?

olivergondza:
That would probably be for the best with 2 (or maybe 3) possible values:
- if plugin exists but version is older it should skip (upstream default),
- if plugin does not exists it will skip (you use-case)
- plugin and version must be available (not sure if useful for now).

varyvol:
@olivergondza @andresrc with this last change I have made the behaviour optional through the configuration:

- If installing/updating, plugin appears and version is correct: go ahead (as upstream).
- If installing/updating, plugin appears but with older version: skip (as upstream).
- If installing/updating, plugin does not appear and no property not set or with wrong value: fail (as upstream)
- If installing, plugin does not appear and property set to installing/any: skip.
- If updating, plugin does not appear and property set to updating/any: skip.

varyvol:
@olivergondza @andresrc did you have the opportunity to review this?

olivergondza:
The implementation is tangential to my suggestion. As I understand this, it is not important if you install or update (IOW, if the plugin happened to be installed before) but rather what should you, as a person that run tests, consider to be a failure/reason to skip tests.

If the plugin required by test does not exists in an update center at all, it might or might not be of an interest to you but it is irrelevant if the plugin was unbundled from core war or not. The same goes for plugin version missing.

olivergondza:
Implementation-vise, using enum just to feed the value to 2 boolean fields is unnecessarily complicated. Stick with enum.

varyvol:
@olivergondza I am not sure I get your point. The actual reasoning is: a plugin might not appear in install/updates page but, since the person running the test is aware of that, that test is skipped.

Implementation-wise, I used an enum to allow 3 values and therefore allow the use of any if the user doesn't mind if the plugin does not appear in install or updates page. The assignment to two boolean  fields in a block is to avoid checking the value on each plugin (for a value which won't change during the whole test).

olivergondza:
> I am not sure I get your point. The actual reasoning is: a plugin might not appear in install/updates page but, since the person running the test is aware of that, that test is skipped.

What appears and what does not is a _consequence_ of the real test configuration such as:

- I test an older update center so latest plugin version might be missing.
- My update center does not contain all plugins community have.

Now, these might map more or less directly to inability to find the element on update/install pages but we should name the options based on what they represent rather than how they happen to manifest.

---

Looking into this after a while I noticed something we missed before. The testsuite should blow up (not just skip the test) when the problem is caused by dependency not explicitly required by test. Consider following situations:

- Upstream tests for LTS update center. It should point out the failure in case some plugin can not be installed because the updatecenter does not have all dependencies. IOW, test that require latest and greatest version is OK to skip but tests that detect problem in mentioned situation must not.
- Downstream test for plugin subset. Tests for plugins that are in the subset are OK to skip. Tests for plugins with dependencies missing in update center must fail to point out update center inconsistency.

andresrc:
@olivergondza From my understanding of your latest comments, a way forward could be:

#### Only for `uploadPlugins = false`

* Keep a single `skipIfPluginNotOffered` option (which consolidates the two available now) with a default value of `false`. The value of `true` could be used when testing a non-latest version (e.g., the LTS scenario you described) or using a custom UC.
* If plugin spec required in the `WithPlugins` annotation is not offered, the test would be skipped if `skipIfPluginNotOffered=true` or failed if  `skipIfPluginNotOffered=false`

#### For all values of `uploadPlugins`

After the installation of plugins, above all if a restart has been required, re-check the requested plugin specs are actually installed, and fail the test if not.

This would cover the cases if which the requested plugin do not start correctly, which is better detected for core >= 2.16 after [JENKINS-21486](https://issues.jenkins-ci.org/browse/JENKINS-21486).

wdyt?

varyvol:
@olivergondza are you fine with Andrés' approach? I'd like to have your "conceptual approval" before tackling the implementation.

olivergondza:
Hmm, I am still not confident about this. Consider following:

My test require `foo@14`.

- If there is no plugin named `foo` I have likely made a typo or the plugin was removed so the test is worth failing pointing me to the problem. For people running ATH against plugin subsets, there might be an escape hatch `IAmAwareNotAllPluginsArePresent` that would cause the test to be skipped.
- If `foo` exists but the version is older (yet to be released, stable update center, injected plugin set, etc.) the test should be skipped. I do not see a reason to have this configurable.

Now consider `foo@14` (available) depends on `bar@42`.

- If there is no plugin named `bar` then the UC is likely inconsistent and the test must fail. If the dependnecy is optional, this is the same case as `IAmAwareNotAllPluginsArePresent`.
- If `bar` exists but the version is older (yet to be released, stable update center, injected plugin set, etc.) then the UC is likely inconsistent and the test must fail.

@varyvol, @andresrc, @scoheb, Can you confirm these assumptions make sense for your ATH use-cases and move from there?

andresrc:
@olivergondza the first part is perfectly fine for me.

For the second, I have some minor comments:

With `uploadPlugins=false` exactly as you say, just that for optional dependencies:
* Not being available in the UC should not be an error (is optional).
* Being present on an older version that requested would be the `IAmAwareNotAllPluginsArePresent ` case

With `uploadPlugins=true` we should not rely on UC metadata (even though it could be customized with options) and let the Jenkins under test do the checking. We could check if the requested plugins are actually installed and always fail the test if they are not.

olivergondza:
> With uploadPlugins=false exactly as you say, just that for optional dependencies:
>   - Not being available in the UC should not be an error (is optional).
>   - Being present on an older version that requested would be the IAmAwareNotAllPluginsArePresent case

It should never happen in community update center or the LTS one (optional dependency is newer than the one in UC). Also note that `IAmAwareNotAllPluginsArePresent` was supposed to mean _no version of that plugin is available_. In other words, I have not identified a situation where it would make sense (to me) to let the user decide whether outdated plugin should cause test to skip or fail. It is just one or the other: skip the test if the version is declared on test, fail if it is a dependency.

> With uploadPlugins=true we should not rely on UC metadata (even though it could be customized with options) and let the Jenkins under test do the checking. We could check if the requested plugins are actually installed and always fail the test if they are not.

This would be the easiest thing to implement (try to install what we can, ignore problems on the way and rely on the final check) but it does not give as what we need. Like skipping tests targeted for newer versions.

---

I hate to support 2 mechanisms for plugin installation but we are not able to replace one with the other fully at this point. The least we can do is make the results of using them as similar as we can. So I am no really fond of introducing further differences.

PR #232, opened 2016-11-30
[JENKINS-40126] Add tests for jgit

raul-arabaolaza:
[JENKINS-40126](https://issues.jenkins-ci.org/browse/JENKINS-40126)

* Added support to install JGit as tool
* Added support to use jgit with `GitSCM`
* Added tests for jgit mimicking the existing ones for git

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/ToolInstallationPageObject.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/ToolInstallationPageObject.java
@@ -45,4 +45,9 @@ public @interface ToolInstallationPageObject {
      * Clients should not interact with this ToolInstalltion before updates was loaded.
      */
     String installer();
+
+    /**
+     *  Name of the concrete variant of the tool to install. IE Git or JGit
+     */
+    String variant() default "";
 }
--- a/src/main/java/org/jenkinsci/test/acceptance/po/ToolInstallationPageObject.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/ToolInstallationPageObject.java
@@ -45,9 +45,4 @@ public @interface ToolInstallationPageObject {
      * Clients should not interact with this ToolInstalltion before updates was loaded.
      */
     String installer();
-
-    /**
-     *  Name of the concrete variant of the tool to install. IE Git or JGit
-     */
-    String variant() default "";
 }
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/ant/AntInstallation.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/ant/AntInstallation.java
@@ -31,7 +31,7 @@ import org.jenkinsci.test.acceptance.po.ToolInstallationPageObject;
 public class AntInstallation extends ToolInstallation {
 
     public static void install(Jenkins jenkins, String name, String version) {
-        installTool(jenkins, AntInstallation.class, name, version);
+        ToolInstallation.installTool(jenkins, AntInstallation.class, name, version);
     }
 
     public AntInstallation(Jenkins context, String path) {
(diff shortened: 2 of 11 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/ToolInstallationPageObject.java
```diff
@@ -45,4 +45,9 @@
      * Clients should not interact with this ToolInstalltion before updates was loaded.
      */
     String installer();
+
+    /**
+     *  Name of the concrete variant of the tool to install. IE Git or JGit
+     */
+    String variant() default "";
```
olivergondza: I am not really familiar with git plugins, but the variant seems to be specific to `git-client` plugin rather than Jenkins tools installers.

src/test/java/plugins/no_docker/GitPluginNoDockerJGitTest.java
```diff
@@ -0,0 +1,116 @@
...
+import org.jenkinsci.test.acceptance.junit.WithPlugins;
+import org.jenkinsci.test.acceptance.plugins.git.GitScm;
+import org.jenkinsci.test.acceptance.plugins.git_client.JGitInstallation;
+import org.jenkinsci.test.acceptance.po.Job;
+import org.junit.Before;
+import org.junit.Test;
+
+// TODO: remove once docker based tests are run regularly
+@WithPlugins("git")
+public class GitPluginNoDockerJGitTest extends AbstractJUnitTest {
```
olivergondza: This looks like a copy of `GitPluginNoDockerTest`. Can we use junit theories to avoid the duplicity?

raul-arabaolaza:
@olivergondza Fixed the duplication, for the moment without theories, do you still believe is necessary? I can add them  but that would imply a dependency upgrade to JUnit 4.12 (AFAIU i would need named `Datapoints`)
I understand your comment about the variant. Yes, the only tool I know that would need this is git... but there is an `Add Git` section on the Global Tools page that can not be manipulated with the current code... So unless somebody changes the git-client plugin to produce two different tools, one for git, another for jgit I do not see any cleaner way to use the existing `ToolInstallationPageObject` abstraction without augmenting it with the variant. Using some specialization just for git would mean a nasty dependency on an specialized type in `ToolInstallation`.
Or I can just create a `JGitInstallation` class without using the existing  `ToolInstallationPageObject` and `ToolInstallation` classes, which somehow feels wrong.
I recognize I am stuck here... any suggestion?

olivergondza:
> Fixed the duplication, for the moment without theories, do you still believe is necessary?

Not really, it is just bit more elegant.

olivergondza:
> Or I can just create a JGitInstallation class without using the existing ToolInstallationPageObject and ToolInstallation classes, which somehow feels wrong.
I recognize I am stuck here... any suggestion?

Since this is the only case we are aware of, how about creating dedicated `GitInstallation#addTools(Jenkins, Class<GitInstallation|JGitInstallation>)` that would make the decision and create the appropriate object. That would duplicate `ToolInstallation#addTool` to some extend but that can be further refactored to share the code or I guess we can live with that as the oddities are isolated to dedicated classes.

---

I am mostly thinking aloud here but IFAICT Jenkins does not require the tool installation would provide those exact control type - most implementations just happen to. So perhaps the logic would be better placed into an instance method (j)git (or implementations with special handling) would override.

raul-arabaolaza:
> I am mostly thinking aloud here but IFAICT Jenkins does not require the tool installation would provide those exact control type - most implementations just happen to. So perhaps the logic would be better placed into an instance method (j)git (or implementations with special handling) would override.

Fully agree, I will walk that path :)

raul-arabaolaza:
Some refactor applied, waiting for feedback from CI to check that I am not breaking anything, once that is done I will continue with refinements and refactorings

raul-arabaolaza:
And with last changes build are back to normal... @olivergondza What do you think?

PR #238, opened 2016-12-05
Distribute docker container to run ATH

olivergondza:
Add Dockerfile to build container ATH can run in - no only on ci.jenkins.io.

@rtyler, I put the form-element-path-plugin on hold as it will reuse large part of this. Please review and enable on ci.jenkins.io once you are ok with this.

 @mafraba, review appreciated. Btw, this is for the first time I got ATH running in container and it seems to work fine without `SHARED_DOCKER_SERVICE=true`, at least for xvnc plugin test. Do I miss something?


docs/DOCKER.md
```diff
@@ -0,0 +1,20 @@
...
+    $ docker run -it --rm -v /var/run/docker.sock:/var/run/docker.sock -v $(pwd):/home/ath-user/ath --user ath-user jenkins/ath bash -c 'cd $HOME/ath && bash'
+    $ eval $(./vnc.sh > /dev/null)
+    $ ./run.sh firefox latest -Dmaven.test.failure.ignore=true -DforkCount=1 -B
+
+Jenkinsfile:
+
+    ... // Checkout ATH
+    sh 'docker build --build-arg=uid=$(id -u) --build-arg=gid=$(id -g) -t jenkins/ath src/main/resources/ath-container'
+    docker.image('jenkins/ath').inside {
+        sh 'eval $(./vnc.sh > /dev/null) && ./run.sh firefox latest -Dmaven.test.failure.ignore=true -DforkCount=1 -B'
```
scoheb: The redirection to /dev/null is not part of your change to Jenkinsfile above.

Jenkinsfile
```diff
@@ -0,0 +1,24 @@
...
+
+node('highram&&docker') {
+    withEnv([
+        "JAVA_HOME=${tool jdkName}",
+        "PATH+MAVEN=${tool mavenName}/bin:${env.JAVA_HOME}/bin"
+    ]) {
+        changelog scm
+
+        sh 'docker build --build-arg=uid=$(id -u) --build-arg=gid=$(id -g) -t jenkins/ath src/main/resources/ath-container'
+        docker.image('jenkins/ath').inside {
```
scoheb: You need to add this:

`String containerArgs = '-v /var/run/docker.sock:/var/run/docker.sock -v $HOME/.m2:/var/maven/.m2'`

and change it to be:

`docker.image('jenkins/ath').inside(containerArgs) {`

Otherwise, `docker ps` will not be able to connect to the docker daemon and docker-based test will fail

Jenkinsfile
```diff
@@ -0,0 +1,24 @@
...
+            sh '''
+                eval $(./vnc.sh > /dev/null 2>&1)
```
scoheb: I think it should be:

`eval $(./vnc.sh 2>/dev/null)`

otherwise the `echo export BROWSER_DISPLAY=$display` has no effect

scoheb:
I would also maybe adjust `vnc.sh` to have this code:

    if [ -x "$(which vncviewer)" ] ; then
        (vncviewer $display || vncviewer localhost$display) > /dev/null &
    fi

no need to execute vncviewer blindly if not installed.

Jenkinsfile
olivergondza: The maven repo is shared for optimization only, right?

Jenkinsfile
olivergondza: `vnc.sh`, needs its output handling reworked I guess.

Jenkinsfile
scoheb: Yes. Forgot to mention that.

mafraba:
@olivergondza wrt `SHARED_DOCKER_SERVICE` property, I guess it can work without it as long as the docker port command returns the correct IP and the docker host and mapped ports are accessible. But it shouldn't be necessary to access the containers through mapped ports in the host anymore, since they can be accessed directly thus avoiding potential connectivity issues.

Jenkinsfile
```diff
@@ -0,0 +1,25 @@
...
+def jdkName = 'jdk8'
+
+node('highram&&docker') {
+    withEnv([
+        "JAVA_HOME=${tool jdkName}",
+        "PATH+MAVEN=${tool mavenName}/bin:${env.JAVA_HOME}/bin"
+    ]) {
+        changelog scm
+
+        sh 'docker build --build-arg=uid=$(id -u) --build-arg=gid=$(id -g) -t jenkins/ath src/main/resources/ath-container'
```
scoheb: As @rtyler  suggested, here is the code to make use of docker.build():

>  def uid = sh returnStdout: true, script: "id -u | tr '\n' ' '" 
>  def gid = sh returnStdout: true, script: "id -g | tr '\n' ' '"

>  def buildArgs = "--build-arg=uid=${uid} --build-arg=gid=${gid} src/test/resources/ath-container"
>  docker.build('jenkins/ath', buildArgs)

olivergondza:
@rtyler can you create the project on ci.jenkins.io so we get this thing moving?

rtyler:
If there isn't an INFRA ticket already, please file one.

olivergondza:
@rtyler ping, https://issues.jenkins-ci.org/browse/INFRA-1070

vnc.sh
```diff
@@ -9,6 +9,8 @@ fi
 
 vncserver -kill $display
 vncserver -geometry 1750x1250 $display > /dev/null
-vncviewer localhost$display > /dev/null &
+if command -v vncviewer >/dev/null 2>&1; then
```
jglick: You should expose port 9042 or whatever it is, so people can attach a desktop `vncviewer` to the container and watch the magic happen.

Jenkinsfile
```diff
@@ -0,0 +1,30 @@
+// For ci.jenkins.io
+// https://github.com/jenkins-infra/documentation/blob/master/ci.adoc
+
+def mavenName = 'mvn'
+def jdkName = 'jdk8'
+
+node('highram&&docker') {
+    withEnv([
+        "JAVA_HOME=${tool jdkName}",
+        "PATH+MAVEN=${tool mavenName}/bin:${env.JAVA_HOME}/bin"
```
jglick: Why do we need these at all, if the build is done inside the image anyway?

Jenkinsfile
```diff
@@ -0,0 +1,30 @@
...
+    ]) {
+        changelog scm
+
+        def uid = sh returnStdout: true, script: "id -u | tr -d '\n'"
+        def gid = sh returnStdout: true, script: "id -g | tr -d '\n'"
+
+        def buildArgs = "--build-arg=uid=${uid} --build-arg=gid=${gid} src/main/resources/ath-container"
+        docker.build('jenkins/ath', buildArgs)
```
jglick: Assign the return value to a local variable and use in place of `docker.image('jenkins/ath')` below.

src/main/resources/ath-container/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+        curl \
+        docker.io \
+        imagemagick \
+        iptables \
+        firefox-esr \
+        maven \
+        openjdk-8-jdk \
+        vnc4server
+
+# Allow injecting uid and git to match directory ownership
```
jglick: This is sufficiently weird you had better document clearly why this is necessary.

Jenkinsfile
```diff
@@ -0,0 +1,30 @@
...
+
+        String containerArgs = '-v /var/run/docker.sock:/var/run/docker.sock'
```
jglick: You might need `--group-add=${gid}` to make sure `docker.sock` is writable. Unclear to me whether that is handled already by the other tricks.

docs/DOCKER.md
```diff
@@ -0,0 +1,23 @@
...
+TO-DO: Instead of depending on setting this environment variable, in the future we could try to somehow automatically detect the situation.
+
+Interactive shell:
+
+    $ docker build --build-arg=uid=$(id -u) --build-arg=gid=$(id -g) -t jenkins/ath src/main/resources/ath-container
+    $ docker run -it --rm -v /var/run/docker.sock:/var/run/docker.sock -v $(pwd):/home/ath-user/ath --user ath-user jenkins/ath bash -c 'cd $HOME/ath && bash'
+    $ eval $(./vnc.sh)
+    $ ./run.sh firefox latest -Dmaven.test.failure.ignore=true -DforkCount=1 -B
+
+Jenkinsfile:
```
jglick: This seems redundant. Anyone who cares can look up the `Jenkinsfile` for themselves.

src/main/resources/ath-container/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+# see docs/DOCKER.md for usage
+#
+
+FROM debian:sid
+MAINTAINER ogondza@gmail.com
+
+RUN apt-get -y update && \
+    apt-get install -y \
+        curl \
+        docker.io \
```
jglick: You can also just download a specific client executable.

src/main/resources/ath-container/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+# TODO seems this can be picked up from the host, which is unwanted:
+ENV XAUTHORITY /home/ath-user/.Xauthority
+
+USER ath-user
+RUN mkdir /home/ath-user/.vnc && (echo ath-user; echo ath-user) | vncpasswd /home/ath-user/.vnc/passwd
+# Default content includes x-window-manager, which is not installed, plus other stuff we do not need (vncconfig, x-terminal-emulator, etc.):
+RUN touch /home/ath-user/.vnc/xstartup && chmod a+x /home/ath-user/.vnc/xstartup
+
+# Set SUID and SGID for docker binary so it can communicate with mapped socket its uid:gid we can not control. Alternative
+# approach used for this is adding ath-user to the group of /var/run/docker.sock but that require root permission we do not
```
jglick: Hence `--group-add`. Maybe.

scoheb:
Stole your Dockerfile and part of Jenkinsfile for use in jms-messaging-plugin. Working well now. Thanks!

olivergondza:
Cool. Be prepared this will receive several updates before it is ready.

olivergondza:
Let's get this merged to so see it run and fail so we can fix as we go.

PR #246, opened 2016-12-19
[JENKINS-40540] Adding to the documentation that in order to run the …

vilacides:
…tests a Firefox 46 or older is needed.

### DIFF
--- a/docs/BROWSER.md
+++ b/docs/BROWSER.md
@@ -24,6 +24,8 @@ Therefore, to run tests with Safari, you'd execute:
 
 See `FallbackConfig.java` for how the browser is selected.
 
+Please note that since Selenium 2.x is the version used, you will need Firefox 46 or older in order to run the tests.
+
 ## Advanced Browser Configuration
 [This test harness internally uses Guice](GUICE.md) to wire tests, and that is how we control
 WebDriver. To further fine-tune how a browser is selected and configured, bind `WebDriver` to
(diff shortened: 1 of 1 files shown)


docs/BROWSER.md
```diff
@@ -24,6 +24,8 @@ Therefore, to run tests with Safari, you'd execute:
 
 See `FallbackConfig.java` for how the browser is selected.
 
+Please note that since Selenium 2.x is the version used, you will need Firefox 46 or older in order to run the tests.
+
```
andresrc: ...in order to run the tests using this browser.

raul-arabaolaza:
:bee:

mafraba:
🐝

olivergondza:
This info will get outdated fairly soon. How about pointing users to selenium support page instead?

vilacides:
@olivergondza that was my intent but there is no such a page that explains compatibility between browser versions and selenium versions 1:1. The [selenium support page](http://www.seleniumhq.org/about/platforms.jsp) talks about supporting the latest ESR (and the previous to that) which is not Firefox 46 but Firefox 45.6. That doesn't match what the support page says.

olivergondza:
I have switched to ESR a while ago as FF got broken with every other update. I do not strictly object to have this line in but it seems more helpful to say we support ESR only, as that is what selenium guys say and has a chance to save someone a ton of time.

vilacides:
@olivergondza it took me quite some time to get the right version of Firefox to run the tests and a specific version or a link to the Selenium doc would have saved me a lot of time. That is why I did this PR and I totally agree with you that is better to limit the documentation to what Selenium guarantees that will work at any time. Change pushed!

PR #247, opened 2016-12-20
[JENKINS-40567] Create test for Pipeline Shared Library stored in Github

varyvol:
[JENKINS-40567](https://issues.jenkins-ci.org/browse/JENKINS-40567)

- Create POs to configure a Pipeline Global Shared Library.
- Create test with a Pipeline which uses the Shared Library.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/workflow_multibranch/GithubBranchSource.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/workflow_multibranch/GithubBranchSource.java
@@ -16,9 +16,9 @@ import java.util.concurrent.TimeUnit;
 @Describable("GitHub")
 public class GithubBranchSource extends BranchSource {
 
-    public final Control owner = control(by.name("_.repoOwner"));
-    public final Control repository = control(by.name("_.repository"));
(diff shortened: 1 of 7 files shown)


src/main/java/org/jenkinsci/test/acceptance/plugins/workflow_multibranch/GithubBranchSource.java
```diff
@@ -15,14 +16,18 @@
 @Describable("GitHub")
 public class GithubBranchSource extends BranchSource {
 
-    public final Control owner = control("source/repoOwner");
-    public final Control repository = control("source/repository");
-    public final Control credential = control("source/scanCredentialsId");
+    public final Control owner = control(by.name("_.repoOwner"));
+    public final Control repository = control(by.name("_.repository"));
+    public final Control credential = control(by.name("_.scanCredentialsId"));
```
olivergondza: Note that the `By` form performs page object wide search while the `path` form search with the page area. In case the PA is guaranteed not to appear on page more than once, it is ok.

mafraba:
LGTM 🐝

varyvol:
Build and tests passing correctly.

armfergom:
🐝

PR #248, opened 2016-12-27
Update ATH to JDK 8 compiler settings

batmat:
@olivergondza @uhafner @jenkinsci/code-reviewers 
Rationale:
* Selenium 3.x requires it
* The core is on its way to Java 8
* Java 8 \o/

See also https://groups.google.com/forum/#!msg/jenkinsci-dev/ckHjD6es4KQ/F7DfD7bQAQAJ

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -431,8 +431,8 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>3.2</version>
         <configuration>
-          <source>1.7</source>
-          <target>1.7</target>
+          <source>1.8</source>
+          <target>1.8</target>
         </configuration>
       </plugin>
       <plugin>


uhafner:
I like this change :-)

Wouldn't it make sense to combine it with the update to Selenium 3.x. Then it is more obvious that the PR has an actual requirement. 

BTW: unless we split the ATH into two components (one for the tests and for the startup of Jenkins under test) this implies that we can't test Jenkins on Java 7 anymore. (That is ok for me, but I'm not sure if everybody is ok with this restriction.)

olivergondza:
@uhafner, `JENKINS_JAVA_HOME` allows you to run Jenkins on any JDK you like. I am not aware why it would not be possible to use older JDK than the one for ATH (except for pooled controller that uses remoting channel).

uhafner:
Ah, even better that it works with any JDK out of the box.

raul-arabaolaza:
:bee:

batmat:
@jglick sorry, missed your comment. Well, if that comment was for me, why not, but I prefer generally to separate changes to reduce scope of discussions and potential disagreements :), one step at a time. But agreed.

PR #249, opened 2016-12-29
[JENKINS-40715] Fix race condition in 'UpdateCenter#waitForInstallationToComplete'

mafraba:
Fixes [JENKINS-40715](https://issues.jenkins-ci.org/browse/JENKINS-40715).

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/Matchers.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/Matchers.java
@@ -6,7 +6,6 @@ import org.jenkinsci.test.acceptance.plugins.analysis_collector.AnalysisPlugin;
 import org.jenkinsci.test.acceptance.po.*;
 import org.openqa.selenium.By;
 import org.openqa.selenium.NoSuchElementException;
-import org.openqa.selenium.StaleElementReferenceException;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 
@@ -59,7 +58,7 @@ public class Matchers {
             @Override
             public boolean matchesSafely(WebDriver item) {
                 try {
-                    item.findElements(selector);
+                    item.findElement(selector);
                     return true;
                 } catch (NoSuchElementException _) {
                     return false;
--- a/src/main/java/org/jenkinsci/test/acceptance/po/UpdateCenter.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/UpdateCenter.java
@@ -2,7 +2,6 @@ package org.jenkinsci.test.acceptance.po;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.google.common.collect.Iterables;
-
 import org.apache.commons.lang3.ArrayUtils;
 import org.jenkinsci.test.acceptance.Matchers;
 import org.jenkinsci.test.acceptance.update_center.PluginSpec;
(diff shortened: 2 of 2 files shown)


src/main/java/org/jenkinsci/test/acceptance/Matchers.java
```diff
@@ -59,7 +58,7 @@ public void describeMismatchSafely(WebDriver item, Description mismatchDescripti
             @Override
             public boolean matchesSafely(WebDriver item) {
                 try {
-                    item.findElements(selector);
+                    item.findElement(selector);
```
oleg-nenashev: How is it related to the issue?

oleg-nenashev:
🐝 in general, but I'm not sure I understand one change in matchers

src/main/java/org/jenkinsci/test/acceptance/po/UpdateCenter.java
```diff
@@ -71,7 +70,8 @@ public boolean isInstalled(String pluginShortName) throws InstallationFailedExce
      */
     public boolean waitForInstallationToComplete(final PluginSpec... specs) throws InstallationFailedException {
         open();
-        waitFor(driver, not(Matchers.hasContent("Pending")), 60); // Wait for all plugins to get installed
+        // Wait for all plugins to get installed
+        waitFor(driver, not(Matchers.hasElement(by.xpath("//*[@id='log']//*[contains(.,'Pending') or contains(.,'Installing')]"))), 60);
```
varyvol: Matchers.hasContent() accepts a Pattern, maybe it could be used to avoid depending on an xpath (especially because of the id)?

src/main/java/org/jenkinsci/test/acceptance/Matchers.java
mafraba: Indeed this matcher was flawed, and we use it in the new implementation of the check so we needed to fix it.

The problem in question is that `findElements` (plural) never throws a `NoSuchElementException`, just returns an empty list if no matches.

src/main/java/org/jenkinsci/test/acceptance/po/UpdateCenter.java
vilacides: Why was the matcher for hasContent("Pending") causing a race condition? How is that different from asserting the xpath element that contains Pending?

mafraba: The 'Installing' string also appears in the title of the page, that's why we need to be more precise, and look for it only inside the "log".

mafraba: @vilacides checking only for 'Pending' is not sufficient, and to check for the 'Installing' string we need to be more precise as explained just above, hence the xpath.

varyvol: Does it keep appearing in the title after it doesn't in the log?

mafraba: Yes

varyvol: Fair enough then!

vilacides: Do they both disappear at the same time? If not, can we use that to fail fast the test?

varyvol:
🐝

vilacides:
🐝

raul-arabaolaza:
:bee:

mafraba:
@reviewbybees done

batmat:
@olivergondza we would really like this fix out, do you plan to release it?
I can handle if you agree.

Thanks!

Sicarri:
Ping @olivergondza

olivergondza:
1.47 released.

batmat:
Thanks Oliver!

PR #250, opened 2017-01-03
[JENKINS-40770] Take into account optional dependencies during installation process.

varyvol:
[JENKINS-40770](https://issues.jenkins-ci.org/browse/JENKINS-40770)

- Analyse optional dependencies for the set of required plugins.
- Check which ones are required to be installed through @WithPlugins annotation.
- Install them before their dependant plugins.

To gather some context: [ATH PR #216](https://github.com/jenkinsci/acceptance-test-harness/pull/216)

@reviewbybees @olivergondza @andresrc

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
@@ -190,7 +190,8 @@ public class PluginManager extends ContainerPageObject {
             if (!someChangeRequired) {
                 return false;
             }
-            List<PluginMetadata> pluginToBeInstalled = ucmd.get().transitiveDependenciesOf(jenkins, Arrays.asList(specs));
+            List<PluginMetadata> pluginToBeInstalled = ucmd.get().transitiveDependenciesIncludingOptionalsOf(jenkins, Arrays.asList(specs));
+
             for (PluginMetadata newPlugin: pluginToBeInstalled) {
                 final String name = newPlugin.getName();
                 String requiredVersion = candidates.get(name);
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
```diff
@@ -213,35 +214,32 @@ public boolean installPlugins(final PluginSpec... specs) throws UnableToResolveD
...
-            System.out.println("Plugins to be updated: " + update);
-            if (!update.isEmpty()) {
-                visit(""); // Updates tab
-                for (PluginSpec n : update) {
-                    tickPluginToInstall(n);
+                if (!optionalDependenciesToInstall.isEmpty()) {
+                    installOrUpdatePlugins(optionalDependenciesToInstall);
+
+                    // Jenkins will be restarted if necessary
+                    new UpdateCenter(jenkins).waitForInstallationToComplete(optionalDependenciesToInstall.toArray(new PluginSpec[]{}));
```
mafraba: Not sure if the "restart" part is necessary, or we could just "wait". In that case maybe a refactor of the `waitForInstallationToComplete` method could save some unneeded restarts.

But anyway, we maybe really need the restart, not really sure ...

mafraba:
🐝  LGTM

src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
varyvol: If the final installation screen (that where a summary of all the installed plugins, including the optional ones and the other ones, appears) still displays the error that forces a restart, I think it would be fine restarting just then, as all the extension points should be functional afterwards.

If it doesn't, I'm afraid we need to restart here to be sure the plugin is available before the other one is installed.

src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
```diff
@@ -213,35 +214,32 @@ public boolean installPlugins(final PluginSpec... specs) throws UnableToResolveD
                 }
             }
         } else {
-            visit("available");
+            List<PluginSpec> allRequiredPlugins = Arrays.asList(specs);
+            Map<Integer, List<String>> optionalDependencies = ucmd.get().optionalDependenciesOf(jenkins, allRequiredPlugins);
```
andresrc: :bug:

For the `uploadPlugins = false` case you can have issues using the provided update center metadata, as it may not be the same the instance is using, it would be the one to use if `uploadPlugins = true`.

olivergondza:
Is still the relevant after the proxy update center (https://github.com/jenkinsci/acceptance-test-harness/pull/379)? I do not think so...

PR #251, opened 2017-01-04
[WIP] [JENKINS-38307] Stop installing detached plugins

olivergondza:
Replacing https://github.com/jenkinsci/acceptance-test-harness/pull/191


olivergondza:
No new failures for 2.x, There should be no problem with 1.x as no plugin was detached since 1.577 and I doubt people care for such an old release to be supported by ATH.

PR #84, opened 2016-03-14
[FIXED JENKINS-33520] make ATH work on windows and other issues.

jtnord:
ATH was just so badly broken on windows :cry: 

[JENKINS-33520](https://issues.jenkins-ci.org/browse/JENKINS-33520)

I beleive I have fixed most of the issues.
There is an open issue of how to detect the OS if running Jenkins remotely (however the current situation assumes it is run on the same OS type with the same tools (@native looks on the system running the ATH not the Jenkins system) - so this is a longer issue - and would significantly add to the test run time for the case where the OS does not match) - so is considered out of scope.

Many tests are skipped as Jenkins will not restart itself when started locally on Windows.  LocalController could possibly be enhanced to do a "manual" restart by stopping Jenkins and then starting it again.

Current state: 

- [x] Maven seem like genuine test failures (Maven has not been correctly added to the path as it is trying to execute mvn.bat and failing) - ([JENKINS-33693](https://issues.jenkins-ci.org/browse/JENKINS-33693))
- [x] some tests are still trying to execute bash scripts on windows :-/
- [ ] ~~I seem to have broken docker portmap (Failed to figure out port map 80)~~ `-DreuseForks=false` fixes this.  Something to investigate later.
- [ ] Firefox driver is unreliable (upgraded driver but still has issues)
- [x] Incorrect assumptions Docker
- [x] Incorrect assumptions in Job page object
- [x] @Native should not require cygwin
- [x] CIFSPublishPluginTest
- [x] PublishOverSSHPluginTest
- [x] CompressArtifactsPluginTest
- [x] DeployPluginTest
- [x] JacocoPluginTest - (fixed by [JENKINS-33693](https://issues.jenkins-ci.org/browse/JENKINS-33693))
- [x] JavadocPluginTest (fixed by [JENKINS-33693](https://issues.jenkins-ci.org/browse/JENKINS-33693))
- [x] LdapPluginTest
- [x] MavenPluginTest
- [x] CopyArtifactPluginTest (fixed by [JENKINS-33693](https://issues.jenkins-ci.org/browse/JENKINS-33693))
- [ ] ~~NodeLabelParameterPluginTest~~  not an issue in windows - tests are genuinley failing on Linux too ([JENKINS-33746](https://issues.jenkins-ci.org/browse/JENKINS-33746)).
- [x] Subversion_Version154_PluginTest
[...]

PR #252, opened 2017-01-04
[JENKINS-40808] Using docker-fixtures repository

jglick:
Downstream of https://github.com/jglick/docker-fixtures/pull/1.

@reviewbybees


src/main/java/org/jenkinsci/test/acceptance/docker/DynamicDockerContainer.java
```diff
@@ -1,36 +0,0 @@
...
-import java.nio.charset.StandardCharsets;
-
-import org.apache.commons.io.FileUtils;
-
-/**
- * A DockerContainer whose <code>Dockerfile</code> is not static but requires some knowledge of the environment it is going to run in.
- *
- * Before the image is created {@link #process(String)} will be called which can be used to replace place holders with given values or manipulate the docker file in anyother way.
- */
-public class DynamicDockerContainer extends DockerContainer {
```
olivergondza: I wanted to get rid of this in favor of using build arguments. Would be great doing that before other projects start to depend on this behavior.

olivergondza:
I am -0.5 on sharing the actual Dockerfiles. I am not confident we will be able to preserve needed compatibility in case we need to change some of the fixtures (update/rewrite/adding features).

src/main/java/org/jenkinsci/test/acceptance/docker/DynamicDockerContainer.java
jglick: From what I can tell, @jtnord introduced this without any actual use cases, in #84. I have no opinion about it, and whether it should stay or be deleted seems orthogonal to this refactoring.

jglick:
> I am -0.5 on sharing the actual Dockerfiles.

I am actually reusing `JavaContainer` (and thus by extension `SshdContainer`) from https://github.com/jenkinsci/mercurial-plugin/issues/93. Yes I could copy all that implementation if I had to, but I would really rather not.

And any functional tests which require a container for running an agent—typically because the plugin runs some special external commands as a build step, rather than making a socket connection to some kind of server, so in practice a large portion of plugins out there—will also need `JavaContainer`.

> I am not confident we will be able to preserve needed compatibility

Is it really a problem to deprecate a container and introduce a replacement? For that matter, this is just a test utility, so backward-incompatible changes are not that big a deal—you add some release notes, and if and when a plugin developer decides to upgrade the test dependency, either their tests continue to pass or they get a failure and figure out what they need to adapt to.

What is more irritating is that derived `Dockerfile`s currently need to hardcode the tag as produced by `Docker.getDockerFileHash`. Probably `build` could retag the newly produced image as `latest`, or simply delete `getDockerFileHash` and stop trying to use clever tags.

jglick:
BTW these comments would probably be better placed in https://github.com/jglick/docker-fixtures/pull/1.

src/main/java/org/jenkinsci/test/acceptance/docker/DynamicDockerContainer.java
jtnord: I found another way to accomplish what I required,  but given I had written it left it incase it was any use in the future.

jglick:
@olivergondza ping?

scoheb:
But do you not need https://issues.jenkins-ci.org/browse/HOSTING-254 first?

batmat:
I suppose this can be merged and move forward now https://github.com/jglick/docker-fixtures has been forked as https://github.com/jenkinsci/docker-fixtures right?

olivergondza:
@batmat, there do not seem to be a release, though.

batmat:
Right @olivergondza but I was looking at the current PR and saw a dependency against 1.0-SNAPSHOT so that should still work. Though granted, we probably don't want ATH release be blocked because of a non-RELEASE dependency.

jglick:
Yes I need to cut a release and integrate—on my to-do list.

jglick:
Test failures look unrelated, so merging.

PR #255, opened 2017-01-11
[Jenkins 40947] Create acceptance test for config-file-provider plugin

varyvol:
[JENKINS-40947](https://issues.jenkins-ci.org/browse/JENKINS-40947)

- Create POs for custom and Maven Settings configuration files.
- Include support for configuration files inside folders.
- Create acceptance test.

@reviewbybees


varyvol:
@imod this is the test that made me notice the failure when updating.

imod:
cool, thanks - when are these tests executed?

andresrc:
@imod they are run (at least) [here](https://jenkins.ci.cloudbees.com/job/core/job/acceptance-test-harness/) and [here](acceptance-test-harness-stable)

armfergom:
Looks good :bee:

src/test/java/plugins/ConfigFileProviderTest.java
```diff
@@ -0,0 +1,95 @@
...
+import org.jenkinsci.test.acceptance.po.Build;
+import org.jenkinsci.test.acceptance.po.WorkflowJob;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.hamcrest.Matchers.containsString;
+import static org.junit.Assert.assertThat;
+
+/**
+ * Tests config-file-provider plugin inside a Pipeline. 2.15.5 includes a major refactor as well as some bug fixes
```
olivergondza: It does not seem to be testing any features on 2.15.5. There is no reason to require it.

varyvol:
@olivergondza I added the requirement because versions 2.15.3 and 2.15.4 has some problems which will actually make the test to fail.

olivergondza:
> because versions 2.15.3 and 2.15.4 have some problems which will actually make the test fail.

That is the reason why we have these tests. To fail when there is a problem with the plugin.

varyvol:
@olivergondza you are right, I thought it did not make any sense to include a test for versions that are gonna fail for sure, but it's true that some people could have included those versions without being aware of the issues.

I have removed the version.

PR #256, opened 2017-01-11
[JENKINS-40123] Login PO includes methods to wait for login success or failure

varyvol:
[JENKINS-40123](https://issues.jenkins-ci.org/browse/JENKINS-40123)

Include in Login PO methods to verify if the login attempt has succeeded or failed.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/Matchers.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/Matchers.java
@@ -178,6 +178,44 @@ public class Matchers {
         };
     }
 
+    public static Matcher<Login> loggedInAs(final String user) {
+        return new Matcher<Login>("has logged in user %s", user) {
+            @Override
+            public boolean matchesSafely(final Login login) {
+                try {
+                    login.find(by.href("/user/" + user));
+                    return true;
+                } catch (NoSuchElementException e) {
+                    return false;
+                }
+            }
+
+            @Override
+            public void describeMismatchSafely(final Login item, final Description desc) {
+                desc.appendText(user + " is not logged in.");
+            }
+        };
+    }
+
+    public static Matcher<Login> hasInvalidLoginInformation() {
+        return new Matcher<Login>("has invalid login information message") {
+            @Override
+            public boolean matchesSafely(final Login login) {
+                try {
+                    login.find(by.xpath("//div[contains(text(), 'Invalid login information. Please try again.')]"));
+                    return true;
+                } catch (NoSuchElementException e) {
+                    return false;
+                }
+            }
+
+            @Override
+            public void describeMismatchSafely(final Login item, final Description desc) {
+                desc.appendText("There is no invalid login message.");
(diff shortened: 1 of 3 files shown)


olivergondza:
This is used to configure Jenkins extensively and asking every client to check for success explicitly seems wrong. Can we throw from `doLogin` in case it does not succeed? I am not sure we need testing for failures at this point ...

varyvol:
I don't know if there are tests (here or in other projects extending this one) for failures, but in case there are, such change would break them.

As this is something that did not exist up to now, I understand it is not needed for the common situations/tests. But it can be useful sometimes, so I consider more appropriate to let those cases use the functionality and the other cases not to worry about it, instead of risking breaking existing and properly working tests.

What do you think?

olivergondza:
Well, there are use-cases or PO methods in ATH that are used lot more often as supporting routines then for testing itself. Consider plugin installation for example: I do not know if there is any test for that but every other test rely on that functionality to work (and as a part of that fail fast and aloud when it stops working). I expect lot more tests will be written in a way *Log me in so I can ...* rather that *Let's see if this login attempt will work* so we should default to throw an exception in case the login fails not to force the majority of the callers to check that.

mafraba:
I see that most tests that invoke the 'doLogin' implicitly expect it to success. But I think it's also a valid case to try to login and check that it fails (e.g. when testing authorization itself).

Plus, I don't think this change introduces the need to check for success. Things should behave as they have done until now if no success/failure check added (and will probably fail later in the test if the login went bad).

IMHO this is valid 🐝

varyvol:
@olivergondza I think plugin installation is a different situation, more like a "utility class". In my opinion a PO should only perform actions that the original page allows and not make any asserts. Those should be in the test side, where there is an actual expected behaviour. 

I see including methods with asserts inside a PO if it is something which really eases a vast set of tests and is called explicitly (i.e: Build.shouldSucceed()). Even if it's a new functionality and including such asserts provides an extra security without the callers intervention, I might agree. But I don't think changing an actual behaviour, which might make some tests to break, is a good choice.

olivergondza:
As I look at it it would certainly break tests.

Can you refactor the should methods to `Matchers` and introduce a method to login and fail to be used by tests that use login as a part of setup?

varyvol:
@olivergondza I guess you mean a method to login with the actual check for its success so it can be used without having to call the check explicitly? I will, of course!

And also the should methods are already using Matchers, or what do you mean?

olivergondza:
I suggest they would be matchers instead so you get waiting and negation for free like:

`assertThat(page, loggedInAs("ogondza"))`/`assertThat(page, not(loggedIn()))`

varyvol:
@olivergondza changes done.

varyvol:
@olivergondza did you have the opportunity to review the latest changes?

varyvol:
@olivergondza sorry, but were you able to review the new changes?

PR #260, opened 2017-01-23
Fix Check version for existing controller

FranckG:
existing controller:
Fix an error in getHeaderField for getting Jenkins version


olivergondza:
Addressed in sibling PR (https://github.com/jenkinsci/acceptance-test-harness/pull/261), my bad.

PR #263, opened 2017-01-26
[JENKINS-41484] Create test for antisamy-markup-formatter plugin

varyvol:
[JENKINS-41484](https://issues.jenkins-ci.org/browse/JENKINS-41484)

Acceptance test for Antisamy Markup Formatter plugin.

Also includes the possibility of changing the Markup Formatter in Global Security config as well as setting a description for TopLevelItem (Folders and Jobs).

@reviewbybees

PR #266, opened 2017-02-02
[JENKINS-41484] Fix wrong comment in Javadoc

varyvol:
Fixes a wrong comment introduced by [PR #263](https://github.com/jenkinsci/acceptance-test-harness/pull/263) 

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/TopLevelItem.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/TopLevelItem.java
@@ -58,8 +58,7 @@ public abstract class TopLevelItem extends ContainerPageObject {
     }
 
     /**
-     * Renames the job. Opens the configuration section, sets the name and saves the form. Finally the rename is
-     * confirmed.
+     * Changes the description. Opens the configuration section, sets the description and saves the form.
      *
      * @param description the description of the job
      * @param withCodeMirror if description field uses CodeMirror or not (depending on Markup Formatter)


src/main/java/org/jenkinsci/test/acceptance/po/TopLevelItem.java
```diff
@@ -58,8 +58,7 @@ public TopLevelItem(PageObject context, URL url, String name) {
     }
 
     /**
-     * Renames the job. Opens the configuration section, sets the name and saves the form. Finally the rename is
-     * confirmed.
+     * Changes the description. Opens the configuration section, sets the description and saves the form.
```
uhafner: I should have noticed this before but wouldn't it be also better if the method is called setDescription? Or changeDescription? Typically, a method name contains a verb...

varyvol: @uhafner in the ATH I think it's an usual pattern not to put it, especially when it's directly referring to a field in the page.

uhafner: Hmm, do you have other examples? In Job and TopLevelItem 99% of all methods contain a verb.

olivergondza: There is particular convention for setting form elements. Someone prefer public `Control`s, I prefer verbless methods, etc. I am fine with this.

varyvol: GithubBranchSource.owner(xxx) for example.

uhafner: Well, I mean in a class that is not yours ;-) your method is the first one without verb in the Job class. But anyway, since we (sadly) don't have a coding style in this project field free to keep it as it is. 
 
BTW: GithubBranchSource is even worse: there is a public field *and* an method with the same name: which one should I choose?

PR #273, opened 2017-02-09
Allow plugin overrides to be specified with safe environment variable names

jglick:
While

```bash
env x.jpi=/tmp/x.hpi mvn test
```

may work,

```bash
env x-y.jpi=/tmp/x-y.hpi mvn test
```

does not, at least in Bash 4.3.46 (Ubuntu Yakkety’s current default), with Maven 3.3.9. (Noticed while testing https://github.com/jenkinsci/ssh-agent-plugin/pull/18.) So let people use

```bash
X_Y_JPI=/tmp/x-y.hpi mvn test
```

instead.

@reviewbybees


olivergondza:
4.4.11 does not even propagate `x.jpi`. While I do not like how this is getting more and more complex, this seems like a way to go.

jglick:
> Are you sure you are using the `env` command

Yes, as written in the description.

olivergondza:
I am +0 on this as it fixes that for you, though the property/envvar duality in ATH is causing nothing but problems. In this case, using properties would avoid the problem just fine.

I have created https://issues.jenkins-ci.org/browse/JENKINS-42078 to track this down. It seems to me accepting the value from both envvar and property seems like the simplest way to go.

jglick:
Will put this on hold while we figure out if that is the right way forward.

jglick:
Since nothing is happening with JENKINS-42078, I would suggest merging this as it solves a problem in the current code and seems to harm nothing.

jglick:
cf. #353

PR #275, opened 2017-02-14
[JENKINS-42017] fix jstack invocation

mafraba:
Details in [JENKINS-42017](https://issues.jenkins-ci.org/browse/JENKINS-42017)

@reviewbybees /cc @olivergondza 

Whenever this gets merged, please also do a new release so that dependent projects can bring in the fix

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/controller/LocalController.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/controller/LocalController.java
@@ -4,20 +4,17 @@ import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import javax.inject.Inject;
 import javax.inject.Named;
-import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
-import java.io.InputStreamReader;
-import java.io.StringWriter;
 import java.lang.reflect.Field;
 import java.net.ServerSocket;
+import java.nio.charset.Charset;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.logging.Logger;
 
 import org.apache.commons.io.FileUtils;
-import org.apache.commons.io.IOUtils;
 import org.codehaus.plexus.util.Expand;
 import org.codehaus.plexus.util.StringUtils;
 import org.jenkinsci.test.acceptance.junit.FailureDiagnostics;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/test/acceptance/controller/LocalController.java
```diff
@@ -374,10 +375,15 @@ private void diagnoseFailedLoad(Exception cause) throws IOException {
...
+                File tmpFile = File.createTempFile("jstack-output", ".tmp");
+                tmpFile.deleteOnExit();
+                Process jstack = new ProcessBuilder("jstack", String.valueOf(pid))
+                        .redirectErrorStream(true)
+                        .redirectOutput(ProcessBuilder.Redirect.to(tmpFile))
+                        .start();
                 if (jstack.waitFor() == 0) {
                     StringWriter writer = new StringWriter();
-                    IOUtils.copy(jstack.getInputStream(), writer);
+                    IOUtils.copy(new FileInputStream(tmpFile), writer);
```
jtnord: Seems more logical to use `IOUtuls.readFileToString(tmpFile, Charset.defaultCharset())`

src/main/java/org/jenkinsci/test/acceptance/controller/LocalController.java
```diff
@@ -374,10 +375,15 @@ private void diagnoseFailedLoad(Exception cause) throws IOException {
             }
 
             try {
-                Process jstack = new ProcessBuilder("jstack", String.valueOf(pid)).start();
+                File tmpFile = File.createTempFile("jstack-output", ".tmp");
+                tmpFile.deleteOnExit();
```
jtnord: Should be explicitly deleted instead of adding shutdown hooks as this is not needed after the file is read.

src/main/java/org/jenkinsci/test/acceptance/controller/LocalController.java
mafraba: Right, changing this

src/main/java/org/jenkinsci/test/acceptance/controller/LocalController.java
mafraba: Can explicitly delete it also, but I'd also leave the hook in case some error happens before deleting it.

olivergondza:
We can save a lot of these troubles delegating to https://github.com/olivergondza/dumpling, which I humbly propose now. If there are no objections, I will provide a fix.

mafraba:
@olivergondza if it fixes the issue please go ahead

olivergondza:
Replacing with https://github.com/jenkinsci/acceptance-test-harness/pull/276

PR #276, opened 2017-02-14
[FIXED JENKINS-42017] Delegate jstack invocation to dumpling

olivergondza:
The particular problem is addressed in https://github.com/olivergondza/dumpling/blob/master/core/src/main/java/com/github/olivergondza/dumpling/factory/PidRuntimeFactory.java#L67-L77.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/controller/LocalController.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/controller/LocalController.java
@@ -4,11 +4,10 @@ import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import javax.inject.Inject;
 import javax.inject.Named;
-import java.io.BufferedReader;
+import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
-import java.io.InputStreamReader;
-import java.io.StringWriter;
+import java.io.PrintStream;
 import java.lang.reflect.Field;
 import java.net.ServerSocket;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/test/acceptance/controller/LocalController.java
```diff
@@ -374,12 +375,12 @@ private void diagnoseFailedLoad(Exception cause) throws IOException {
...
-                if (jstack.waitFor() == 0) {
-                    StringWriter writer = new StringWriter();
-                    IOUtils.copy(jstack.getInputStream(), writer);
-                    return writer.toString();
-                }
+                ThreadDumpRuntime runtime = new PidRuntimeFactory().fromProcess(pid);
+                ByteArrayOutputStream baos = new ByteArrayOutputStream();
+                PrintStream printStream = new PrintStream(baos);
+                runtime.toString(printStream, ModelObject.Mode.MACHINE);
+                printStream.close();
```
scoheb: Do we need to put this in a finally block?

olivergondza: The only point for that stream to exist is to collect the output. `close()` effectively flushes the writer to ensure whole content is written to `toString()` returns everything. If exception gets propagated we do no longer care for the output so flushing is not necessary.

mafraba:
looks good!

batmat:
@olivergondza I'm releasing, please shout if you disagree? Thanks!

batmat:
1.51 is out

PR #278, opened 2017-02-20
Use the selenium driver to retrieve the console text

Vlatombe:
Otherwise the call doesn't work with secured Jenkins instance.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Build.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Build.java
@@ -141,6 +141,10 @@ public class Build extends ContainerPageObject {
         return url("consoleFull");
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/Build.java
```diff
@@ -150,13 +154,7 @@ public void openStatusPage() {
...
 
     public String getConsole() {
-        visit(getConsoleUrl());
-
-        try {
-            return IOUtils.toString(url("consoleText").openStream());
-        } catch (IOException ex) {
-            throw new AssertionError(ex);
-        }
+        return visit(getConsoleTextUrl()).getPageSource();
```
jglick: This is now returning HTML. It should be returning plain text.

PR #280, opened 2017-03-01
Allow writing ATHs for the install wizard

alvarolobato:
@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/junit/WithInstallWizard.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/junit/WithInstallWizard.java
@@ -1,0 +1,56 @@
+package org.jenkinsci.test.acceptance.junit;
+
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import org.jenkinsci.test.acceptance.controller.JenkinsController;
+import org.jenkinsci.test.acceptance.controller.LocalController;
+import org.junit.internal.AssumptionViolatedException;
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+import com.google.inject.Inject;
+
+/**
+ * Enables the install wizard to run the test. This is only possible for LocalControllers. Otherwise the test is
+ * skipped.
+ */
+@Retention(RUNTIME)
+@Target({ TYPE })
+@Inherited
+@Documented
+@RuleAnnotation(value = WithInstallWizard.RuleImpl.class, priority = -10) // Run before Jenkins startup
+public @interface WithInstallWizard {
+    
+    public class RuleImpl implements TestRule {
+        @Inject
+        JenkinsController controller;
+
+        @Override
+        public Statement apply(final Statement base, final Description d) {
+            return new Statement() {
+                @Override
+                public void evaluate() throws Throwable {
+                    enableWizard(d.getAnnotation(WithInstallWizard.class));
+                    enableWizard(d.getTestClass().getAnnotation(WithInstallWizard.class));
+                    base.evaluate();
+                }
+
+                private void enableWizard(WithInstallWizard n) {
+                    if (n==null) return;
+                    if (controller instanceof LocalController) {
+                        ((LocalController) controller).setRunInstallWizard(true);
+                    } else {
+                        throw new AssumptionViolatedException("Testing the setup wizard is only supported if a LocalController is in use. Test will be skipped.");
+                    }
+                }
+            };
+        }
+    }
+}
--- a/src/main/java/org/jenkinsci/test/acceptance/junit/WithInstallWizard.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/junit/WithInstallWizard.java
@@ -1,6 +1,7 @@
 package org.jenkinsci.test.acceptance.junit;
 
 import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.ElementType.METHOD;
 import static java.lang.annotation.RetentionPolicy.RUNTIME;
 
 import java.lang.annotation.Documented;
@@ -22,7 +23,7 @@ import com.google.inject.Inject;
  * skipped.
  */
 @Retention(RUNTIME)
-@Target({ TYPE })
+@Target({ TYPE, METHOD })
 @Inherited
 @Documented
 @RuleAnnotation(value = WithInstallWizard.RuleImpl.class, priority = -10) // Run before Jenkins startup
--- a/src/main/java/org/jenkinsci/test/acceptance/po/WizardCustomizeJenkins.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/WizardCustomizeJenkins.java
@@ -1,0 +1,98 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2017 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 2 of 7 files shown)


raul-arabaolaza:
:bee: Given tests run successfully

olivergondza:
Can you please provide some actual tests to see how this should be used. Currently, the controller spins Jenkins without knowing what test it will run to turn this on selectively. Turning it on for the whole testsuite will break the rest of the tests. There seems to require a bit of a redesign for controllers.

alvarolobato:
@olivergondza added the happy path for the install wizard, with suggested plugins and selecting "pipeline"

src/main/java/org/jenkinsci/test/acceptance/po/WizardCustomizeJenkins.java
```diff
@@ -0,0 +1,74 @@
...
+ *
+ */
+public class WizardCustomizeJenkins extends PageObject {
+
+    public WizardCustomizeJenkins(Jenkins jenkins) {
+        super(jenkins.injector, jenkins.url(""));
+    }
+
+    public WizardCustomizeJenkins doInstallSuggested() {
+        By suggestedButtonLocator = By.partialLinkText("Install suggested plugins");
```
raul-arabaolaza: I would suggest using `by` if possible, `by` is an instance of [ByFactory](https://github.com/jenkinsci/acceptance-test-harness/blob/69b1fae7987c219022d999f45e5dee44e8a69524/src/main/java/org/jenkinsci/test/acceptance/ByFactory.java), so if you need a new selector and simply use it via Selenium in your PO the new selector can not be reused, but if you add it to `ByFactory` it is available on all PO, also you can add some validation or error control logic easily without polluting your PO

src/main/java/org/jenkinsci/test/acceptance/po/WizardCustomizeJenkins.java
```diff
@@ -0,0 +1,74 @@
...
+        return this;
+    }
+
+    @SuppressWarnings("deprecation") // FIXME: change the detection method when we have id or name in the frame.
+    public void shouldFinishInstallSuccessfully() {
+        waitForCond(new Callable<Boolean>() {
+            @Override
+            public Boolean call() throws Exception {
+                try {
+                    WebElement element = driver.switchTo().defaultContent().switchTo().frame(1).findElement(By.name("username"));
```
raul-arabaolaza: IMHO better use  `by.name` instead of `By.name`

src/main/java/org/jenkinsci/test/acceptance/po/WizardCustomizeJenkins.java
```diff
@@ -0,0 +1,74 @@
...
+    public WizardCustomizeJenkins doInstallSuggested() {
+        By suggestedButtonLocator = By.partialLinkText("Install suggested plugins");
+        waitFor(suggestedButtonLocator, 30);
+        Control installSuggested = control(suggestedButtonLocator);
+        installSuggested.click();
+        return this;
+    }
+
+    public WizardCustomizeJenkins doSelectPluginsToInstall() {
+        By selectPluginsLocator = By.partialLinkText("Select plugins to install");
```
raul-arabaolaza: Same as above

src/main/java/org/jenkinsci/test/acceptance/junit/JenkinsAcceptanceTestRule.java
```diff
@@ -56,6 +70,14 @@ public void evaluate() throws Throwable {
 
                 injector.injectMembers(this);
 
+                if (enableInstallWizard) {
```
olivergondza: Note that in case of pooled controller it can be started already so this will have no effect.

src/main/java/org/jenkinsci/test/acceptance/junit/JenkinsAcceptanceTestRule.java
```diff
@@ -38,6 +40,18 @@
  * @author Kohsuke Kawaguchi
  */
 public class JenkinsAcceptanceTestRule implements MethodRule { // TODO should use TestRule instead
+
+    private boolean enableInstallWizard;
```
olivergondza: I do not like this is passed through the rule. You are injecting the controller anyway so you can configure that directly on the controller.

src/main/java/org/jenkinsci/test/acceptance/po/WizardCustomizeJenkins.java
alvarolobato: It doesn't work. It is a different frame

src/main/java/org/jenkinsci/test/acceptance/junit/JenkinsAcceptanceTestRule.java
alvarolobato: Yes the test is only going to be executed in the case of LocalController for all the other cases where we cannot control the start of the instance or ensure insolation the test will be skiped. See line 77

src/main/java/org/jenkinsci/test/acceptance/po/WizardCustomizeJenkins.java
alvarolobato: I can do that, but in this case I'm not creating a new selector, just using an existing one... or maybe I'm not correctly understanding you suggestion.

src/main/java/org/jenkinsci/test/acceptance/po/WizardCustomizeJenkins.java
alvarolobato: nevermind, I misread your comment. I'm changing it. What is the reason for that preference?

src/main/java/org/jenkinsci/test/acceptance/junit/JenkinsAcceptanceTestRule.java
alvarolobato: That would would force every test to check if they are running on a LocalController or not and  decide if the test should be skipped. By doing it this way we centralice it and make the decision transparent to the test.

src/main/java/org/jenkinsci/test/acceptance/po/WizardLogin.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /**
+     * Gets the generated password from the file on disk
+     * 
+     * @return the read password
+     * @throws IOException if there is an IO error 
+     */
+    public String getPassword() throws IOException {
+        File passwordFile = new File(((LocalController) controller).getJenkinsHome(), "secrets/initialAdminPassword");
```
amuniz: Unsafe cast?

src/test/java/core/InstallWizardTest.java
```diff
@@ -0,0 +1,108 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2014 Red Hat, Inc.
```
amuniz: :)

src/test/java/core/InstallWizardTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+public class InstallWizardTest extends AbstractJUnitTest {
+
+    public InstallWizardTest() {
+        super();
+        rules.setEnableInstallWizard(true);
+    }
+
+    @Before
+    public void setUp() {
```
amuniz: remove it?

src/main/java/org/jenkinsci/test/acceptance/po/WizardCustomizeJenkins.java
raul-arabaolaza: Because `by.name` has been created explicitly to be used against the jenkins ui, and if for some reason something changes in the UI that brokes the name convention is likely that `by.name` will be modified accordingly. Relying on `By` means that your new tests will have to be fixed particularly, so IMO is a litle more robust to use `by.name`

src/main/java/org/jenkinsci/test/acceptance/po/WizardLogin.java
alvarolobato: Yes and no... this test will only run if the controller is `LocalController`, but potentially this could happen if we remove the restriction. Will change it.

src/main/java/org/jenkinsci/test/acceptance/junit/JenkinsAcceptanceTestRule.java
alvarolobato: that could also be a good approach, let me try...

alvarolobato: Committed a change with a new annotation @WithInstallWizard, now that looks a lot better. Thanks for the suggestion @olivergondza

src/main/java/org/jenkinsci/test/acceptance/junit/WithInstallWizard.java
```diff
@@ -0,0 +1,56 @@
...
+import org.junit.runners.model.Statement;
+
+import com.google.inject.Inject;
+
+/**
+ * Enables the install wizard to run the test. This is only possible for LocalControllers. Otherwise the test is
+ * skipped.
+ */
+@Retention(RUNTIME)
+@Target({ TYPE })
```
olivergondza: The rule seems to support `METHOD` as well, which is recommended.

alvarolobato:
Added method support.

batmat:
Going to release this. Please shout if you disagree. Thanks!

PR #281, opened 2017-03-06
Change check for number of branches.

varyvol:
Change the number of branches processed for the multibranch project.

@reviewbybees @olivergondza

### DIFF
--- a/src/test/java/plugins/WorkflowMultibranchTest.java
+++ b/src/test/java/plugins/WorkflowMultibranchTest.java
@@ -62,7 +62,7 @@ public class WorkflowMultibranchTest extends AbstractJUnitTest {
         assertThat(branchIndexingLog, containsString("Scheduled build for branch: jenkinsfile_failure"));
         assertThat(branchIndexingLog, containsString("Scheduled build for branch: jenkinsfile_success"));
         assertThat(branchIndexingLog, not(containsString("Scheduled build for branch: master")));
-        assertThat(branchIndexingLog, containsString("2 branches were processed"));
+        assertThat(branchIndexingLog, containsString("3 branches were processed"));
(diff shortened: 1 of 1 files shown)


olivergondza:
Where is the change coming from?

varyvol:
@olivergondza from this repo: https://github.com/varyvoltest/maven-basic

olivergondza:
My question is why it used to be 2 and it is 3 now. Was the test wrong? Has the plugin changed implementation?

The code change is so trivial that I expect it is the rational to be reviewed here. And it is not clear to me.

varyvol:
@olivergondza sorry Oliver, I didn't understand your question properly. Neither the test or the plugin have changed, only the repo used to "trigger" the functionality expected by the test.

The test creates a multibranch pipeline project with one repo and make some assertions. As the repo has changed (changes not related with this test), now one more branch is processed, but that is the only thing that affects this test.

All other assertions that were made before are still made and they continue to be valid, as this test guarantees.

olivergondza:
Oh, I see. How about basing the on https://github.com/jenkinsci/selenium-tests/branches, since it have to be on github, there is no reason to ever modify that repo and we do not have to rely on personal account and repo that clearly has a reason to change.

varyvol:
@olivergondza the branches need to follow some conditions, i.e: (not) having Jenkinsfile files. That's why I created a specific repo that could guarantee those conditions would be met.

This repo should not changed (the change was made during some testing but should not happen anymore). Anyway yes, I will create a non-personal repo and will change this test to use it.

But if you don't mind, we can merge this to avoid the test failing now and I will do such change in a future PR.

olivergondza:
Sure

PR #286, opened 2017-03-16
Update to latest parent jenkins:1.37

batmat:
@reviewbybees


mafraba:
@batmat @olivergondza not exactly sure what's happening, but this seems to break the build

PR #287, opened 2017-03-17
Revert "Update to latest parent jenkins:1.37"

batmat:
This failed the build, so reveals an issue. We need to work on understanding the failure this upgrades causes. I will refile a PR with that change and see if we can fix the failures.

Reverts jenkinsci/acceptance-test-harness#286

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci</groupId>
     <artifactId>jenkins</artifactId>
-    <version>1.37</version>
+    <version>1.36</version>
   </parent>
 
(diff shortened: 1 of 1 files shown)


mafraba:
🐝

batmat:
Went a bit expedite on this one because of the failure I introduced. This will help me work more serenely on introducing a non failing PR/update.

oleg-nenashev:
Likely this is because of the `java.level` property. If you define source and test levels, it may be still sneaking this property into the TestSource and TestTarget definitions indirectly (hence they may legitimately become Java6 after the update even if you define `source` and `target`). So it is kinda 🐜 for me. The proper solution is to define `java.level` explicitly.

PR #288, opened 2017-03-17
[JENKINS-42884] @WithPlugins find plugins in superclasses

varyvol:
[JENKINS-42884](https://issues.jenkins-ci.org/browse/JENKINS-42884)

@WithPlugins annotation to inspect superclasses for plugin specs.

@reviewbybees @olivergondza

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/junit/WithPlugins.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/junit/WithPlugins.java
@@ -6,6 +6,7 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
 import java.util.ArrayList;
 import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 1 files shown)


varyvol:
@olivergondza as previously superclasses were not inspecting, this should not break any test as those superclasses won't have the annotation. But if you can think of a harmful scenario, please let me know.

mafraba:
🐝  looks good

src/main/java/org/jenkinsci/test/acceptance/junit/WithPlugins.java
```diff
@@ -106,7 +113,7 @@ public void evaluate() throws Throwable {
                     base.evaluate();
                 }
 
-                private List<PluginSpec> combinePlugins(WithPlugins... wp) {
+                private List<PluginSpec> combinePlugins(List<WithPlugins> wp) {
```
jtnord: should this not be be a set to prevent the case where people use the annotation on both the parent and the current and have set it both the same to work around this issue.

varyvol: @jtnord if I am not wrong, that is already handled by the plugins installation logic. It would be the same as if someone put the same plugin twice in the annotation.

PR #291, opened 2017-03-24
[FIXED JENKINS-43086] add ssh verification strategy after SECURITY-161

varmenise:


src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_slaves/SshSlaveLauncher.java
```diff
@@ -17,14 +17,15 @@
...
-    public final Control jvmOptions = control("jvmOptions");
+    public final Control host = control("/host");
+    public final Control credentialsId = control("/credentialsId");
+    public final Control prefixCmd = control("/prefixStartSlaveCmd");
+    public final Control suffixCmd = control("/suffixStartSlaveCmd");
+    public final Control timeout = control("/launchTimeoutSeconds");
+    public final Control retries = control("/maxNumRetries");
+    public final Control javaPath = control("/javaPath");
+    public final Control jvmOptions = control("/jvmOptions");
+    public final Control sshHostKeyVerificationStrategy = control("/"); //weird path because of JENKINS-43076
```
amuniz: This is the only control added, why the diff chunk includes other lines?

varmenise: so otherwise I needed to use either  control("")  for the latest which 1) I think it's even more confusing 2) causes issues when resolving the path.

An alternative was to use control("/") only for the latest as it is now, but then it would have had a different format compared with the others. Plus other classes (e.g. DumSlave) use this approach.. So I can roll back, but I think it's clearer this way than having host, javaPath and then a "/"

src/test/java/plugins/SshSlavesPluginTest.java
```diff
@@ -160,12 +160,13 @@ private void verifyLog(String message) {
...
     private SshSlaveLauncher configureDefaultSSHSlaveLauncher() {
-        return configureSSHSlaveLauncher(sshd.ipBound(22), sshd.port(22));
+        return configureSSHSlaveLauncher(sshd.ipBound(22), sshd.port(22), "Non verifying Verification Strategy");
     }
     
-    private SshSlaveLauncher configureSSHSlaveLauncher(String host, int port) {
+    private SshSlaveLauncher configureSSHSlaveLauncher(String host, int port, String sshHostKeyVerificationStrategy) {
         SshSlaveLauncher launcher = slave.setLauncher(SshSlaveLauncher.class);
         launcher.host.set(host);
+        launcher.sshHostKeyVerificationStrategy.select(sshHostKeyVerificationStrategy);
```
jglick: Seems more straightforward than the idiom used in #300.

rsandell: I did a quick try, but couldn't make it work in the few minutes I poked at it.

varmenise:
@rsandell I think I can close this right? As you had fixed the same in your PR

varmenise:
As looks fixed by #300 (late due to my sabbatical time off which made me lose track of my stuff)

PR #297, opened 2017-04-04
Add an alternative mechanism for overriding plugins

stephenc:
Passing environment variables that contain a . in the name is problematic, much easier to pass the path of a `.properties` file and load the plugin versions or plugin files from that properties file, e.g.

```
echo "ldap=$(pwd)/ldap.hpi" > force.plugins
./run.sh firefox latest -Dtest=LdapPluginTest "-DFORCE_PLUGIN_FILES=$(pwd)/force.plugins"
```

or

```
echo "ldap=1.14" > force.plugins
./run.sh firefox latest -Dtest=LdapPluginTest "-DFORCE_PLUGIN_VERSIONS=$(pwd)/force.plugins"
```

(you can do a combination of both using different files for the versions and the local .hpi files)

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,7 +32,7 @@
     <maven.version>3.1.0</maven.version>
     <groovy.version>2.3.1</groovy.version>
     <shrinkwrap.version>1.2.2</shrinkwrap.version>
-    <monte.version>0.7.7</monte.version>
+    <monte.version>0.7.7.0</monte.version>
     <mockito.version>1.10.19</mockito.version>
     <cucumber.options />
     <trimStackTrace>false</trimStackTrace> <!-- surefire -->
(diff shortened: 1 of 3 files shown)


olivergondza:
It would be better to have a single property/envvar to identify properties its content will be injected as system properties. We can take advantage of this for all properties and the code that reads the properties does not have to change.

stephenc:
@olivergondza 

> It would be better to have a single property/envvar to identify properties its content will be injected as system properties. We can take advantage of this for all properties and the code that reads the properties does not have to change.

I am having problems parsing what your suggestion is. Could you try rephrasing

olivergondza:
Rereading what I wrote I am surprised Manuel managed to make any kind of sense out of it.
Let me retry: this adds 2 new properties to overcome the fact that plugin controlling properties are awkward to pass due to limitations in different shells. What I do not like is you are fixing the problem for "plugin injection" only. Not to mention this add confusion over what plugin version will have preference (the explicit one or the one from file) and it complicates the code.

-- 

What I suggest is to have one envvar/property, say `ATH_PROPERTIES`, that would point to a properties file to populate the ath with system properties that would be picked up bu their existing handling code. This will allow you to put all the configuration properties (not only those of plugin injection) into a file so you do not have to think about special chars in their names any longer or change the code that reads it (as in this PR).

jglick:
#273 solves this for the local file paths.

jglick:
I think #353 handles this more simply.

olivergondza:
Right, closing.

PR #298, opened 2017-04-05
[JENKINS-43371] Include Matcher for page URL

varyvol:
[JENKINS-43371](https://issues.jenkins-ci.org/browse/JENKINS-43371)

Include Matcher for page URL.

@reviewbybees @olivergondza

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/Matchers.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/Matchers.java
@@ -3,6 +3,7 @@ package org.jenkinsci.test.acceptance;
 import javax.annotation.CheckForNull;
 import java.io.IOException;
 import java.net.HttpURLConnection;
+import java.net.URL;
 import java.util.List;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/test/acceptance/Matchers.java
```diff
@@ -77,6 +78,20 @@ public void describeMismatchSafely(WebDriver item, Description d) {
         };
     }
 
+    public static Matcher<WebDriver> hasURL(final URL url) {
```
olivergondza: For the convenience of the caller it would be better to use String.

recena: @olivergondza an `URL` is better because includes validation.

varyvol: @olivergondza I used URL because the TopLevelItems url() method returns a URL, so it's quite straightforward to do something like waitFor(driver, hasURL(jenkins.url), 2).

olivergondza: Makes sense. Pity this arguments was not part of the description, though...

PR #300, opened 2017-04-13
Set the host key verification strategy to none

rsandell:
To fix the ssh slaves tests on the new ssh slaves plugin.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_slaves/SshSlaveLauncher.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_slaves/SshSlaveLauncher.java
@@ -10,6 +10,8 @@ import org.jenkinsci.test.acceptance.po.ComputerLauncher;
 import org.jenkinsci.test.acceptance.po.Control;
 import org.jenkinsci.test.acceptance.po.Describable;
 import org.jenkinsci.test.acceptance.po.PageObject;
+import org.openqa.selenium.By;
+
 import static org.junit.Assert.assertTrue;
 
 /**
@@ -25,6 +27,19 @@ public class SshSlaveLauncher extends ComputerLauncher {
     public final Control retries = control("maxNumRetries");
     public final Control javaPath = control("javaPath");
     public final Control jvmOptions = control("jvmOptions");
+    public final Control hostKeyVerificationStrategy = control("/");
+
+
+    //<select class="setting-input dropdownList" path="/launcher/"><option value="0">Known hosts file Verification Strategy</option><option value="1">Manually provided key Verification Strategy</option><option value="2">Manually trusted key Verification Strategy</option><option value="3">Non verifying Verification Strategy</option></select>
+
+    public enum HostKeyVerificationStrategy {
+        Known_hosts_file(0), Manually_provided(1), Manually_trusted(2), Non_verifying(3);
+        public final String value;
+
+        HostKeyVerificationStrategy(int value) {
+            this.value = String.valueOf(value);
+        }
+    }
 
     public SshSlaveLauncher(PageObject context, String path) {
         super(context, path);
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_slaves/SshSlaveLauncher.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_slaves/SshSlaveLauncher.java
@@ -6,10 +6,7 @@ import java.util.concurrent.TimeUnit;
 import org.jenkinsci.test.acceptance.plugins.credentials.UserPwdCredential;
 import org.jenkinsci.test.acceptance.plugins.ssh_credentials.SshCredentialDialog;
 import org.jenkinsci.test.acceptance.plugins.ssh_credentials.SshPrivateKeyCredential;
-import org.jenkinsci.test.acceptance.po.ComputerLauncher;
-import org.jenkinsci.test.acceptance.po.Control;
-import org.jenkinsci.test.acceptance.po.Describable;
-import org.jenkinsci.test.acceptance.po.PageObject;
+import org.jenkinsci.test.acceptance.po.*;
 import org.openqa.selenium.By;
 
 import static org.junit.Assert.assertTrue;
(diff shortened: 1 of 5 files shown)


src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_slaves/SshSlaveLauncher.java
```diff
@@ -25,6 +27,19 @@
     public final Control retries = control("maxNumRetries");
     public final Control javaPath = control("javaPath");
     public final Control jvmOptions = control("jvmOptions");
+    public final Control hostKeyVerificationStrategy = control("/");
+
+
+    //<select class="setting-input dropdownList" path="/launcher/"><option value="0">Known hosts file Verification Strategy</option><option value="1">Manually provided key Verification Strategy</option><option value="2">Manually trusted key Verification Strategy</option><option value="3">Non verifying Verification Strategy</option></select>
```
recena:
```
// <select class="setting-input dropdownList" path="/launcher/">
//   <option value="0">Known hosts file Verification Strategy</option>
//   <option value="1">Manually provided key Verification Strategy</option>
//   <option value="2">Manually trusted key Verification Strategy</option>
//   <option value="3">Non verifying Verification Strategy</option>
// </select>

Easier to read.
```

rsandell: I mainly added it as a means of copy pasting the values, but kept it around for future reference.

recena: @rsandell It is fine. Now you can copy and paste from my comment 😄

recena: @rsandell Do you plan to address my suggestion?

src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_slaves/SshSlaveLauncher.java
```diff
@@ -25,6 +27,19 @@
...
+
+    public enum HostKeyVerificationStrategy {
+        Known_hosts_file(0), Manually_provided(1), Manually_trusted(2), Non_verifying(3);
```
jglick: This will be a maintenance burden if we ever add or reorder those strategies. Consider using the ATH feature of associating selections with a `Descriptor.id` or `.displayName`.

src/main/java/org/jenkinsci/test/acceptance/po/DumbSlave.java
```diff
@@ -48,9 +50,14 @@ public void asLocal() {
         File jar = new File(tmpDir, "slave"+createRandomName()+".jar");
         String command;
         if (SystemUtils.IS_OS_UNIX) {
+            String java = "java";
+            String jh = System.getProperty("java.home");
+            if (!StringUtils.isEmpty(jh)) {
+                java = new File(jh, "bin/java").getAbsolutePath();
```
jglick: Safer to check `File.executable`.

src/test/java/plugins/WorkflowPluginTest.java
```diff
@@ -135,6 +137,8 @@
         build.shouldContainsConsoleOutput("Building version 1.0-SNAPSHOT");
         jenkins.restart();
         // Default 120s timeout of Build.waitUntilFinished sometimes expires waiting for RetentionStrategy.Always to tick (after initial failure of CommandLauncher.launch: EOFException: unexpected stream termination):
+        System.out.println("Waiting for slave to come online...");
+        Thread.sleep(10000); //Jenkins needs some extra time to wake up before we can query the api
```
jglick: Not sure what this refers to. Is this purporting the address the comment on line 139?

src/test/java/plugins/WorkflowPluginTest.java
```diff
@@ -243,7 +248,7 @@
             "}");
         job.sandbox.check();
         job.save();
-        assertThat(job.startBuild().shouldSucceed().getConsole(), containsString("-> FETCH_HEAD"));
+        assertThat(job.startBuild().shouldSucceed().getConsole(), containsString("-&gt; FETCH_HEAD"));
```
jglick: @Vlatombe it seems that #278 was an incompatible change? This change should not be necessary.

src/test/java/plugins/WorkflowPluginTest.java
rsandell: Maybe, I had issues on my machine that the json api call below failed with connection refused, and I did various tests for some better waitFor but there wasn't much available to wait for, so this solved it in a simple way, yet ugly.

src/main/java/org/jenkinsci/test/acceptance/po/DumbSlave.java
rsandell: Huh? I needed the full path to java because my PATH env var was not available when I run from within the IDE, it still worked when running mvn from commandline, but this was needed to debug easier.

jglick: I was just pointing out that it would be safer to write

```java
File javaBin = new File(jh, "bin/java");
if (javaBin.isExecutable()) {
    java = javaBin.getAbsolutePath();
}
```

src/test/java/plugins/WorkflowPluginTest.java
jglick: Sorry, which JSON API call? If `waitUntilOnline` is not working as advertised then it needs to be fixed.

rsandell: `waitUntilOnline` checks `slave.isOnline` which in essence is `!getJson().get("offline").asBoolean()`

It works, but only when Jenkins is fully online apparently. It seems like `jenkins.restart()` isn't waiting long enough or something. I tried various tweaks but I just ended up tumbling further and further down the rabbit hole, so I decided to dig myself up before I spent too much time on something I wasn't supposed to spend time on.

rsandell:
I see now this has already been implemended once before in #291 by @varmenise but it's not merged yet.
Conflicting tasks?

rsandell:
@jglick unrelated changes reverted.

src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_slaves/SshSlaveLauncher.java
```diff
@@ -100,4 +100,50 @@ private void waitForCredentialVisible(final String credUsername) {
...
+            SshHostKeyVerificationStrategy strategy = type.newInstance();
+            hostKeyVerificationStrategy.select(strategy.id());
+        } catch (InstantiationException | IllegalAccessException e) {
+            throw new IllegalArgumentException(type.getName() + " is missing a default constructor.", e);
+        }
+
+    }
+
+    public abstract static class SshHostKeyVerificationStrategy {
+        public abstract String id();
```
jglick: Is this really necessary? Normally `Describable`s in a list are selected by `Descriptor.displayName` or `.id`. Relying on a numeric position is pretty fragile. Perhaps you overlooked the typical methods for picking an item by label?

rsandell: Picking item by label didn't work, something is off with the UI control that makes it not behave like all the other descibable selections.

jglick: Did you check whether @varmenise’s much simpler version works?

rsandell:
@reviewbybees done

src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_slaves/SshSlaveLauncher.java
rsandell: I applied the same approach on top of mine, but I might have done something wrong since it didn't work.

PR #301, opened 2017-04-13
Skip tests explicitly requesting plugins not in update center

olivergondza:
When ATH is run against downstream update center with plugins missing, the
tests will be correctly skipped. This also permits upstream tests for
plugins that are not distributed in upstream UC while they are in the
downstream one.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/update_center/UpdateCenterMetadata.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/update_center/UpdateCenterMetadata.java
@@ -17,6 +17,7 @@ import java.util.List;
 import java.util.Map;
 
 import org.jenkinsci.test.acceptance.po.Jenkins;
+import org.junit.internal.AssumptionViolatedException;
 
 /**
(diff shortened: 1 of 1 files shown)


scoheb:
LGTM

src/main/java/org/jenkinsci/test/acceptance/update_center/UpdateCenterMetadata.java
```diff
@@ -78,7 +79,11 @@ private void init() {
         List<PluginMetadata> set = new ArrayList<>();
         for (PluginSpec n : plugins) {
             PluginMetadata p = this.plugins.get(n.getName());
-            if (p==null) throw new IllegalArgumentException("No such plugin " + n.getName());
+            if (p==null) {
```
recena: `p == null` always means `Update Center is not offering the plugin`?

olivergondza: If the collection contained `null` value the `#init()` would blow up so I am quite confident about it.

PR #302, opened 2017-04-17
[JENKINS-34522] Adapt to change in case

daniel-beck:
ATH regression probably due to https://github.com/jenkinsci/jenkins/pull/2772 in 2.50.

Untested.

Probably safer to go with `Schedule a` instead?


jglick:
Yeah let us rather go with `Schedule a`.

uhafner:
You should also spend some more time in core to add IDs to your UI elements. It will make ATH testing much simpler...

daniel-beck:
@jglick Done.

PR #305, opened 2017-05-02
[JENKINS-43388] Update ATH for https://github.com/jenkinsci/ldap-plugin/pull/18

stephenc:
See [JENKINS-43388](https://issues.jenkins-ci.org/browse/JENKINS-43388)

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/LdapSecurityRealm.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/LdapSecurityRealm.java
@@ -1,8 +1,18 @@
 package org.jenkinsci.test.acceptance.po;
 
+import com.google.common.base.Function;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
+import org.hamcrest.BaseMatcher;
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
 import org.jenkinsci.test.acceptance.plugins.ldap.LdapDetails;
 import org.jenkinsci.test.acceptance.plugins.ldap.LdapEnvironmentVariable;
 import org.jenkinsci.test.acceptance.plugins.ldap.LdapGroupMembershipStrategy;
+import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
 
 /**
@@ -40,6 +50,10 @@ public class LdapSecurityRealm<T extends LdapGroupMembershipStrategy> extends Se
      * since version 1.8
      */
     private final Control mailAddressAttributeName = control("mailAddressAttributeName");
+    /**
+     * since version 1.15
+     */
+    private final Control validateButton = control("validateLdapSettings");
 
     public LdapSecurityRealm(GlobalSecurityConfig context, String path) {
         super(context, path);
@@ -96,6 +110,36 @@ public class LdapSecurityRealm<T extends LdapGroupMembershipStrategy> extends Se
         }
     }
 
+    public void validate(String username, String password) {
+        final String id = validateButton.resolve().getAttribute("id");
+        final By by = By.id(id+"_dialog");
+        for (WebElement element : all(by)) {
+            if (element.isDisplayed()) {
+                throw new AssertionError(String.format("Form at path %s is already visible", element.getAttribute("path")));
+            }
+        }
+        validateButton.click();
+        String formPath = waitFor().withTimeout(10, TimeUnit.SECONDS).until(new Function<CapybaraPortingLayer, String>() {
+            @Nullable
+            @Override
+            public String apply(@Nullable CapybaraPortingLayer input) {
+                for (WebElement element: all(by)) {
+                    if (element.isDisplayed()) {
+                        return id+"_dialog";
+                    }
+                }
+                return null;
+            }
+
+            @Override
+            public String toString() {
+                return "Form area to appear: " + by;
+            }
+        });
+        new Validation(this.getPage(), formPath).submit(username, password);
+        ldapServer.click();
+    }
+
     /**
      * Subclasses can override this to handle group membership differently.
      */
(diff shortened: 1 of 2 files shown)


src/test/java/plugins/LdapPluginTest.java
```diff
@@ -32,7 +32,7 @@
  */
 @Category(DockerTest.class)
 @WithDocker
-@WithPlugins("ldap@1.10")
+@WithPlugins("ldap@1.15")
```
olivergondza: This effectively removes ATH coverage from LDAP plugin for all older version than the latest - released an hour ago. Only the test methods that really need it should require it.

src/main/java/org/jenkinsci/test/acceptance/po/LdapSecurityRealm.java
```diff
@@ -96,6 +110,36 @@ public void configure(LdapDetails<T> ldapDetails) {
...
+                return null;
+            }
+
+            @Override
+            public String toString() {
+                return "Form area to appear: " + by;
+            }
+        });
+        new Validation(this.getPage(), formPath).submit(username, password);
+        ldapServer.click();
```
rsandell: A bit strange that you need to click there.

src/main/java/org/jenkinsci/test/acceptance/po/LdapSecurityRealm.java
```diff
@@ -106,4 +150,140 @@ protected void configureGroupMembership(LdapDetails<T> ldapDetails) {
...
+                                for (WebElement element : realm.all(by)) {
+                                    if (element.findElements(By.tagName("div")).isEmpty()) {
+                                        continue;
+                                    }
+                                    if (element.findElements(By.className("error")).isEmpty()
+                                            && element.findElements(By.className("warning")).isEmpty()) {
+                                        return Boolean.TRUE;
+                                    } else {
+                                        return Boolean.FALSE;
+                                    }
```
jglick: Could all be simplified to

```java
return element.findElements(By.className("error")).isEmpty()
    && element.findElements(By.className("warning")).isEmpty();
```

src/main/java/org/jenkinsci/test/acceptance/po/LdapSecurityRealm.java
```diff
@@ -106,4 +150,140 @@ protected void configureGroupMembership(LdapDetails<T> ldapDetails) {
...
+                            public Boolean apply(@Nullable CapybaraPortingLayer input) {
+                                for (WebElement element : realm.all(by)) {
+                                    if (element.findElements(By.tagName("div")).isEmpty()) {
+                                        continue;
+                                    }
+                                    if (!element.findElements(By.className("warning")).isEmpty()) {
+                                        return Boolean.TRUE;
+                                    } else {
+                                        return Boolean.FALSE;
+                                    }
```
jglick:
```java
return !element.findElements(By.className("warning")).isEmpty();
```

src/main/java/org/jenkinsci/test/acceptance/po/LdapSecurityRealm.java
```diff
@@ -106,4 +150,140 @@ protected void configureGroupMembership(LdapDetails<T> ldapDetails) {
...
+                            public Boolean apply(@Nullable CapybaraPortingLayer input) {
+                                for (WebElement element : realm.all(by)) {
+                                    if (element.findElements(By.tagName("div")).isEmpty()) {
+                                        continue;
+                                    }
+                                    if (!element.findElements(By.className("error")).isEmpty()) {
+                                        return Boolean.TRUE;
+                                    } else {
+                                        return Boolean.FALSE;
+                                    }
```
jglick:
```java
return !element.findElements(By.className("error")).isEmpty();
```

src/test/java/plugins/LdapPluginTest.java
```diff
@@ -70,6 +70,34 @@ private LdapDetails createDefaultsWithoutManagerCred(LdapContainer ldapContainer
     }
 
     @Test
+    @WithPlugins("ldap@1.15")
```
jglick: I think you need to fix [this](https://github.com/jenkinsci/acceptance-test-harness/blob/bb8772bccbd8f882761df91d102be2b7fe343583/src/main/java/org/jenkinsci/test/acceptance/junit/WithPlugins.java#L121) first, or simply expect 1.15 for the whole test suite.

(ATH in its normal run mode will always download the latest version of the plugin, so a version annotation is only useful then for documentation or blocking a test from running until a snapshot is released. There are less commonly used modes whereby you can test against a live server or a WAR bundling plugins, in which case a newer version annotation can force a test to be skipped or an update applied IIRC. But in practice we do not much care about testing historical plugin versions, any more than we expect historical revisions of ATH to continue to pass today.)

olivergondza: Well, I do. Red Hat maintains an internal downstream update centers with verified plugins and this is an essential feature for testing it. I presume CB have such internal UCs as well. This will also kick in when not the latest LTS is tested and it reached one of the upstream update center where the latest plugins might be missing as they require newer Jenkins version.

Good point on this still does not work.

olivergondza: Fix proposed in https://github.com/jenkinsci/acceptance-test-harness/pull/306.

jglick: > This will also kick in when not the latest LTS is tested

Though in general tests do not pass very consistently on old LTS lines I think.

jglick: > a newer version annotation can force a test to be skipped or an update applied IIRC

FTR: seems to be controlled by `$NEVER_REPLACE_EXISTING_PLUGINS`.

olivergondza:
The validations can be simplified with https://github.com/jenkinsci/acceptance-test-harness/pull/324

stephenc:
I've given up on this PR

PR #309, opened 2017-05-11
LogParser Plugin Test

sikelerd:
- Adding PageObjects for configuring the logparser plugin.
- First simple test of the logparser plugin, further testcase will be added soon.


src/main/java/org/jenkinsci/test/acceptance/plugins/logparser/LogParserPublisher.java
```diff
@@ -0,0 +1,93 @@
+package org.jenkinsci.test.acceptance.plugins.logparser;
+
+import org.jenkinsci.test.acceptance.po.*;
+
+/**
+ *
```
uhafner: description

src/main/java/org/jenkinsci/test/acceptance/plugins/logparser/LogParserPublisher.java
```diff
@@ -0,0 +1,93 @@
...
+     * Constructor.
+     * Should only be instantiated by addPublisher(...) of a job
+     *
+     * @param parent
+     * @param path
+     */
+    public LogParserPublisher(Job parent, String path) {
+        super(parent, path);
+
+        controlMarkOnUnstableWarning = control("unstableOnWarning");
```
uhafner: initialize controls directly in the field declaration.

src/test/java/plugins/LogParserTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+@WithPlugins("log-parser")
+public class LogParserTest extends AbstractJUnitTest {
+
+    // GlobalConfig for the LogParser-Plugin
+    private LogParserGlobalConfig config;
+    // Available Rules for the tests
+    private Map<String, String> rules;
+
+    @Before
```
uhafner: If the information is constant it makes sense to rather use @BeforeClass

src/test/java/plugins/LogParserTest.java
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * Check whether trend is visible
+     */
+    @Test
+    public void trendVisible(){
+        FreeStyleJob job = jenkins.jobs.create(FreeStyleJob.class, "simple-job");
+
+        job.configure();
+        // sample use of the LogParserPublisher
+        LogParserPublisher lpp = job.addPublisher(LogParserPublisher.class);
```
uhafner: Or: job.configure(() -> {lambda})

src/test/java/plugins/LogParserTest.java
engelmi: Within the @Before function multiple non-static references, like jenkins, are accessed. When is a new instance of jenkins created? On each start of a test case or test class?

uhafner: Jenkins is started for each test, so you get a clean version of Jenkins for each test. 

If you change the state of Jenkins within this method, then @Before is required indeed. I missed that part...

src/main/java/org/jenkinsci/test/acceptance/plugins/logparser/LogParserGlobalConfig.java
```diff
@@ -0,0 +1,54 @@
...
+public class LogParserGlobalConfig extends PageAreaImpl {
+
+    /**
+     * Repeatable-add Button.
+     */
+    private final Control addButton;
+    
+    public LogParserGlobalConfig(JenkinsConfig context) {
+        super(context, "/hudson-plugins-logparser-LogParserPublisher/log-parser");
+        addButton = control("repeatable-add");
```
olivergondza: This can be moved to initializer as control is resolved lazily.

src/main/java/org/jenkinsci/test/acceptance/plugins/logparser/LogParserOutputPage.java
```diff
@@ -0,0 +1,92 @@
...
+
+    // URL to return from the opened frame to the default window
+    private String restoreURL = null;
+
+    /**
+     * Opens the defined frame in the current window.
+     *
+     * @param frame Enum of the frame to open in the current window.
+     */
+    public void openFrameInWindow(LogParserFrame frame) {
```
olivergondza: `openFrameInWindow`,  `restoreWindow`, `switchToFrame`, `switchToDefaultContent`, `switchToMainframe` are unused. Do you need that? Provided you do, I would like the page object to be redesigned so the fact the page is build from frames is encapsulated in that page object rather then exposed.

engelmi: Unfortunately, there is no reference work in the project concerning frames yet (to the best of my knowledge).
The logparser plugin uses frames/iframes to display the parsed console output. Consequently, we need to access those to write further (yet to come) tests. As a frame is a context within a page, we assumed an exposure of such context-changing methods would be appropriate. If I understand you correctly, there should be (in our case) one class for each frame type?

olivergondza: What I do not like about this approach (or the design of `WebDriver` in fact) is changing the global context. Everybody who use this API needs to keep in mind to put the context back (what does it mean *back*, anyway? ` restoreWindow`, `switchToDefaultContent` or `switchToMainframe`?) otherwise some latter calls might fail with an error that makes no sense (as it will interact with something else then it should be). This includes failed assertions so the test to be written correctly will be full of `finally` blocks to put the context back.

I suggested instead the API should expose a series of command/query methods that would change the context to do its task / get the information and then put the context back in `finally` so users do not have to care it uses frame at all.

src/main/java/org/jenkinsci/test/acceptance/plugins/logparser/LogParserOutputPage.java
```diff
@@ -0,0 +1,84 @@
...
+    public void executeInFrame(LogParserFrame targetFrame, Runnable action) {
+        try{
+            switchToFrame(targetFrame);
+            action.run();
+        }
+        finally{
+            switchToDefaultContent();
+        }
+    }
+}
```
engelmi: @olivergondza Please have a look at the implemented changes.

olivergondza: This is better in a way the clients are hard to implement wrong. Though, I still prefer not to expose the fact that `LogParserOutputPage` is using frames on the first place. What I had in mind is the class would have actual "business" methods like `getFailureCount` (that would extract some data from content frame) or `rerunAnalysis` (that would click some button in the sidebar frame) that would do the switching inside uisng the `try`/`finally` as above so users do not have to care. And the API do not have to be touched when plugin developer decides to replace frames with something else.

This way, you actually expect that whoever provides the `Runnable` knows the details of the page markup (ids, html structure, etc.) which is something page objects are here to encapsulate.

engelmi: Ah, I understand. If page objects are used to encapsulate details of the page markup, then this approach is indeed not best practice. However, I would like to use the `Runnable ` as we don't know at this point which "business" methods will be needed later on and remain flexible. Therefore my question: 
Should we implement a few "business" methods in the `LogParserOutputPage `and add functionality when the need arises or should we change the extend-class (although I don't know which would be suitable) and keep the `Runnable` approach?

olivergondza: Feel free to implement only the business methods you need for the tests here and add more methods (expose more actions/data) when (and if) you need it. There should be no problem with is since this is in single repo.

PR #310, opened 2017-05-12
[JENKINS-44232] Fallback when exception is thrown in CapybaraPortingLayerImpl.check

armfergom:
See [JENKINS-44232](https://issues.jenkins-ci.org/browse/JENKINS-44232)

If for some reason, performing a click on a checkbox throws an exception, the Javascript fallback is never tried. We should capture the exception, log it, and try the Javascript fallback. More details in the ticket attached.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
@@ -7,10 +7,13 @@ import java.net.URL;
 import java.util.List;
 import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 import org.hamcrest.StringDescription;
 import org.jenkinsci.test.acceptance.junit.Resource;
 import org.jenkinsci.test.acceptance.junit.Wait;
+import org.jenkinsci.test.acceptance.junit.WithPlugins;
 import org.jenkinsci.test.acceptance.utils.ElasticTime;
 import org.openqa.selenium.Alert;
 import org.openqa.selenium.By;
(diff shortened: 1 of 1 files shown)


mafraba:
🐝

olivergondza:
What exceptions are these? Why do they occur? Is the radio the only element affected?

Looking at the stacktrace in JIRA, the problem is the page was not scrolled properly or rendered properly at the time of the click. The reason why JavaScript approach would prevent the exception being thrown is it does not care whether the element is hidden by some other or not. Selenium is checking for that on purpose - to ensure the elements you interact with are visible. We either want that and we should the failure properly or turn the visibility verification off if we do not want it. Fixing for one kind of interaction with one kind of control

armfergom:
@olivergondza they occur with checkboxes that are out of sight **horizontally**. It happens with small resolutions. It does try to scroll horizontally but it doesn't do it right, causing the exception. The Javascript fallback is here for a similar reason. Back in the day, when it was implemented there was no exception, but there was also an horizontal scrolling problem. I have basically done the same we did, overcome the horizontal scrolling problem with the same fallback.

If we were to disable the visibility verification, would that be with a global scope, or can that be done for a single test or test suite? I don't think we should disable it globally anyway.

armfergom:
Closing as this is superseded by #311

olivergondza:
> If we were to disable the visibility verification, would that be with a global scope, or can that be done for a single test or test suite? I don't think we should disable it globally anyway.

Well, in theory, we should keep it always on to be on the safe side. I am willing to admit that this is causing way too much of a hassle for what amount of safety it provides (I do not recall this has ever coughed a bug) and turn it off globally.

Turning it off selectively - per control or per test suite - seems to combine the negatives of both. We can no longer rely that all controls that are interacted with are visible and we still will run onto those exceptions in edge cases, albeit less often.

PR #315, opened 2017-05-26
make pageLoadTimeout configurable

sikelerd:
On slow hardware it isn't possible to run any test because the
pageLoadTimeout is exceeded.

### DIFF
--- a/docs/BROWSER.md
+++ b/docs/BROWSER.md
@@ -42,6 +42,8 @@ For example,
 
 See [WIRING.md](WIRING.md) for details of where to put this.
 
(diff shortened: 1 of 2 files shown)


olivergondza:
Use `ElasticTime.factor`[1] to slow down all the timeout proportionally.

https://github.com/jenkinsci/acceptance-test-harness/blob/3b0a9f6ba37f6453f60273f14e101f6a632a17c6/src/main/java/org/jenkinsci/test/acceptance/utils/ElasticTime.java#L47

uhafner:
Would be nice if we can have this referenced somewhere in the docs...

PR #318, opened 2017-05-30
Ignore failing test

jglick:
@reviewbybees esp. @varyvol

### DIFF
--- a/src/test/java/plugins/WorkflowMultibranchTest.java
+++ b/src/test/java/plugins/WorkflowMultibranchTest.java
@@ -10,6 +10,7 @@ import org.jenkinsci.test.acceptance.po.WorkflowJob;
 import org.jenkinsci.test.acceptance.po.WorkflowMultiBranchJob;
 import org.junit.Assert;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Test;
 
 import javax.mail.MessagingException;
@@ -25,7 +26,7 @@ import static org.junit.Assert.assertEquals;
 /**
  * Tests a multibranch pipeline flow
  */
-@WithPlugins({"git", "javadoc@1.4", "workflow-job", "workflow-cps", "workflow-basic-steps", "workflow-durable-task-step", "workflow-multibranch", "github-branch-source"})
+@WithPlugins({"git", "javadoc@1.4", "workflow-basic-steps", "workflow-durable-task-step", "workflow-multibranch", "github-branch-source"})
 public class WorkflowMultibranchTest extends AbstractJUnitTest {
 
     @Before
@@ -33,6 +34,7 @@ public class WorkflowMultibranchTest extends AbstractJUnitTest {
         MavenInstallation.installMaven(jenkins, "M3", "3.3.9");
     }
 
+    @Ignore("cannot run quickly as anonymous")
     @Test
     public void testMultibranchPipeline() throws IOException, MessagingException {
         final WorkflowMultiBranchJob multibranchJob = jenkins.jobs.create(WorkflowMultiBranchJob.class);


src/test/java/plugins/WorkflowMultibranchTest.java
```diff
@@ -1,40 +1,51 @@
...
-import static org.hamcrest.Matchers.anyOf;
-import static org.hamcrest.Matchers.not;
-import static org.jenkinsci.test.acceptance.Matchers.hasAction;
-import static org.junit.Assert.assertEquals;
-
 /**
  * Tests a multibranch pipeline flow
  */
-@WithPlugins({"git", "javadoc@1.4", "workflow-job", "workflow-cps", "workflow-basic-steps", "workflow-durable-task-step", "workflow-multibranch", "github-branch-source"})
+@WithPlugins({"git", "javadoc@1.4", "workflow-basic-steps", "workflow-durable-task-step", "workflow-multibranch", "github-branch-source"})
```
jglick: These were dependencies of `workflow-multibranch` so did not need to be called out. And it was broken to request them: `CachedUpdateCenterMetadataLoader` fails to pass `?version=…` to the UC, so it gets back `workflow-job` 2.12, which requires Jenkins core 2.60+. I thought @olivergondza claimed this was not a problem, but it was for me.

src/test/java/plugins/WorkflowMultibranchTest.java
```diff
@@ -1,40 +1,51 @@
...
     }
 
     @Test
     public void testMultibranchPipeline() throws IOException, MessagingException {
+        final CredentialsPage mc = new CredentialsPage(jenkins, ManagedCredentials.DEFAULT_DOMAIN);
+        mc.open();
+        final UserPwdCredential cred = mc.add(UserPwdCredential.class);
+        cred.username.set("varyvoltest");
+        cred.password.set("ff6f95500a17b42872ef1d525e94cc1066aec488");
+        cred.setId(GH_CRED_ID);
```
jglick: Lifted from `SharedLibraryGithubTest`. `MultibranchPipelineTest` was similar, too; I am not sure how these two tests are supposed to differ: seems like there is a lot of overlap.

recena: Why are you using credential here when the GitHub account is public and has public repos?

jglick: That is exactly why the test failed: branch indexing could take far longer than 20s when not using an access token, because the GitHub API rate limit for anonymous access is very low, and the plugin deliberately sleeps if it is not assured of plenty of remaining calls.

recena: I'm not sure whether the root of the problem was that.

src/test/java/plugins/WorkflowMultibranchTest.java
```diff
@@ -1,40 +1,51 @@
...
         MavenInstallation.installMaven(jenkins, "M3", "3.3.9");
     }
 
     @Test
     public void testMultibranchPipeline() throws IOException, MessagingException {
+        final CredentialsPage mc = new CredentialsPage(jenkins, ManagedCredentials.DEFAULT_DOMAIN);
+        mc.open();
+        final UserPwdCredential cred = mc.add(UserPwdCredential.class);
+        cred.username.set("varyvoltest");
+        cred.password.set("ff6f95500a17b42872ef1d525e94cc1066aec488");
```
recena: How this token was generated?

src/test/java/plugins/WorkflowMultibranchTest.java
jglick: Well I reproduced that problem when running the test locally; made the change shown here; ran the test again and it passed. So if you know of some other problem, let me know or offer a PR for it.

src/test/java/plugins/WorkflowMultibranchTest.java
jglick: You will have to ask @varyvol, who wrote this test and the others I refer to.

recena: @jglick I assume this credentials are being used in other tests, and you are using them here.

src/test/java/plugins/WorkflowMultibranchTest.java
recena: I said that because the test is failing constantly. But you run locally, it is fine!

src/test/java/plugins/WorkflowMultibranchTest.java
jglick: Yes, see my comment above.

src/test/java/plugins/WorkflowMultibranchTest.java
varyvol: @jglick be careful, you are confusing projects.

varyvol:
@jglick looks good and it should fix the test failure, but please in the future ask before making public a private key.

src/test/java/plugins/WorkflowMultibranchTest.java
olivergondza: No sure I fully understand. I recall the problem with workflow-job 2.12 being served but I thought it is infra problem only. You are right `CachedUpdateCenterMetadataLoader` should be fixed to announce the version.

jglick: Filed #326 for this.

varyvol:
🐝

jglick:
@reviewbybees done

PR #320, opened 2017-06-05
Mission control view plugin: test basis and two test cases

engelmi:
Implemented a basis for tests of the mission control plugin as well as two test cases.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/mission_control/MissionControlView.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/mission_control/MissionControlView.java
@@ -14,8 +14,6 @@ import java.util.List;
 
 /**
  * {@link View} class of the mission control plugin.
- *
- * @author engelmi
  */
 @Describable("Mission Control")
 public class MissionControlView extends View {
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/mission_control/MissionControlView.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/mission_control/MissionControlView.java
@@ -1,0 +1,285 @@
+package org.jenkinsci.test.acceptance.plugins.mission_control;
+
+import com.google.inject.Injector;
+import org.jenkinsci.test.acceptance.po.ConfigurablePageObject;
+import org.jenkinsci.test.acceptance.po.Control;
+import org.jenkinsci.test.acceptance.po.Describable;
+import org.jenkinsci.test.acceptance.po.View;
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebElement;
+
+import java.net.URL;
+import java.util.LinkedList;
+import java.util.List;
+
+/**
+ * {@link View} class of the mission control plugin.
+ *
+ * @author engelmi
+ */
+@Describable("Mission Control")
+public class MissionControlView extends View {
+
+    // Every element of the mission control view to configure it properly...
+    private Control viewName = new Control(this, "/name");
+    private Control viewDescription = new Control(this, "/description");
+    private Control filterQueue = new Control(this, "/filterQueue");
+    private Control filterExecutors = new Control(this, "/filterExecutors");
+    private Control filterJobBuildHistory = new Control(this, "/useRegexFilterBuildHistory");
+    private Control filterJobBuildStatuses = new Control(this, "/useRegexFilterJobStatuses");
+    private Control fontSize = new Control(this, "/fontSize");
+    private Control historySize = new Control(this, "/buildHistorySize");
+    private Control queueSize = new Control(this, "/buildQueueSize");
+    private Control condensedTables = new Control(this, "/useCondensedTables");
+    private Control filterByFailures = new Control(this, "/filterByFailures");
+    private Control hideBuildHistory = new Control(this, "/hideBuildHistory");
+    private Control hideJobs = new Control(this, "/hideJobs");
+    private Control hideBuildQueue = new Control(this, "/hideBuildQueue");
+    private Control hideNodes = new Control(this, "/hideNodes");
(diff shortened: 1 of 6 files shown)


src/main/java/org/jenkinsci/test/acceptance/plugins/mission_control/MissionControlView.java
```diff
@@ -0,0 +1,285 @@
...
+import org.openqa.selenium.WebElement;
+
+import java.net.URL;
+import java.util.LinkedList;
+import java.util.List;
+
+/**
+ * {@link View} class of the mission control plugin.
+ *
+ * @author engelmi
```
uhafner: If you want the credits put your real name here ;-)

src/main/java/org/jenkinsci/test/acceptance/plugins/mission_control/MissionControlView.java
```diff
@@ -0,0 +1,285 @@
...
+     */
+    public void setHeightRatio(String ratioOption) {
+        ensureConfigPage();
+        heightRatio.set(ratioOption);
+    }
+
+    /**
+     * Reloads the configuration of jenkins from disk.
+     * Necessary to display the build history if a new {@link MissionControlView} is created.
+     * (Inefficient) alternative: {@code jenkins.restart()}
```
uhafner: Is this a bug of the plug-in? Or why is this required?

src/test/java/plugins/MissionControlTest.java
```diff
@@ -0,0 +1,106 @@
...
+        // add shell build step to cause the job to fail
+        job.addShellStep("sh <");
+        job.save();
+        job.startBuild().waitUntilFinished();
+
+        // open mission control view and assert that no build entries are being displayed
+        // (after the creation of a new mission control view the configuration needs to be reloaded)
+        view.open();
+        WebElement buildHistory = driver.findElement(By.id("jenkinsBuildHistory"));
+        Assert.assertEquals(0, buildHistory.findElements(By.xpath(".//tbody/tr")).size(), 0);
```
uhafner: Use Harmcrest matcher
assertThat(buildHistory.findElements(By.xpath(".//tbody/tr")), hasSize(0));

src/test/java/plugins/MissionControlTest.java
```diff
@@ -0,0 +1,106 @@
...
+        job.configure();
+        // add shell build step to cause the job to fail
+        job.addShellStep("sh <");
+        job.save();
+        job.startBuild().waitUntilFinished();
+
+        // open mission control view and assert that no build entries are being displayed
+        // (after the creation of a new mission control view the configuration needs to be reloaded)
+        view.open();
+        WebElement buildHistory = driver.findElement(By.id("jenkinsBuildHistory"));
```
uhafner: Couldn't this be a method of the view so we can call `view.getHistory()`.

src/test/java/plugins/MissionControlTest.java
```diff
@@ -0,0 +1,106 @@
...
+        Assert.assertEquals(0, buildHistory.findElements(By.xpath(".//tbody/tr")).size(), 0);
+
+        // reload configuration (alternative: jenkins.restart(), very inefficient for this task)
+        view.reloadConfiguration();
+
+        // open mission control view again and assert that the n-builds are being displayed
+        view.open();
+        buildHistory = driver.findElement(By.id("jenkinsBuildHistory"));
+        Assert.assertEquals(historySize, buildHistory.findElements(By.xpath(".//tbody/tr")).size(), 0);
```
uhafner: Harmcrest

src/test/java/plugins/MissionControlTest.java
```diff
@@ -0,0 +1,106 @@
...
+        // check for correct highlighting of the builds
+        WebElement failedBuild = buildHistory.findElement(By.xpath(".//tbody/tr[td='" + strFailedJob + "']"));
```
uhafner: If you plan to write more history tests then this code should be placed into a PageArea sub-class.

src/test/java/plugins/MissionControlTest.java
```diff
@@ -0,0 +1,106 @@
...
+        jobFailed.configure();
+        // add shell build step to cause the job to fail
+        jobFailed.addShellStep("sh <");
+        jobFailed.save();
+        jobFailed.startBuild().waitUntilFinished();
+
+        view.open();
+        // check for the correct highlighting of the jobs
+        WebElement jobContainer = driver.findElement(By.id("jenkinsJobStatuses"));
+        WebElement btnNotBuild = jobContainer.findElement(By.xpath("//button[text()='" + strJobNotBuild + "']"));
```
uhafner: Extract into method with parameter `name`

src/main/java/org/jenkinsci/test/acceptance/plugins/mission_control/MissionControlView.java
Larcagur: Default at my pc, didn't heed it actually ;)

src/main/java/org/jenkinsci/test/acceptance/plugins/mission_control/MissionControlView.java
Larcagur: It is required (see: [https://wiki.jenkins-ci.org/display/JENKINS/Mission+Control+Plugin](url) at "Setting Up"). I don't know how mission control works internally to display the build history, but it seems the plugin developers are aware of this behavior.

src/test/java/plugins/MissionControlTest.java
Larcagur: Good point. I don't know how many tests of the build history will be written, but a PageArea seems to be a better approach. Should the View provide a method to create the PageArea object (as it is only valid if the view exists) or should the PageArea be decoupled from the View?

PR #321, opened 2017-06-06
Job dsl

maxock:
New acceptance tests for Job DSL plugin.


src/main/java/org/jenkinsci/test/acceptance/po/CodeMirror.java
```diff
@@ -62,4 +71,14 @@ public void set(String content) {
                     "}" +
                     "codemirror.setValue(arguments[1]);" +
                     "codemirror.save();";
+
+    private static final String scriptGet =
+                    "cmElem = document.evaluate(" +
+                    "        arguments[0], document, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null" +
+                    ").singleNodeValue;" +
```
uhafner: extract the sub-string so it can be used for get and set

src/main/java/org/jenkinsci/test/acceptance/po/GlobalSecurityConfig.java
```diff
@@ -86,4 +86,13 @@ private void selectMarkupFormatter(final String formatter) {
 
         return newInstance(type, this, radio.getAttribute("path"));
     }
+
+    /**
+     * Decides whether script security for Job DSL scripts is enabled.
+     *
+     * @param enable Use script security if true.
+     */
+    public void setJobDslScriptSecurity(boolean enable) {
```
uhafner: Is this control only available for the job-dsl plug-in? If not remove job-dsl from the name of the method.

src/main/java/org/jenkinsci/test/acceptance/po/Control.java
```diff
@@ -112,6 +112,15 @@ public void setAtOnce(String text){
...
     }
 
+    /**
+     * Returns the value of the input field.
+     *
+     * @return the value of the input field.
+     */
+    public String get() {
+        WebElement e = resolve();
+        return ((JavascriptExecutor)driver).executeScript("return arguments[0].value;", e).toString();
```
olivergondza: Do we need javascript for this? `WebElement#getAttribute("value")` should do. Also, note this does not work for all kinds of imputs (textarea, select, etc.)

src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
```diff
@@ -296,4 +296,14 @@ public void installPlugin(File localFile) {
         }
         return shortNamesVersion;
     }
+
+    /**
+     * Enable oder disable the specified plugin.
+     * @param spec plugin id (e.g. "ldap")
+     * @param state enable plugin if true, disable plugin if false
+     */
+    public void enablePlugin(PluginSpec spec, boolean state) {
```
olivergondza: This probably belong to `PluginManager`. Also, the use of `PluginSpec` is probably confusing as the request can never specify a version.

uhafner: It is part of the PluginManager.

So it is simpler to use the parameter `String pluginName`?

olivergondza: > It is part of the PluginManager.

Not sure how I missed that...

> So it is simpler to use the parameter String pluginName?

I would say so.

src/main/java/org/jenkinsci/test/acceptance/po/Control.java
maxock: `WebElement#getAttribute("value")` works fine. Should I remove the method `Control#get()` and do it in `JobDslBuildStep`?
In inputs like textarea and select the method works, but in inputs like checkbox or radio I get nothing useful.

src/test/java/plugins/JobDslPluginTest.java
```diff
@@ -25,12 +41,846 @@
...
+        jobDsl.clickLookOnFilesystem();
+        assertThat(jobDsl.isIgnoreMissingFilesShown(), is(true));
+    }
+
+    /**
+     * Verifies that all configurations, done on the job configuration page,
+     * are saved correctly.
+     */
+    @Test
+    public void should_save_configurations() {
```
uhafner: Can you please have a look why this test fails on our CI server?
https://jenkins.ci.cloudbees.com/job/core/job/acceptance-test-harness-stable/1562/testReport/plugins/JobDslPluginTest/should_save_configurations/

uhafner: Are there additional plugins required to run this test?

maxock: Yes, there are additional plugins needed.
We fixed it in the pull request #333  of my team member.

PR #325, opened 2017-06-08
Git plugin tests

pmwar:
Added tests for several additional behaviours:

- Calculate changelog against a specific branch
- Clean before checkout
- Clean after checkout
- Create tag for every build
- Custom SCM name

Besides that I added Behaviour classes for the remaining additional behaviours.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java
@@ -164,6 +164,19 @@ public class GitRepo implements Closeable {
     }
 
     /**
+     * Create a new branch with the given name
+     * @param name Branch name
+     * @return this, to allow function chaining
+     * @throws IOException
+     * @throws InterruptedException
+     */
+    public GitRepo branch(String name) throws IOException, InterruptedException {
+        git("branch", name);
+
+        return this;
+    }
+
+    /**
      * Refers to a file relative to the workspace directory.
      */
     private File file(String name) {
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java
@@ -96,14 +96,14 @@ public class GitRepo implements Closeable {
         }
     }
 
-    public void git(Object... args) {
-        gitDir(this.dir, args);
+    public String git(Object... args) {
+        return gitDir(this.dir, args);
     }
 
     /**
      * Execute git command in specified directory.
      */
-    public void gitDir(File dir, Object... args) {
+    public String gitDir(File dir, Object... args) {
         List<String> cmds = new ArrayList<>();
         cmds.add("git");
         for (Object a : args) {
@@ -116,13 +116,27 @@ public class GitRepo implements Closeable {
 
         String errorMessage = cmds + " failed";
         try {
-            int r = pb.directory(dir)
+            Process p = pb.directory(dir)
                     .redirectInput(INHERIT)
                     .redirectError(INHERIT)
-                    .redirectOutput(INHERIT).start().waitFor();
+                    .start();
+
+            BufferedReader reader = new BufferedReader(new InputStreamReader(p.getInputStream()));
+            StringBuilder builder = new StringBuilder();
+            String line;
+
+            while ( (line = reader.readLine()) != null) {
+                builder.append(line);
+                builder.append(System.getProperty("line.separator"));
+            }
+
+            int r = p.waitFor();
             if (r != 0) {
                 throw new AssertionError(errorMessage);
             }
+
+            return builder.toString();
+
         } catch (InterruptedException | IOException e) {
             throw new AssertionError(errorMessage, e);
         }
@@ -161,6 +175,19 @@ public class GitRepo implements Closeable {
         } catch (IOException e) {
             throw new AssertionError("Can't change file " + fileName, e);
         }
+    }
+
+    /**
+     * Get sha1 hash of the most recent commit.
+     *
+     * @return Hash value
+     */
+    public String getLastSha1() {
+        return git("rev-parse", "HEAD").trim();
+    }
+
+    public void checkout(String name) {
+        git("checkout", name);
     }
 
     /**
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java
@@ -164,19 +164,6 @@ public class GitRepo implements Closeable {
     }
 
     /**
-     * Create a new branch with the given name
-     * @param name Branch name
-     * @return this, to allow function chaining
-     * @throws IOException
-     * @throws InterruptedException
-     */
-    public GitRepo branch(String name) throws IOException, InterruptedException {
-        git("branch", name);
-
-        return this;
-    }
-
-    /**
      * Refers to a file relative to the workspace directory.
      */
     private File file(String name) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java
```diff
@@ -164,6 +164,19 @@ public void touch(final String fileName) {
     }
 
     /**
+     * Create a new branch with the given name
+     * @param name Branch name
+     * @return this, to allow function chaining
+     * @throws IOException
+     * @throws InterruptedException
+     */
+    public GitRepo branch(String name) throws IOException, InterruptedException {
```
uhafner: There is already a createBranch method

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitScm.java
```diff
@@ -136,4 +198,199 @@ public RecursiveSubmodules(GitScm git, String path) {
...
+
+    public static class PruneStableRemoteBranches extends Behaviour {
+
+        public PruneStableRemoteBranches(GitScm git, String path) {
+            super(git, path);
+            clickLink("Prune stable remote-tracking branches");
+        }
+    }
+
+    // TODO Check if addPath actually works
```
uhafner: Wouldn't it be better to remove this class then?

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitScm.java
```diff
@@ -136,4 +198,199 @@ public RecursiveSubmodules(GitScm git, String path) {
...
+
+        public Behaviour addPath () {
+            String relativePaths = "sparseCheckoutPaths[" + paths.size() + "]/path";
+            btnAdd.click();
+            paths.add(control(relativePaths));
+            return this;
+        }
+    }
+
+    // TODO Check if selStrategy is correctly referenced
```
uhafner: See above: if there is no test yet it is hard too see if the code is working.

src/main/java/org/jenkinsci/test/acceptance/po/Build.java
```diff
@@ -157,6 +157,14 @@ public String getConsole() {
         return visit(getConsoleTextUrl()).getPageSource();
     }
 
+    public URL getGitBuildDataUrl() {
```
uhafner: These methods should not be part of Build: a helper method in the test case would make more sense.

src/main/java/org/jenkinsci/test/acceptance/po/Build.java
```diff
@@ -157,6 +157,14 @@ public String getConsole() {
...
+        return url("git");
+    }
+
+    public String getGitBuildData() {
```
uhafner: see above

src/test/java/plugins/GitPluginTest.java
```diff
@@ -209,6 +210,142 @@ public void update_submodules_recursively() throws IOException, InterruptedExcep
...
 
+    @Test
+    public void calculate_changelog() throws IOException, InterruptedException, SftpException, JSchException {
+        final URL changesUrl;
+        final String TEST_COMMIT_MESSAGE = "Second commit";
+        Build b;
+        GitRepo repo = buildGitRepo();
+
+        repo.branch("testBranch");
+        repo.commit(TEST_COMMIT_MESSAGE);
```
uhafner: Test fails here since there is nothing to commit. The branch in line 220 is already committed.

src/test/java/plugins/GitPluginTest.java
```diff
@@ -209,6 +210,142 @@ public void update_submodules_recursively() throws IOException, InterruptedExcep
...
+        job.removeFirstBuildStep();
+        job.addShellStep("ls && test ! -f untrackedFile.txt");
+        job.save();
+
+        job.startBuild().shouldSucceed();
+
+    }
+
+    @Test
+    public void clean_before_checkout() throws IOException, InterruptedException, SftpException, JSchException {
```
uhafner: Same as clean_after_checkout: is there no test possible that highlights the difference between clean_after/before_checkout ?

src/test/java/plugins/GitPluginTest.java
```diff
@@ -209,6 +210,142 @@ public void update_submodules_recursively() throws IOException, InterruptedExcep
         job.startBuild().shouldSucceed();
     }
 
+    @Test
+    public void calculate_changelog() throws IOException, InterruptedException, SftpException, JSchException {
```
uhafner: throws can be removed now (since https://github.com/jenkinsci/acceptance-test-harness/commit/220c16baf0b4320eff0b59471d7d0bfe929f13b7)

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java
pmwar: I really missed that one. Will remove my .branch() method.

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitScm.java
pmwar: Instead of removing it, I decided to create another test which uses this class. Although it seemed to work, I rewrote .appPath() method to get rid of the ArrayList.

src/test/java/plugins/GitPluginTest.java
pmwar: Method .commit() previously added a new file before committing. I overlooked the changed implementation while rebasing.

src/test/java/plugins/GitPluginTest.java
pmwar: The only thing I can think of spontaneously is that modified files could provoke a conflict when using clean_after_checkout... correct?

src/main/java/org/jenkinsci/test/acceptance/po/Build.java
pmwar: Oh, I didn't take into account that GitBuildData is only available if Git-Plugin is installed.. I'll move it somewhere else

src/test/java/plugins/GitPluginTest.java
```diff
@@ -209,16 +202,226 @@ public void update_submodules_recursively() throws IOException, InterruptedExcep
...
         repo.changeAndCommitFoo("Initial commit");
         return repo;
     }
+
+    /**
+     * Invoked by {@link #clean_after_checkout()} and {@link #clean_before_checkout()}
+     *
+     * @param before Select "clean before" or "clean after"
+     */
+    private void test_clean_while_checkout(boolean before) {
```
uhafner: @MarkEWaite what is the difference between clean before/after checkout? Is there an easy way to verify the differences of these two variants?

MarkEWaite: @uhafner, the only behavioral difference between them is the order of operations.  The checkout happens and is either preceded or followed by a git clean.

The rationale for adding "clean before" was that there are cases where a git checkout will fail if not preceded by a clean.  If the git checkout fails, then the "clean after" would not be run.  Thus, "clean before" was added in an attempt to resolve those cases where "clean after" was not reached because checkout failed.

Refer to [JENKINS-22510](https://issues.jenkins-ci.org/browse/JENKINS-22510) for more discussion on the issue.

The git plugin still has several cases where a checkout does not construct into an existing workspace does not create the same directory structure as a checkout into an empty workspace.  Refer to [JENKINS-38860](https://issues.jenkins-ci.org/browse/JENKINS-38860), [JENKINS-41553](https://issues.jenkins-ci.org/browse/JENKINS-41553), and [JENKINS-43977](https://issues.jenkins-ci.org/browse/JENKINS-43977)

PR #326, opened 2017-06-08
Fixing CachedUpdateCenterMetadataLoader to take into account the version number of Jenkins

jglick:
Solves the problem alluded to in #318, which can be seen for example by

    JENKINS_WAR=…/jenkins-war-2.46.3.war mvn -Dtest=plugins.FolderPluginTest\#folderScopedCredentialsTest test

Previously this would skip the test case, since metadata for an unspecified and presumed trunk version of Jenkins would be downloaded, so ATH would attempt to install `workflow-job` 2.12, then abort when it saw that this requires 2.60+. Now it will download `update-center.json?version=2.46.3` and install `workflow-job` 2.11 as expected.

@reviewbybees


jglick:
@reviewbybees done

PR #327, opened 2017-06-12
LogParser Plugin: Check for unstable and failed builds

engelmi:
Implemented two test cases to check if a build is appropriately marked as failed or unstable.

### DIFF
--- a/src/test/java/plugins/LogParserTest.java
+++ b/src/test/java/plugins/LogParserTest.java
@@ -3,7 +3,6 @@ package plugins;
 import org.jenkinsci.test.acceptance.junit.AbstractJUnitTest;
 import org.jenkinsci.test.acceptance.junit.WithPlugins;
 import org.jenkinsci.test.acceptance.plugins.logparser.LogParserGlobalConfig;
-import org.jenkinsci.test.acceptance.plugins.logparser.LogParserOutputPage;
 import org.jenkinsci.test.acceptance.plugins.logparser.LogParserPublisher;
 import org.jenkinsci.test.acceptance.po.*;
 import org.junit.Before;
@@ -40,7 +39,7 @@ public class LogParserTest extends AbstractJUnitTest {
      * Check whether trend is visible
      */
     @Test
-    public void trendVisible(){
+    public void trendVisible() {
         FreeStyleJob job = jenkins.jobs.create(FreeStyleJob.class, "simple-job");
 
         job.configure(() -> {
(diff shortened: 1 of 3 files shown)


src/test/java/plugins/LogParserTest.java
```diff
@@ -63,6 +62,58 @@ public void trendVisible(){
     }
 
     /**
+     * Test case:
+     * Check for the build to be marked as unstable.
+     */
+    @Test
+    public void checkMarkedUnstableOnWarning() {
+        // Create a special rule set for this test case
```
uhafner: Remove comment and rename rule to additionalRule or specialRule.

src/test/java/plugins/LogParserTest.java
```diff
@@ -63,6 +62,58 @@ public void trendVisible(){
...
+        Resource rule = resource("/logparser_plugin/rules/log-parser-rule-markings");
+        addLogParserRule("error_warning_rule", rule.url.getPath());
+
+        // Create a new freestyle job
```
uhafner: // comments tend to get outdated if the source will be refactored, 
so remove comments that just explain obvious code.

src/test/java/plugins/LogParserTest.java
```diff
@@ -63,6 +62,58 @@ public void trendVisible(){
...
 
     /**
+     * Test case:
+     * Check for the build to be marked as unstable.
+     */
+    @Test
+    public void checkMarkedUnstableOnWarning() {
+        // Create a special rule set for this test case
+        Resource rule = resource("/logparser_plugin/rules/log-parser-rule-markings");
+        addLogParserRule("error_warning_rule", rule.url.getPath());
```
uhafner: add an overloaded addLogParserRule that accepts a Resource

src/test/java/plugins/LogParserTest.java
```diff
@@ -63,6 +62,58 @@ public void trendVisible(){
...
+
+        // Create a new freestyle job
+        FreeStyleJob job = jenkins.jobs.create(FreeStyleJob.class, "simple-job");
+        job.configure(() -> {
+            job.addShellStep("echo marked as warn");
+            LogParserPublisher lpp = job.addPublisher(LogParserPublisher.class);
+            lpp.setMarkOnUnstableWarning(true);
+            lpp.setRule(LogParserPublisher.RuleType.GLOBAL, "" + rule.url.getPath());
```
uhafner: add an overloaded setRule that accepts a Resource

src/test/java/plugins/LogParserTest.java
```diff
@@ -63,6 +62,58 @@ public void trendVisible(){
...
+        });
+
+        job.startBuild().waitUntilFinished().open();
+
+        // Find the status image and check if it is set correctly
+        WebElement imgUnstable = driver.findElement(By.xpath("//div[@id='main-panel']/h1[@class='build-caption page-headline']/img"));
```
uhafner: Wouldn't it bee simpler to call `job.startBuild().waitUntilFinished().shouldBeUnstable()`

src/test/java/plugins/LogParserTest.java
```diff
@@ -63,6 +62,58 @@ public void trendVisible(){
...
+        assertThat(imgUnstable.getAttribute("title"), containsString("Unstable"));
+    }
+
+    /**
+     * Test case:
+     * Check for the build to be marked as failed.
+     */
+    @Test
+    public void checkMarkedFailedOnError() {
```
uhafner: You can combine this and the previous method by using 2 booleans and set:
```
lpp.setMarkOnUnstableWarning(isWarning);
lpp.setMarkOnUnstableError(isError);
```
Then you need not to duplicate code.

src/test/java/plugins/LogParserTest.java
engelmi: I didn't know that method. Thats simpler indeed.

src/test/java/plugins/LogParserTest.java
engelmi: I'll use the overloaded setRule(Resource) for this test cases. But I'll implement this as well as it could be useful for future tests.

src/test/java/plugins/LogParserTest.java
```diff
@@ -63,6 +62,30 @@ public void trendVisible(){
...
+        job.configure(() -> {
+            job.addShellStep("echo marked as warning");
+            LogParserPublisher lpp = job.getPublisher(LogParserPublisher.class);
+            lpp.setMarkOnBuildFail(false);
+            lpp.setMarkOnUnstableWarning(true);
+        });
+        job.startBuild().waitUntilFinished().shouldBeUnstable();
+    }
+
+    /**
```
uhafner: What I actually meant was (to get rid of the duplication):
```
@Test
public void checkMarkedFailed() {
       FreeStyleJob job = jenkins.jobs.create(FreeStyleJob.class);
       configureBuildStatus(true, false);
       job.startBuild().waitUntilFinished().shouldFail();
}
@Test
public void checkMarkedUnstable() {
       FreeStyleJob job = jenkins.jobs.create(FreeStyleJob.class);
       configureBuildStatus(false, true);
       job.startBuild().waitUntilFinished().shouldBeUnstable();
}
```

Or even simpler:
@Test
public void checkMarkedFailed() {
       modifyBuildStatus(true, false, Result.FAILED);
}
@Test
public void checkMarkedUnstable() {
       modifyBuildStatus(false, true, Result.UNSTABLE);
}

PR #328, opened 2017-06-14
[JENKINS-44876] Use a explicit wait for name field

raul-arabaolaza:
[JENKINS-44876](https://issues.jenkins-ci.org/browse/JENKINS-44876)

We are experiencing problems with `JobsMixIn#create` being unable to find the `name` input field on heavy load conditions. I have checked the last page and it seems the field is there so i suspect this is a rendering performance issue because the current code uses a `find` instead of a `waitFor`

Aditionally I switched to `ByFactory` instead of using a Selenium `By#name`

@reviewbybees


varyvol:
🐝

PR #329, opened 2017-06-20
Plotplugin test

stefanSchuhbaeck:
add test cases for Plot Plugin


src/main/java/org/jenkinsci/test/acceptance/plugins/plot/Plot.java
```diff
@@ -0,0 +1,122 @@
...
+     * Create a DataSeries of selected Type and select the Control. If not the first DataSeries create
+     * a new one with Add Button.
+     * @param seriesClass the series class type
+     * @param <S> series class name as generic
+     * @return series
+     */
+    public <S extends DataSeries> S addDataSeries(Class<S> seriesClass){
+        if (this.series.size()>=1){
+            control("repeatable-add").click();
+            //todo insert wait period to make sure the new PageArea appers
```
uhafner: Before merging all TODOs should be done.

src/main/java/org/jenkinsci/test/acceptance/plugins/plot/Plot.java
```diff
@@ -0,0 +1,122 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * @author Stefan Schuhbaeck, Lukasz Koceluch
+ */
+public class Plot extends PageAreaImpl{
+
+    private PlotPublisher parant;
```
uhafner: parent

stefanSchuhbaeck: corrected

src/main/java/org/jenkinsci/test/acceptance/plugins/plot/Plot.java
stefanSchuhbaeck: added elasticSleep

src/test/java/plugins/plot/PlotPluginTest.java
```diff
@@ -0,0 +1,111 @@
...
+        job.startBuild().shouldSucceed();
+        assertThatBuildHasPlot("Some plot","Plots");
+    }
+
+
+    /**
+     * Test if NumberFormatException is logged only if a non-zero length string is found in the data set.
+     */
+    @Test
+    @WithPlugins("plot@1.10") @Issue("JENKINS-25849")
```
olivergondza: This seems like too minor bug to me to have a regression test around in ATH. Also, I suspect you enforce the version 10 as that is where it was fixed. That is undesirable as the test will either succeed or get skipped, but it will not fail when run against unfixed version of plugin.

.gitignore
```diff
@@ -23,3 +23,4 @@ jenkins*home/
 jenkins-plugins/
 .jenkins_test/
 !src/test/resources/jacoco/test/target
+jut.sh
```
olivergondza: This does not belong to shared gitignore.

uhafner: @stefanSchuhbaeck @koceluch 
Can you please remove the .gitignore changes so we can merge this change?

src/test/java/plugins/plot/PlotPluginTest.java
stefanSchuhbaeck: I removed the enforcement of the version. Now the test can fail if the wrong version is used in the test environment. 
For the significant of the test for the ATH. We used this bug as the entry point for the student-project. I can remove if needed for the merge.

uhafner:
@olivergondza any additional changes required?

PR #330, opened 2017-06-21
Mission Control View Plugin

sikelerd:
Add tests for the node and build queue section.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/SlavesMixIn.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/SlavesMixIn.java
@@ -54,4 +54,12 @@ public class SlavesMixIn extends MixIn {
 
         return base + name;
     }
+
+    /**
+     * Get the master node of jenkins.
+     * @return The master node.
+     */
+    public DumbSlave getMaster() {
+        return newInstance(DumbSlave.class, jenkins, "(master)");
+    }
 }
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/SlavesMixIn.java
```diff
@@ -54,4 +54,12 @@ private String remoteFs(String name) {
 
         return base + name;
     }
+
+    /**
+     * Get the master node of jenkins.
+     * @return The master node.
+     */
+    public DumbSlave getMaster() {
```
olivergondza: Note that master *is not* a `DumbSlave`. All shared logic between master and slaves should be in `Node` class which `Jenkins` implements so it should be ok for you to talk to `Jenkins` directly. You might need to push some behavior up the class hierarchy from `DumbSlave` to `Node`, though.

sikelerd: This is a problem, because than there exist two configure pages for the same `ConfigurablePageObject` (the system configuration and the node configuration). So when I start configuring jenkins I come to the system configuration but I intend to configure the node. I would add a method in `Jenkins` like 
```
public Node getMasterNode() {
    return new Node(this, url("computer/%s/", getName())) {
        @Override
            return Jenkins.this.getName();
        }
    };
}
```
to provide access to the master node. Do you have a better solution?

sikelerd: I found a better solution.

olivergondza: All node related configuration options for master should be on the system config page.

PR #332, opened 2017-06-22
Warnings plugin Issue JENKINS-17787 with docker

stefanSchuhbaeck:
reproduce NotSerializableException with codenarcParser when build is on remote Slave using Docker


src/test/java/plugins/WarningsPluginTest.java
```diff
@@ -86,6 +82,289 @@
...
+        assertThat("Assert the proper parser." , thirdHeader[1].trim(), is("Java Compiler (javac)"));
+        assertThat("Assert the proper warning priority." , thirdHeader[2].trim(), is("Priority: Normal"));
+        assertThat("Assert the proper detail text" , thirdValue.trim(), is("division by zero"));
+
+    }
+
+    @WithDocker
+    @Test
+    public void detailsTabContentWithOneWarningTest() throws ExecutionException, InterruptedException {
+        // TODO maybe refactoring is necessary - removing duplicates
```
uhafner: What is the difference between this test and the previous one?

src/test/java/plugins/WarningsPluginTest.java
```diff
@@ -86,6 +82,289 @@
...
+        assertThatActionExists(job, build, "Codenarc Warnings");
+    }
+
+
+    /**
+     * Create a Slave with the supplied Docker Container
+     * @param container     SshContainer for Slave to live in
+     * @return
+     */
+    private Slave getDockerSlave(SshdContainer container){
```
uhafner: return DumbSlave so no casts are needed above

src/test/java/plugins/WarningsPluginTest.java
```diff
@@ -86,6 +82,289 @@
...
+        Build build = job.startBuild().shouldSucceed();
+
+        assertThatActionExists(job, build, "Java Warnings");
+
+    }
+
+    @WithDocker
+    @Test
+    @Issue("JENKINS-17787")
+    @Ignore
```
uhafner: @Ignore("Reproduces JENKINS-17787")

src/test/java/plugins/WarningsPluginTest.java
```diff
@@ -86,6 +82,289 @@
...
+
+        job.configure();
+        job.copyResource(resource("/warnings_plugin/out.txt"));
+        job.addPublisher(WarningsPublisher.class).addWorkspaceFileScanner(JAVA_ID, "out.txt");
+        job.save();
+
+        Build build = job.startBuild().shouldSucceed();
+
+        assertThatActionExists(job, build, "Java Warnings");
+
```
uhafner: Add a check for the number of warnings

src/test/java/plugins/WarningsPluginTest.java
```diff
@@ -86,6 +82,289 @@
...
+    private DockerContainerHolder<JavaContainer> dockerContainer;
+
+    @Inject
+    private SshdContainer sshdDocker;
+
+    @WithDocker
+    @Test
+    public void detailsTabContentTest() throws ExecutionException, InterruptedException {
+        sshdDocker = dockerContainer.get();
+        DumbSlave slave = jenkins.slaves.create(DumbSlave.class);
```
uhafner: If you are using docker then you should use the getDockerSlave method as well

src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_slaves/SshSlaveLauncher.java
```diff
@@ -94,7 +94,9 @@ public SshSlaveLauncher keyCredentials(String username, String key) {
      * to check whether it has already been rendered in the dropdown.
      */
     private void waitForCredentialVisible(final String credUsername) {
-        assertTrue(waitFor().withTimeout(5, TimeUnit.SECONDS).until(() ->
+        System.out.println("WaitforCredName: " + credUsername);
```
stefanSchuhbaeck: removed

src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_slaves/SshSlaveLauncher.java
```diff
@@ -94,7 +94,9 @@ public SshSlaveLauncher keyCredentials(String username, String key) {
...
+//        assertTrue(waitFor().withTimeout(5, TimeUnit.SECONDS).until(() ->
```
stefanSchuhbaeck: removed

src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/DockerAgentContainer/entrypoint.sh
```diff
@@ -1,4 +1,8 @@
 groupadd -g $DOCKER_GROUP docker
+whoami
```
stefanSchuhbaeck: not necessary. reverted to previous commit

src/test/java/plugins/WarningsPluginTest.java
```diff
@@ -43,15 +25,29 @@
...
-import static org.jenkinsci.test.acceptance.po.PageObject.*;
+import static org.jenkinsci.test.acceptance.po.PageObject.createRandomName;
+import static org.junit.Assert.assertTrue;
 
 /**
  * Tests various aspects of the warnings plug-in. Most tests copy an existing file with several warnings into the
  * workspace. This file is then analyzed by console and workspace parsers.
  */
+@WithDocker
+@Category(DockerTest.class)
```
stefanSchuhbaeck: removed

src/test/java/plugins/WarningsPluginTest.java
```diff
@@ -86,6 +82,289 @@
             + "Priority test = Priority.NORMAL;\n"
             + "return new Warning(all, 42, all, all, all);";
 
+    public static final String RESOURCE_WARNING_MAIN_JAVA = "WarningMain.java";
```
stefanSchuhbaeck: changed to private

src/test/java/plugins/WarningsPluginTest.java
stefanSchuhbaeck: included

src/test/java/plugins/WarningsPluginTest.java
```diff
@@ -86,6 +82,289 @@
...
+
+    @Inject
+    private DockerContainerHolder<JavaContainer> dockerContainer;
+
+    @Inject
+    private SshdContainer sshdDocker;
+
+    @WithDocker
+    @Test
+    public void detailsTabContentTest() throws ExecutionException, InterruptedException {
```
koceluch: I thought there are already enough tests **without** docker. We wrote some different scenarios to ensure, that it works also with docker.

src/test/java/plugins/WarningsPluginTest.java
koceluch: detailsTabContentWithOneWarningTest: This tests the table with one Warning.
detailsTabContentTest: This tests the table with more than one Warning. (3 warnings)

Tests with 0 warnings are unnecessary, because in that case there is no table to check.

src/test/java/plugins/WarningsPluginTest.java
koceluch: added.

src/test/java/plugins/WarningsPluginTest.java
```diff
@@ -86,6 +82,289 @@
...
+        assertThat("Assert the proper header array length." , firstHeader.length, is(3));
+        assertThat("Assert the proper file and line." , firstHeader[0].trim(), is("WarningMain.java:10"));
+        assertThat("Assert the proper parser." , firstHeader[1].trim(), is("Java Compiler (javac)"));
+        assertThat("Assert the proper warning priority." , firstHeader[2].trim(), is("Priority: Normal"));
+        assertThat("Assert the proper detail text" , firstValue.trim(), is("redundant cast to TextClass"));
+
+        // removing tested entry from map.
+        map.remove(firstKey);
+
+        String secondKey = map.firstKey();
```
koceluch: refactored tests to avoid code duplication.

src/test/java/plugins/WarningsPluginTest.java
```diff
@@ -86,6 +82,289 @@
...
+        WarningsAction action = createJavaResultAction(build);
+
+        action.open();
+
+        SortedMap<String, String> map = action.getDetailsTabContents();
+        Set<Map.Entry<String, String>> set = map.entrySet();
+
+        assertThat("Assert the proper detail count.", set.size(), is(3));
+
+        // String is a comparable -> keys are sorted.
```
koceluch: refactored tests to avoid code duplication.

src/test/java/plugins/WarningsPluginTest.java
stefanSchuhbaeck: changed to DumbSlave

src/test/java/plugins/WarningsPluginTest.java
stefanSchuhbaeck: using getDockerSlave  now

batmat:
@uhafner hallo! That's great to see so many improvements ongoing, thanks a lot for encouraging those! 

Do you think we could/should release an ATH version? I think it would be valuable to have a pointcut. That would enable people using the released version check if there are changes to be adjusted more progressively.

I can definitely handle the release, but I just prefer to get an ack from you and/or Oliver before proceeding.

Thanks! cc @olivergondza

uhafner:
Wouldn't it make sense to wait for all student PRs? Otherwise we are in a somewhat intermediate state. (And additionally, the status should be stable, too many tests are not yet stable, see https://jenkins.ci.cloudbees.com/job/core/job/acceptance-test-harness-stable/).

batmat:
@uhafner we can definitely wait if you say so, and you're the one working on this. I'm simply inclined to like releasing more often so that downstream users can assess/test smaller chunks. 
But maybe let's possibly discuss/follow up this on the dev ML then.

uhafner:
BTW: who is using the ATH as released version?

PR #333, opened 2017-06-22
JobDSL: Added ListView tests

mreinbol:


src/test/java/plugins/JobDslPluginTest.java
```diff
@@ -817,6 +822,615 @@ public void should_alert_if_generated_job_changed() {
...
+     * Test if a new ListView is created via Job DSL script TextBox.
+     * The created View shall contain a description.
+     */
+    @Test
+    public void is_ListView_created() {
+        String descriptionText = "This is the description of testView";
+        String jobDslScript = "listView('"+ LIST_VIEW_NAME +"') {\n" +
+                "    description('"+descriptionText+"')\n" +
+                "}";
+        openNewlyCreatedListView(jobDslScript, LIST_VIEW_NAME);
```
uhafner: shouldn't this method better return the created view?

src/test/java/plugins/JobDslPluginTest.java
```diff
@@ -817,6 +822,615 @@ public void should_alert_if_generated_job_changed() {
...
+        WebElement description = find(By.xpath("//div[@id='description']/div"));
```
uhafner: move to View (there is already a setter there)

src/test/java/plugins/JobDslPluginTest.java
```diff
@@ -817,6 +822,615 @@ public void should_alert_if_generated_job_changed() {
...
+        Assert.assertThat(description.getText(), containsString(descriptionText));
```
uhafner: assertThat (Assert is already imported)

src/test/java/plugins/JobDslPluginTest.java
```diff
@@ -817,6 +822,615 @@ public void should_alert_if_generated_job_changed() {
...
+                "    lastFailure()\n" +
+                "    lastDuration()\n" +
+                "    buildButton()\n" +
+                "  }\n" +
+                "  jobs{\n" +
+                "    name('"+job1.name+"')\n" +
+                "  }\n" +
+                "}";
+        openNewlyCreatedListView(jobDslScript, LIST_VIEW_NAME);
+        Assert.assertThat(driver, hasElement(By.xpath("//th[contains(@tooltip, 'Status of the last build')]")));
```
uhafner: The XPATH should be part of the view. (Or is it possible to work with the existing getColumn method?)

Or even better:
```
assertThat(view, hasColumnHeader("Name"))
```

src/test/java/plugins/JobDslPluginTest.java
```diff
@@ -817,6 +822,615 @@ public void should_alert_if_generated_job_changed() {
...
+                "  columns {\n" +
+                "    name()\n" +
+                "  }\n" +
+                "  jobs {\n" +
+                "    names('"+jobs.get(0).name+"', '"+jobs.get(1).name+"', '"+jobs.get(2).name+"', '"+jobDisabled+"')\n" +
+                "  }\n" +
+                "}";
+        openNewlyCreatedListView(jobDslScript, LIST_VIEW_NAME);
+        Assert.assertThat(driver, not(hasElement(By.xpath("//tbody/tr[contains(@id, '"+jobDisabled+"')]"))));
+        Assert.assertThat(driver, hasElement(By.xpath("//tbody/tr[contains(@id, '"+jobs.get(0).name+"')]")));
```
uhafner: if the right part of this line could not be placed into a method in the View class then there should be at least a helper method in this test case that avoids the duplications.

src/test/java/plugins/JobDslPluginTest.java
```diff
@@ -817,6 +822,615 @@ public void should_alert_if_generated_job_changed() {
...
+     * This method creates a given amount of jobs and returns them in a List.
+     * The caller can choose whether to build the jobs or not.
+     * @param amount amount to be created
+     * @param isBuilt flag if the jobs should be built
+     * @return List of Jobs
+     */
+    private List<Job> createAmountOfJobs(int amount, boolean isBuilt) {
+        List<Job> jobs = new ArrayList<>();
+        if (isBuilt) {
+            for (int i = 0; i < amount; i++) {
```
uhafner:
```
for (int i = 0; i < amount; i++) {
    Job job;
    if (isBuild) {
        job = createJobAndBuild();
    }
    else {
        job = jenkins.jobs.create(FreeStyleJob.class)
    }
   jobs.add(job);
 }
```

src/test/java/plugins/JobDslPluginTest.java
mreinbol: The description should be set via script, since it is intended to test this DSL element

uhafner:
Can you please resolve the conflict in JobDslPluginTest.java so we can merge using the UI?

PR #352, opened 2017-08-31
[JENKINS-46564] Added acceptance test for custom configurations files

fcojfernandez:
[JENKINS-46564](https://issues.jenkins-ci.org/browse/JENKINS-46564)

It is possible to create custom config files since config-file-provider-2.16.0. Some other plugins like maven-plugin test how they manage their configuration files, but this feature is not tested by any plugin. It would be recommendable to create a test to cover it.

@reviewbybees

### DIFF
--- a/src/test/java/plugins/ConfigFileProviderTest.java
+++ b/src/test/java/plugins/ConfigFileProviderTest.java
@@ -27,6 +27,8 @@ public class ConfigFileProviderTest extends AbstractJUnitTest {
     private static final String CRED_USR = "fakeUser";
     private static final String CRED_PWD = "fakePass";
     private static final String SERVER_ID = "fakeServer";
+    private static final String CUSTOM_CONF_CONTENT = "test_custom";
+
 
     @Before
     public void setup() {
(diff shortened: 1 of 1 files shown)


src/test/java/plugins/ConfigFileProviderTest.java
```diff
@@ -91,4 +94,36 @@ private String createPipelineAndGetConsole(final MavenSettingsConfig mvnConfig)
         return b.getConsole();
     }
 
+    @Test
+    public void testCustomConfigFile() {
+        final CustomConfig customConfig = this.createCustomConfig();
+        final String jobLog = this.createPipelineAndGetConsole(customConfig);
+
+        assertThat(jobLog, containsString(CUSTOM_CONF_CONTENT));
```
olivergondza: I presume you intend to verify the content of the file is injected correctly but the reason this passes is the filename is the same as the content  that is echoed with the command.

src/test/java/plugins/ConfigFileProviderTest.java
```diff
@@ -91,4 +94,36 @@ private String createPipelineAndGetConsole(final MavenSettingsConfig mvnConfig)
...
+        return customConfig;
+    }
+
+    private String createPipelineAndGetConsole(final CustomConfig customConfig) {
+        final WorkflowJob job = jenkins.jobs.create(WorkflowJob.class);
+        job.script.set(String.format("node {\n" +
+                "    configFileProvider(\n" +
+                "        [configFile(fileId: '%s')]) {\n" +
+                "            \n" +
+                "        sh 'echo test_custom '\n" +
```
olivergondza: `s/echo/cat/`

PR #353, opened 2017-09-02
[JENKINS-41196] demonstrate that core component override plugin can work in ATH

kohsuke:
Most of this change just adds a capability to test a plugin that's not yet released and available in the update center.

The test itself worked as expected once this was cleared.

### DIFF
--- a/docs/SUT-VERSIONS.md
+++ b/docs/SUT-VERSIONS.md
@@ -21,17 +21,10 @@ When you are testing locally developed Jenkins plugin, you'd like the test harne
 local version as opposed to download the plugin from update center. This can be done by instructing the harness
 accordingly.
 
-One way to do so is to set the environment variable:
-
-    $ env ldap.jpi=/path/to/your/ldap.jpi mvn test
+    $ env LOCAL_JARS=path/to/your/ldap.jpi:path/to/another.jpi
 
 You can also do this from [the groovy wiring script](WIRING.md).
-
-    envs['ldap.jpi'] = '/path/to/your.ldap.jpi'
-
-Since newer shells do not support environment variables with hyphens in the name, rather than `scm-api.jpi` you may use `SCM_API_JPI`.
-
-This same scheme works for a plugin that's not yet released.
+This scheme also works for a plugin that's not yet released.
 
 ### Install plugins from local maven repository
 
--- a/docs/SUT-VERSIONS.md
+++ b/docs/SUT-VERSIONS.md
@@ -31,6 +31,8 @@ You can also do this from [the groovy wiring script](WIRING.md).
 
 Since newer shells do not support environment variables with hyphens in the name, rather than `scm-api.jpi` you may use `SCM_API_JPI`.
 
+This same scheme works for a plugin that's not yet released.
+
 ### Install plugins from local maven repository
 
 As a convenience, you can also run with the variable `LOCAL_SNAPSHOTS=true`.
--- a/src/main/java/org/jenkinsci/test/acceptance/update_center/LocalOverrideUpdateCenterMetadataDecoratorImpl.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/update_center/LocalOverrideUpdateCenterMetadataDecoratorImpl.java
@@ -1,17 +1,17 @@
 package org.jenkinsci.test.acceptance.update_center;
 
 import com.cloudbees.sdk.extensibility.Extension;
-import java.io.File;
-import java.util.Iterator;
-import java.util.Locale;
-import java.util.Map;
-import javax.xml.parsers.DocumentBuilderFactory;
-
 import org.eclipse.aether.artifact.DefaultArtifact;
 import org.eclipse.aether.util.version.GenericVersionScheme;
 import org.eclipse.aether.version.Version;
 import org.eclipse.aether.version.VersionScheme;
 import org.w3c.dom.NodeList;
+
+import javax.xml.parsers.DocumentBuilderFactory;
+import java.io.File;
+import java.util.Iterator;
+import java.util.Locale;
+import java.util.Map;
 
 /**
  * Allow local plugins specified via environment variables to override plugin metadata from update center.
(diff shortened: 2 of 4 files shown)


src/main/java/org/jenkinsci/test/acceptance/update_center/LocalOverrideUpdateCenterMetadataDecoratorImpl.java
```diff
@@ -61,32 +61,31 @@ public void decorate(UpdateCenterMetadata ucm) {
...
-            if (!file.exists()) throw new IllegalArgumentException("Plugin file for " + name + " does not exist: " + file.getAbsolutePath());
+            if (!file.exists()) throw new IllegalArgumentException("Plugin file for " + key + " does not exist: " + file.getAbsolutePath());
+
             PluginMetadata m = PluginMetadata.LocalOverride.create(file);
-            System.err.println("Overriding " + name + " " + stock.getVersion() + " with local build of " + m.getVersion());
+            PluginMetadata stock = ucm.plugins.get(m.getName());
+            if (stock == null) {
+                System.err.println("Creating new plugin " + m.getName() + " with local build of " + m.getVersion());
+            } else {
+                System.err.println("Overriding " + m.getName() + " " + stock.getVersion() + " with local build of " + m.getVersion());
```
oleg-nenashev: Maybe makes sense to create a logger instead

src/main/java/org/jenkinsci/test/acceptance/update_center/LocalOverrideUpdateCenterMetadataDecoratorImpl.java
```diff
@@ -61,32 +61,31 @@ public void decorate(UpdateCenterMetadata ucm) {
         }
         for (Map.Entry<String,String> e : System.getenv().entrySet()) {
             String key = e.getKey();
-            String name;
-            if (key.endsWith(".jpi")) {
-                name = key.replace(".jpi", "");
-            } else if (key.endsWith("_JPI")) { // http://stackoverflow.com/a/36992531/12916
-                name = null;
-                for (String _name : ucm.plugins.keySet()) {
-                    if ((_name.toUpperCase(Locale.ENGLISH).replaceAll("[^A-Z0-9_]", "_") + "_JPI").equals(key)) {
```
jglick: So what is now happening that allows this to be deleted? Does `LocalOverride.create` reliably infer the name? In that case there is no reason to set distinct environment variables at all. You could simply define

    LOCAL_OVERRIDES=/path/to/whatever.jpi:/another/path/to/something-else-1.33-irrelevant-filename.jpi

src/test/java/plugins/RemotingPluginTest.java
```diff
@@ -0,0 +1,61 @@
...
+     * <p>
+     * We check this by:
+     * <ol>
+     *   <li>Testing the version of the plugin as declared installed
+     *   <li>Making sure that the actual jar file in the system is replaced
+     *  </ol>
+     */
+    @Test
+    @WithPlugins({"remoting"})
+    @Ignore // meant to be used to locally demonstrate JENKINS-41196 support
```
jglick: I think you can omit that, since it will not run the test unless the plugin is available anyway.

src/test/java/plugins/RemotingPluginTest.java
```diff
@@ -0,0 +1,61 @@
...
+    public void test() throws IOException, InterruptedException {
+        final String version = "3.12-SNAPSHOT";
```
jglick: Presumably you need to actually fix this.

src/main/java/org/jenkinsci/test/acceptance/update_center/LocalOverrideUpdateCenterMetadataDecoratorImpl.java
kohsuke: Yes, the name is inferred from the plugin by looking at its manifest.

src/test/java/plugins/RemotingPluginTest.java
kohsuke: No, the test will fail saying that the said plugin is not available in the update center, as it should be. If you made a typo in a plugin name you'd want the test to fail to tell you something is wrong, instead of ignoring a test.

src/test/java/plugins/RemotingPluginTest.java
kohsuke: Addressed

src/main/java/org/jenkinsci/test/acceptance/update_center/LocalOverrideUpdateCenterMetadataDecoratorImpl.java
kohsuke: Added a variable `LOCAL_JARS` to do this

docs/SUT-VERSIONS.md
```diff
@@ -21,15 +21,10 @@ When you are testing locally developed Jenkins plugin, you'd like the test harne
 local version as opposed to download the plugin from update center. This can be done by instructing the harness
 accordingly.
 
-One way to do so is to set the environment variable:
-
-    $ env ldap.jpi=/path/to/your/ldap.jpi mvn test
+    $ env LOCAL_JARS=path/to/your/ldap.jpi:path/to/another.jpi
```
olivergondza: How come we never though of this?

jglick:
@olivergondza this was not supposed to be merged unless and until the core change was merged.

jglick:
I mean, it is harmless in that the test is still `@Ignore`d, but as a rule we do not want `master` to contain experimental code.

olivergondza:
Right, I have got so excited about the plugin overrides I have merged the whole thing. Anyway, I reverted the [offending commit](https://github.com/jenkinsci/acceptance-test-harness/commit/2ef2ee88b27764f09258eb9227bf0e9497e6ca49) and created https://github.com/jenkinsci/acceptance-test-harness/pull/356 to deliver after the feature is ready.

PR #354, opened 2017-09-15
[JENKINS-46754] Reproduce bug and demonstrate fix

jglick:
Tests https://github.com/jenkinsci/jenkins/pull/3026.

@reviewbybees @jenkinsci/code-reviewers

PR #355, opened 2017-09-15
When using $LOCAL_JARS, give installPlugin a moment to catch up

jglick:
When checking #354 with a local plugin build from https://github.com/jenkinsci/ssh-credentials-plugin/pull/30, the test did not work: first Jenkins installed `ssh-credentials@1.14-SNAPSHOT` as requested, and then immediately afterwards it clobbered it with `ssh-credentials@1.13` from the UC. After the restart (BTW why do we need to restart when all the plugins are fresh installations?!), 1.13 was installed, thus defeating the purpose of specifying `$LOCAL_JARS`.

I found that by introducing a delay into the process, the asynchronous update center jobs had time to actually register that `ssh-credentials@1.14-SNAPSHOT` was already there and thus did not need to be added as a dependent installation of `ssh-slaves`.

I do not like this but I have not found a better way yet. Not clear if there is a core bug here, or if ATH just needs to poll `updateCenter/api/json?tree=jobs[success]` (?) to wait for `pluginManager/uploadPlugin` to actually take effect. (I think the CLI command `install-plugin` blocks until it is finished, but the REST variant definitely does not.)

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
@@ -301,6 +301,13 @@ public class PluginManager extends ContainerPageObject {
                     IOUtils.toString(response.getEntity().getContent()));
         } else {
             System.out.format("Plugin %s installed\n", localFile);
+            if (System.getenv("LOCAL_JARS") != null) {
+                try {
+                    Thread.sleep(3000); // TODO find a better way to ensure that core has actually applied the update
+                } catch (InterruptedException x) {
+                    x.printStackTrace();
+                }
+            }
         }
     }
 


olivergondza:
`UpdateCenter#waitForInstallationToComplete` should take care of the waiting and restarting if needed.

jglick:
Does not suffice. That waits at the _end_ of installing all plugins. The problem I observed here is that _in the middle_ of installing some plugins, Jenkins core had not yet registered an earlier update request, and so went and redundantly installed a different version of the same plugin as a dependency.

dwnusbaum:
@jglick Do you have the logs from the local build? It looks like dependencies are [checked in the request handler thread](https://github.com/jenkinsci/jenkins/blob/34a877814656f554913cd98ff3835b7e6f99a648/core/src/main/java/hudson/model/UpdateSite.java#L1208) without synchronization before they are installed in the background, so it seems plausible that simultaneous requests could check dependencies at the same time, and then overwrite each other's plugins when doing the install.

🐝  This change looks good to me as a workaround.

jglick:
> Do you have the logs from the local build?

Not handy. Since #354 is merged, it should be possible to recreate the situation by running `acceptance-test-harness` on that test from `master` with `JENKINS_WAR` set to 2.73, with `LOCAL_JARS` pointing to a `ssh-credentials@1.14-SNAPSHOT` build. You would see some log messages go by about `1.14-SNAPSHOT` being installed, followed shortly thereafter from a message about `1.13`, then the Jenkins restart; when the test runs, `1.13` will be installed, and thus the launch log will show no stack trace. With this patch applied, you will see only `1.14-SNAPSHOT` installed, and the launch log will show the stack trace.

At any rate, @reviewbybees done but I would like @olivergondza’s opinion on whether this is “good enough for now” or whether we should try to fix ATH to properly track the status of install/update jobs.

olivergondza:
It is certainly good enough for now.

PR #359, opened 2017-09-21
WIP: Switch to using phantomjs for less overhead (in theory) when running in CI

rtyler:
ci.jenkins.io has more capacity than my laptop, so I'm curious to see how executing this works out :smile:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,7 +12,7 @@ pipeline {
     stages {
         stage('Run ATH') {
             steps {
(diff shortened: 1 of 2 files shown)


rtyler:
For my own remembering: it appears that Phantom when installed from Apt will cause failures because some assets are missing (AFAICT) from the package. The `Dockerfile` should grab the static pre-built biinaries from the phantomjs site.

rtyler:
phantom is pretty dead, this is superceded by #360

PR #360, opened 2017-09-27
WIP: New selenium 3.6.0

rtyler:
Closes #284

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/FallbackConfig.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/FallbackConfig.java
@@ -2,16 +2,15 @@ package org.jenkinsci.test.acceptance;
 
 import javax.annotation.CheckForNull;
 import javax.inject.Named;
+
 import java.io.File;
 import java.io.IOException;
-import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
-import java.util.logging.Level;
 
 import org.apache.commons.lang.StringUtils;
 import org.eclipse.aether.RepositorySystem;
@@ -36,9 +35,9 @@ import org.jenkinsci.test.acceptance.utils.SauceLabsConnection;
 import org.jenkinsci.test.acceptance.utils.aether.ArtifactResolverUtil;
 import org.jenkinsci.test.acceptance.utils.mail.MailService;
 import org.jenkinsci.test.acceptance.utils.mail.Mailtrap;
-import org.jenkinsci.test.acceptance.utils.pluginreporter.ConsoleExercisedPluginReporter;
 import org.jenkinsci.test.acceptance.utils.pluginreporter.ExercisedPluginsReporter;
 import org.jenkinsci.test.acceptance.utils.pluginreporter.TextFileExercisedPluginReporter;
+import org.jenkinsci.test.acceptance.utils.pluginreporter.ConsoleExercisedPluginReporter;
 import org.junit.runners.model.Statement;
 import org.openqa.selenium.Dimension;
 import org.openqa.selenium.UnsupportedCommandException;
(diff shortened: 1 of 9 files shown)


uhafner:
Maybe using [headless chome](https://developers.google.com/web/updates/2017/04/headless-chrome) helps to improve the performance. Hopefully it works with this new selenium version.

rtyler:
@uhafner I tried actually, unfortunately in my environment, `options.setHeadless(true)` caused weird driver errors with `chromedriver` and Chrome 59.

Sticking with `geckodriver` and Firefox has been pretty consistent for me. Still need to figure out why some things aren't working in the Docker container however.

src/main/java/org/jenkinsci/test/acceptance/FallbackConfig.java
```diff
@@ -98,33 +98,25 @@ private WebDriver createWebDriver(TestName testName) throws IOException {
...
-
+            FirefoxOptions firefoxOptions = new FirefoxOptions();
+            firefoxOptions.addPreference(LANGUAGE_SELECTOR, "en");
             // Config screen with many plugins can cause FF to complain JS takes too long to complete - set longer timeout
-            profile.setPreference(DOM_MAX_SCRIPT_RUN_TIME, (int)getElasticTime().seconds(600));
-            profile.setPreference(DOM_MAX_CHROME_SCRIPT_RUN_TIME, (int)getElasticTime().seconds(600));
-
-            FirefoxBinary binary = new FirefoxBinary();
-            if (display != null) {
-                binary.setEnvironmentProperty("DISPLAY", display);
```
olivergondza: Why dropping `DISPLAY` support?

vnc.sh
```diff
@@ -14,3 +14,4 @@ if command -v vncviewer >/dev/null 2>&1; then
 fi
 
 echo export BROWSER_DISPLAY=$display
+echo export DISPLAY=$display
```
olivergondza: We use different variable name for a reason. So you can set `BROWSER_DISPLAY` in terminal you work in and only browser windows will be launched in vnc session.

src/main/java/org/jenkinsci/test/acceptance/FallbackConfig.java
rtyler: This was some older code which I had cherry-picked from #284. I'll re-instate that code and try it out.

vnc.sh
rtyler: Ah, I'll make a comment to that effect and revert this change

src/main/java/org/jenkinsci/test/acceptance/FallbackConfig.java
rtyler: It looks like I was a bit premature in my statement. This code must come out because it's geckodriver (as far as I can tell) which must know about the "display". I'm currently experimenting with a `geckodriver.sh` wrapper script which will do the right thing with regards to swapping environment variables around.

rtyler:
I have been a little bit pre-occupied over the past two weeks, I'm hoping I can pick this work back up over the next couple days

src/main/resources/ath-container/Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN apt-get clean && \
         git \
         imagemagick \
         iptables \
-        firefox-esr \
+        firefox \
```
jtnord: https://seleniumhq.wordpress.com/2017/08/09/firefox-55-and-selenium-ide/

xenial seems to use [56+](https://launchpad.net/ubuntu/xenial/+source/firefox) so using esr is a good idea (if it exists...)

lvotypko:
It seems that we will have to switch on selenimu 3.x, since the old one does not work with new versions of Firefox anymore.

@rtyler is there any progress? Can you please resolve the conflicts? Thank you!

rtyler:
@lvotypko unfortunately I no longer have any time to spend cleaning up this pull request. So please feel free to take this work and run with it!

lvotypko:
@rtyler Ok, I did a few changes to make it work with current master branch and it seems working.  I will try to finish it, thank you!

olivergondza:
Thanks, I am closing this then.

PR #361, opened 2017-10-02
API to add additional SSH pubkey to Git server

amuniz:
Having a single authorized key is not much flexible. This adds a way to add more authorized keys to access the Git repository started by `GitContainer`.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitContainer.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitContainer.java
@@ -3,6 +3,7 @@ package org.jenkinsci.test.acceptance.docker.fixtures;
 import java.io.IOException;
 import java.net.URL;
 
+import org.jenkinsci.test.acceptance.docker.Docker;
 import org.jenkinsci.test.acceptance.docker.DockerContainer;
 import org.jenkinsci.test.acceptance.docker.DockerFixture;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitContainer.java
```diff
@@ -44,4 +45,14 @@ public String getRepoUrlInsideDocker(String alias) throws IOException {
         return "ssh://git@" + alias + REPO_DIR;
     }
 
+    /**
+     * Add an additional certificate to <code>~/.ssh/authorized_keys</code>
+     * @param cert the certificate public key
+     */
+    public void addSSHCertificate(String cert) throws IOException, InterruptedException {
```
jtnord: NIT: not a certificate, a key..

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitContainer.java
```diff
@@ -44,4 +45,14 @@ public String getRepoUrlInsideDocker(String alias) throws IOException {
...
+        Docker.cmd("exec", getCid()).add("/bin/bash",  "-c",  "echo " + cert + " >> /home/git/.ssh/authorized_keys")
```
varmenise: NIT: would not this be the result of the String builder on exec, id, /bin/bash, -c etc etc?
Just wondering why you passed some in the initializer and some others in the .add

amuniz: Yeah, it would be the same I think, but I wanted to separate them on different calls for clarity.

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitContainer.java
amuniz: Fixed.

batmat:
@olivergondza we would need to have a release with this, and I see from the history the last release was around August, do you mind if I do a release in the current state? I think it makes sense in any case to have ATH release with a slightly smaller changelog.

I will wait for your approval, or possibly I'll just go ahead if you are not available in the next few days.

Thanks!

olivergondza:
@batmat, I do not depend on ATH releases so I cut one only when people request it explicitly. Feel free to release as you meed to.

batmat:
Cool, thanks for answering, doing.

batmat:
`1.57` released. Thanks everyone.

PR #363, opened 2017-10-05
Do not set credential ID for foreman tests

pjanouse:
Added JUnit @Cathegory


src/test/java/plugins/ForemanNodeSharingPluginTest.java
```diff
@@ -77,7 +80,10 @@
         sc.username.set("test");
         sc.selectEnterDirectly().privateKey.set(sshslave1.getPrivateKeyString());
         sc.scope.select("GLOBAL");
-        sc.setId("test");
+
+        // Not supported in 1.x Jenkins
+        // sc.setId("test");
```
olivergondza: Let's remove it if we it is not needed.

pjanouse: Done

PR #365, opened 2017-10-06
Update tests in SmokeTest category

olivergondza:
Updating the collection of smoke tests for ATH to cover fundamental use-cases from core and some of the prominent plugins.

See https://groups.google.com/forum/#!topic/jenkinsci-dev/R8w7uhh-xbgfor discussion.
CC @jenkinsci/code-reviewers

### DIFF
--- a/src/test/java/core/ConfigureSlavesTest.java
+++ b/src/test/java/core/ConfigureSlavesTest.java
@@ -21,7 +21,7 @@ public class ConfigureSlavesTest extends AbstractJUnitTest {
 
     @Test
     @Category(SmokeTest.class)
-    public void tie_job_to_specifid_label() throws Exception {
+    public void tie_job_to_specified_label() throws Exception {
         FreeStyleJob j = jenkins.jobs.create();
 
         Node s = slave1.install(jenkins).get();
--- a/src/test/java/core/ConfigureSlavesTest.java
+++ b/src/test/java/core/ConfigureSlavesTest.java
@@ -1,11 +1,13 @@
 package core;
 
 import org.jenkinsci.test.acceptance.junit.AbstractJUnitTest;
+import org.jenkinsci.test.acceptance.junit.SmokeTest;
 import org.jenkinsci.test.acceptance.po.Build;
 import org.jenkinsci.test.acceptance.po.FreeStyleJob;
 import org.jenkinsci.test.acceptance.po.Node;
 import org.jenkinsci.test.acceptance.slave.SlaveController;
 import org.junit.Test;
+import org.junit.experimental.categories.Category;
 
 import javax.inject.Inject;
 
@@ -18,6 +20,7 @@ public class ConfigureSlavesTest extends AbstractJUnitTest {
     SlaveController slave1;
 
     @Test
+    @Category(SmokeTest.class)
     public void tie_job_to_specifid_label() throws Exception {
         FreeStyleJob j = jenkins.jobs.create();
 
--- a/src/test/java/core/FreestyleJobTest.java
+++ b/src/test/java/core/FreestyleJobTest.java
@@ -6,6 +6,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
 
 import org.jenkinsci.test.acceptance.junit.AbstractJUnitTest;
+import org.jenkinsci.test.acceptance.junit.SmokeTest;
 import org.jenkinsci.test.acceptance.junit.Wait;
 import org.jenkinsci.test.acceptance.po.Artifact;
 import org.jenkinsci.test.acceptance.po.ArtifactArchiver;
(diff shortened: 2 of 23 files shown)


olivergondza:
@daniel-beck, I do not see why not. Though, I see reasons why I won't be doing that today  :stuck_out_tongue:. I guess that is way to see how long will this take to execute so we will see how big the set can actually be for the tests to run in a reasonable time for PR runs.

rsandell:
+1
But I would suggest something from `GitPluginTest` and `JUnitPluginTest` as well.

jglick:
> Is there a way to get a downstream PR build for core

-1 on that. Some of these use `@WithPlugins`, meaning their behavior will vary depending on the current contents of the update center, making the core build unreproducible.

jglick:
> perhaps `WorkflowPluginTest#linearFlow`

…though this has a downside: the builds retrieves Maven from an external download site. Seems like we need to revisit our default tool installation locations. Maybe offer them from a Jenkins download site? Or just adjust this test so that it can load common tools locally, say from `jenkins-test-harness-tools`? Or modify it to use a Dockerized agent with Maven preloaded?

olivergondza:
> Some of these use @WithPlugins, meaning their behavior will vary depending on the current contents of the update center

Such failures will probably not be strictly relevant to PRs but it would illustrate how the change would behave if released as is. I would argue this makes sense for core master builds.

jglick:
> it would illustrate how the change would behave if released as is

So would daily test failures run against `master`, or probably the last LTS line too. There are other ways interested parties—generally _not_ the PR author—could get such information.

Strong -1 on introducing any source of nondeterminism to PR builds. To the greatest extent we can possibly manage, a PR build should fail if and only if there is a mistake in the PR diff; and `master` builds should always be passing.

jglick:
(This also implies that not only should any introduction of ATH into core builds either run exclusively no-plugin tests or use pinned versions of necessary plugins and all dependencies; but the version of ATH itself should also be pinned. If you want to update any of these versions to something more recent, file a PR and see if it works.)

daniel-beck:
@jglick Then @oleg-nenashev and I cannot be the only ones looking for issues to include into the weekly release, and we need to do it during the week, not on weekends. There's no way we're able to identify whatever issue broke master in the little time we have.

uhafner:
We should also adapt the JavaDoc in SmokeTest to reflect the new usage of the category (before it was intended to show that the ATH still works).

olivergondza:
So where are we? Do we want to limit the smoke tests to avoid plugin installation so it would not cover (a) plugin use-cases and (b) plugin installation itself? I do not oppose, just trying to clarify what is the consensus if there is any.

daniel-beck:
IMO smoke test minimally need to include the plugins that are part of the default install (the 'suggested' subset of https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/jenkins/install/platform-plugins.json).

Those are what users are likely to experience out of the box, and I doubt a single user would GAF about the distinction between core and plugins, when Jenkins come pre-broken because any of those don't work right.

We also need to run them for PR builds for the reason I mention above. We're already understaffed for proper weekly release preparation, so limiting ATH runs to master won't cut it. Perhaps there's a way to not fail the PR build but instead add a comment to the PR when the smoke test fails?

jglick:
> Then @oleg-nenashev and I cannot be the only ones looking for issues to include into the weekly release, and we need to do it during the week, not on weekends.

Let me put it this way: who makes the decision to cut a weekly release to begin with? That person should be responsible for verifying that there are no significant known regressions in the Jenkins test ecosystem. So long as that is @kohsuke well then @kohsuke needs to do that, or find someone reliable to delegate to on a weekly basis. Or simply rerun these tests as part of the weekly release script itself.

> Perhaps there's a way to not fail the PR build but instead add a comment to the PR when the smoke test fails?

There is a `githubNotify` step. I am not sure offhand how GH will display “success” with a custom message, and whether anyone will pay attention to such a message.

At any rate, I am still generally opposed to doing this as part of a core PR build because it implies that the problem is somehow associated with that core PR, which is just one possibility out of several.

MarkEWaite:
I'm willing to attempt to monitor the smoke tests 2-3 evenings each week.

Will that be enough to begin?

If I report those results to Kohsuke or to some other location, will that
satisfy your request, Jesse?

I'm not an acceptance test harness user (yet), and am still not persuaded
that the acceptance test harness technique is as effective as other
techniques.  I'm willing to spend some evening and weekend time trying to
convince myself that the technique does more good than harm, and that the
project will benefit from the effort.

Will I need to run the smoke tests on my own equipment, or is there already
a location in the infrastructure which hosts the tests?

Where do I find instructions that will introduce me to running the smoke
tests from within the acceptance test harness?

In the case of flaky tests, I should warn everyone about my biases.  I've
not had really great results in the past when trying to persuade others to
fix flaky tests, so my tendency will be to submit pull requests to delete
flaky tests, rather than attempt to fix them.

Mark Waite

olivergondza:
Alright, removing the category from plugin tests so this can be hooked to core PRs: https://github.com/jenkinsci/jenkins/pull/3073

daniel-beck:
@olivergondza What's needed to make this part of core?

olivergondza:
@daniel-beck, wdym, running the smoke tests during core CI builds? https://github.com/jenkinsci/jenkins/pull/3073

PR #370, opened 2017-10-16
[JENKINS-47264] Increase test coverage and fix UI wrapper

fcojfernandez:
[JENKINS-47264](https://issues.jenkins-ci.org/browse/JENKINS-47264)

Features:

- In git-3.4.0, changes refactoring the Git Branch Source UI / UX were improved. Wrapper class GitScm updated to include new UI componentes.
- Fixed errors in names of UI components
- Added new tests git-plugin
 
@reviewbybees @olivergondza what do you think?


MarkEWaite:
@fcojfernandez I'd like to be able to run these tests on my own machines.  Can you point me to instructions that explain how to configure and run the acceptance test harness?

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitScm.java
```diff
@@ -103,6 +111,16 @@ public GitScm calculateChangelog(String remote, String branch) {
     }
 
     /**
+     * Add behaviour "Use commit author in changelog"
+     *
+     * @return this, to allow function chaining
+     */
+    public GitScm commitAuthorInChangelog() {
+        CommitAuthorInChangelog behaviour = addBehaviour(CommitAuthorInChangelog.class);
```
varyvol: No need to store this in a variable then?

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitScm.java
```diff
@@ -155,6 +186,24 @@ public SparseCheckoutPaths sparseCheckout() {
...
+        return addBehaviour(AdvancedClone.class);
+    }
+
+    /**
+     * Add behaviour "Advanced checkout behaviours"
+     *
+     * @return behaviour, to access its method
+     */
+    public AdvancedCheckout advancedCheckout() {
+        return addBehaviour(AdvancedCheckout.class);
```
varyvol: For these ones, you don't want to return this?

src/test/java/plugins/GitPluginTest.java
```diff
@@ -371,6 +372,331 @@ public void inverse_strategy_to_choose_build() {
...
+                .credentials(USERNAME);
+        job.addShellStep("git log");
+        job.save();
+
+        b = job.startBuild();
+        b.shouldSucceed();
+        String console = b.getConsole();
+        assertThat(
+                console,
+                Matchers.containsRegexp(TEST_INITIAL_COMMIT, Pattern.MULTILINE)
```
varyvol: Why should this initial commit appear, since you are not commiting it?

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitScm.java
fcojfernandez: @varyvol you're right!

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitScm.java
fcojfernandez: My intention is allow to configure the "Advanced checkout" behaviour any time you need in your test to compare the different results.

src/test/java/plugins/GitPluginTest.java
fcojfernandez: All test initialise the git repository with the following method, which has commited a "foo" file:

    private GitRepo buildGitRepo() {
        GitRepo repo = new GitRepo();
        return repo;
    }

src/test/java/plugins/GitPluginTest.java
```diff
@@ -371,6 +372,331 @@ public void inverse_strategy_to_choose_build() {
...
+        try {
+            assertThat(
+                    console,
+                    Matchers.containsRegexp(TEST_COMMIT1_MESSAGE, Pattern.MULTILINE)
+            );
+            fail("Shallow clone depth is 1");
+        } catch (AssertionError e) {
+            // It is expected an AssertionError to be thrown
+        }
+        assertThat(
```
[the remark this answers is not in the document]
fcojfernandez: That makes sense

fcojfernandez:
@MarkEWaite you can find the full instructions [here](https://github.com/jenkinsci/acceptance-test-harness#running-tests), but to summarize, just follow these steps:

1. Set an environment variable `BROWSER`. (e.g. "firefox" or "chrome" are suitable values)
2. Set an environment variable `JENKINS_WAR` pointing the path to a jenkins war with the correct plugin version or an enviornment
3. Execute `mvn test` using those environment variables

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitScm.java
```diff
@@ -29,13 +29,21 @@
...
+import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 
 @Describable("Git")
 public class GitScm extends Scm {
+
     private final Control branch = control("branches/name");
     private final Control url = control("userRemoteConfigs/url");
     private final Control tool = control("gitTool");
+    private final Control repositoryBrowser = this.control(new String[]{"/"});
```
olivergondza: `control` accepts variadic argument list so there is no need to wrap arguemnts to array - especially when there only a single element. Use `control("foo")` or `control("foo", "bar")`

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitScm.java
```diff
@@ -103,6 +111,16 @@ public GitScm calculateChangelog(String remote, String branch) {
     }
 
     /**
+     * Add behaviour "Use commit author in changelog"
```
olivergondza: Do we need all these "Add behaviour" wrapper methods? To me, `addBehaviour(CommitAuthorInChangelog.class)` is readable enough to put into test case directly...

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitScm.java
```diff
@@ -190,9 +239,93 @@ public GitScm remoteName(String name) {
...
+     * @return behaviour, to access its method
+     */
+    public MergeBeforeBuild mergeBeforeBuild() {
+        return addBehaviour(MergeBeforeBuild.class);
+    }
+
+
     public <T extends Behaviour> T addBehaviour(Class<T> type) {
         control("hetero-list-add[extensions]").click();
-        return newInstance(type, this, "extensions");   // FIXME: find the last extension added
```
olivergondza: Oh, the original logic here is quite tangled. `addBehaviour` opens the menu, creates the instance of page area (that does not yet exist!) and let that instance to do the menu selection. It would probably be better if we expose the selector/caption over `@Describable` annotation so `addBehaviour` knows how to select it so we can do things in proper order: open menu, select, instantiate page area. 

With that change done, we can benefit from `Control.selectDropdownMenu`* methods that already workarounds the problem with menu scrolling.

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitScm.java
```diff
@@ -341,15 +518,40 @@ public GitLfsPull(GitScm git, String path) {
...
 
     public static class MergeBeforeBuild extends Behaviour {
-        private final Control txtMergeRemote = control("mergeRemote");
-        private final Control txtMergeTarget = control("mergeTarget");
-        private final Control selMergeStrategy = control("mergeStrategy");
-        private final Control selFastForwardMode = control("fastForwardMode");
+        private final Control txtMergeRemote = control("options/mergeRemote");
+        private final Control txtMergeTarget = control("options/mergeTarget");
+        private final Control selMergeStrategy = control("options/mergeStrategy");
+        private final Control selFastForwardMode = control("options/fastForwardMode");
```
olivergondza: Will this work for older git plugin releases? If not, please use both paths - newest first - to preserve the compatibility.

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitScm.java
```diff
@@ -29,13 +29,21 @@
...
 @Describable("Git")
 public class GitScm extends Scm {
+
     private final Control branch = control("branches/name");
     private final Control url = control("userRemoteConfigs/url");
     private final Control tool = control("gitTool");
+    private final Control repositoryBrowser = this.control(new String[]{"/"});
+    private final Control urlRepositoryBrowser = this.control(new String[]{"browser/repoUrl"});
+    // UI components for additional configuration for some repository browsers
+    private final Control projectName = this.control(new String[]{"browser/projectName"}); // gitlib & viewgit
```
olivergondza: These should be factored out to SCM browser abstraction so `GitScm` API si not polluted by gitlab specific controls.

src/test/java/plugins/GitPluginTest.java
```diff
@@ -371,6 +372,299 @@ public void inverse_strategy_to_choose_build() {
...
+        job.configure();
+        GitScm.AdvancedClone behaviour = gitScm.advancedClone()
+                .checkShallowClone(true)
+                .setNumDepth("2");
+        job.save();
+
+        b = job.startBuild();
+        b.shouldSucceed();
+        console = b.getConsole();
+        assertThatFail(console, TEST_INITIAL_COMMIT, "Shallow clone depth is 2");
```
olivergondza: `assertThat(console, not(ANY_MATCHER_HERE))`?

fcojfernandez: Changed

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitScm.java
fcojfernandez: In that case is not necessary to preserve compatibility. I did test with version 2.5.X (first time this behaviour is included), 2.3.2 and 2.6.0. The correct path has been always `options/XXX`.

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitScm.java
fcojfernandez: Thanks for the suggestions. I didn't know this behaviour using `@Describable`.

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitScm.java
fcojfernandez: Changed

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitScm.java
fcojfernandez: I have refactor these controls and added new sub-classes to wrap this specifics UI components.

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitScm.java
fcojfernandez: I agree with you about `addBehaviour` method is readable enough. Since some versions ago, the UI has suffered many changes and they could still change more. If we use the `addBehaviour` method and those behaviours change, all the test involved will have to be updated. I'm trying to isolate as much as possible test from all these possible changes since class `GitScm` will have to be updated in that case. It could be helpful in the future (IMHO)

fcojfernandez:
After analyse the tests that have failed, none of them are related to the changes committed.

@olivergondza, what do you think about new commits?

PR #374, opened 2017-11-08
Avoiding use of workflow-aggregator

jglick:
Following up #366 etc. `workflow-aggregator` pulls in a lot of dependencies which are not actually being tested and which we do not care about; this plugin should really just be deprecated. And especially pending https://github.com/jenkinsci/acceptance-test-harness/pull/207#issuecomment-342691560, these deps add a lot to execution time.

@reviewbybees

### DIFF
--- a/src/test/java/plugins/WorkflowPluginTest.java
+++ b/src/test/java/plugins/WorkflowPluginTest.java
@@ -67,9 +67,6 @@ import static org.jenkinsci.test.acceptance.Matchers.*;
 import static org.junit.Assert.*;
 import static org.junit.Assume.*;
 
-/**
- * Roughly follows <a href="https://github.com/jenkinsci/workflow-plugin/blob/master/TUTORIAL.md">the tutorial</a>.
- */
 public class WorkflowPluginTest extends AbstractJUnitTest {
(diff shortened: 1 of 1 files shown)


src/test/java/plugins/WorkflowPluginTest.java
```diff
@@ -83,8 +80,11 @@
     @Inject DockerContainerHolder<DockerAgentContainer> agent;
     @Inject JenkinsController controller;
 
-    @Test @WithPlugins({"workflow-aggregator", "git"}) @WithDocker
+    @Category(DockerTest.class)
```
jglick: The `Category` was missing before.

src/test/java/plugins/WorkflowPluginTest.java
```diff
@@ -106,20 +106,7 @@ private String createGitRepositoryInDockerContainer() {
         return container.getRepoUrl();
     }
 
-    @WithPlugins("workflow-aggregator@1.1")
-    @Test public void hello_World() {
```
jglick: Just deleting this. Does not test anything not covered by other tests.

jglick:
@reviewbybees done

PR #376, opened 2017-11-08
Increasing timeout for waitForInstallationToComplete

jglick:
As noted in https://github.com/jenkinsci/acceptance-test-harness/pull/207#issuecomment-342691560 the download of transitive dependencies can be quite slow, several minutes, so imposing a 60s timeout just makes tests fail when you run on a slowish network.

@reviewbybees


alecharp:
LGTM :bee:

PR #379, opened 2017-11-13
MockUpdateCenter

jglick:
Implements the notion I had in #207, and supersedes that PR. This seems to run a _lot_ faster: there is a single update center click which processes all changes, all dependencies (incl. implicit ones like `bouncycastle-api-2.16.2.hpi`) get served from the local Maven repository rather than the network, and there is no bogus restart.

- [X] serve a mock update center, with `uploadPlugins=false`
- [X] redirect other resources like tool metadata
- [x] allow plugin metadata other than `version` to be overridden
- [X] smoke test: `WorkflowPluginTest.linearFlow`
- [x] test with `LOCAL_JARS` (or similar metadata decorators)
- [x] test with an OEM version of ATH
- [x] consider cleaning up obsolete code, including handling of `uploadPlugins`, and hacks such as #355
- [x] check impact on unusual `JenkinsController`s
- [x] consider caching tool installer metadata locally

Probably supersedes or clashes with #188 & #227.

@reviewbybees @jenkinsci/code-reviewers

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
@@ -35,6 +35,7 @@ import static org.apache.http.entity.ContentType.APPLICATION_OCTET_STREAM;
 import org.apache.http.entity.mime.MultipartEntityBuilder;
 import org.apache.http.impl.client.DefaultHttpClient;
 import org.eclipse.aether.resolution.ArtifactResolutionException;
+import org.jenkinsci.test.acceptance.update_center.MockUpdateCenter;
 
 
 /**
@@ -69,6 +70,9 @@ public class PluginManager extends ContainerPageObject {
     @Named("forceRestartAfterPluginInstallation")
     public boolean forceRestart;
 
+    @Inject
+    public MockUpdateCenter mockUpdateCenter;
+
     public PluginManager(Jenkins jenkins) {
         super(jenkins.injector, jenkins.url("pluginManager/"));
         this.jenkins = jenkins;
@@ -76,13 +80,14 @@ public class PluginManager extends ContainerPageObject {
         // injection happens in the base class, so for us to differentiate default state vs false state,
         // we need to use Boolean
         if (uploadPlugins==null)
-            uploadPlugins = true;
+            uploadPlugins = false;
     }
 
     /**
      * Force update the plugin update center metadata.
      */
     public void checkForUpdates() {
+        mockUpdateCenter.ensureRunning();
         visit("advanced");
         final String current = getCurrentUrl();
         // The check now button is a form submit (POST) with a redirect to the same page only if the check is successful.
--- a/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
@@ -61,10 +61,12 @@ public class PluginManager extends ContainerPageObject {
      * Optional configuration value that selects whether to resolve plugins locally and upload to Jenkins
      * (better performing when Jenkins is closer to the test execution), or install plugins from within Jenkins
      * (more accurate testing.)
+     * @deprecated Blocks use of {@link MockUpdateCenter}.
      */
     @Inject(optional = true)
     @Named("uploadPlugins")
-    public Boolean uploadPlugins;
+    @Deprecated
+    public boolean uploadPlugins;
 
     @Inject(optional = true)
     @Named("forceRestartAfterPluginInstallation")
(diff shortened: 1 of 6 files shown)


olivergondza:
This looks good to me. Provided this supports all metadata overrides, executes Jenkins UI and dependency resolution, is there a reason to preserve the other means of installation (`uploadPlugins`)?

jglick:
> this supports all metadata overrides

As far as I know. At least I tested with `LOCAL_JARS` and it worked as expected. (In fact better than before, since the #355 hack becomes unnecessary.)

> executes Jenkins UI and dependency resolution

Yes, once the mock UC has been configured, the actual installation consists simply of going to `/pluginManaged/available`, clicking some checkboxes, and asking Jenkins to install. Which will generally require no restart, and is very fast, assuming all required plugins are in fact in the local Maven repository cache.

> is there a reason to preserve the other means of installation (`uploadPlugins`)?

For now, yes. I tested this patch against two kinds of OEM extensions:

* Those which simply create more tests but do not make any interesting changes to infrastructure otherwise. Works fine.
* Those which deal with custom WARs which set their own dedicated update center. In the case I tested, Jenkins actually gets _two_ update sites, which triggers the `found an unexpected number of update sites` escape hatch, and you are back to using the stock update center, unless you want to configure `updatePlugins` enabled.

As noted in TODO comments, I think the best approach would be to extend this fix to deprecate `CachedUpdateCenterMetadataLoader.url` and directly handle even installations with multiple custom update sites by proxying to them all. But that would require further development and testing so I felt this could be left to a later step.

jglick:
Hmm, [this](https://ci.jenkins.io/job/Core/job/acceptance-test-harness/job/PR-379/3/testReport/plugins/PostBuildScriptPluginTest/skip_for_jobs_that_failed/) looks wrong:

```
Unable to locate By.xpath: //input[starts-with(@name,'plugin.postbuildscript.')] in http://127.0.0.1:65219/pluginManager/available
…
	at org.jenkinsci.test.acceptance.po.CapybaraPortingLayerImpl.find(CapybaraPortingLayerImpl.java:221)
	at org.jenkinsci.test.acceptance.po.PluginManager.tickPluginToInstall(PluginManager.java:269)
	at org.jenkinsci.test.acceptance.po.PluginManager.installPlugins(PluginManager.java:236)
	at org.jenkinsci.test.acceptance.junit.WithPlugins$RuleImpl$1.installPlugins(WithPlugins.java:177)
	at org.jenkinsci.test.acceptance.junit.WithPlugins$RuleImpl$1.evaluate(WithPlugins.java:130)
	at …
```

dwnusbaum:
I tried to reproduce that error locally but was not able to do so. There are some changes to button text in `postbuildscript:1.1.0` that break its test harness, but after specifying version 0.18 the test passed for me multiple times. 

I was able to reproduce the same error for the [Scriptler plugin](https://wiki.jenkins.io/display/JENKINS/Scriptler+Plugin) [tests](https://ci.jenkins.io/job/Core/job/acceptance-test-harness/job/PR-379/3/testReport/plugins/ScriptlerPluginTest/):

    Unable to locate By.xpath: //input[starts-with(@name,'plugin.scriptler.')] in http://127.0.0.1:54085/pluginManager/available

Since that plugin's distribution has been suspended I don't think it is an error with `MockUpdateCenter`.

jglick:
Not sure, cannot reproduce that error locally—the plugin gets installed fine. The test suite _does_ fail, but for unrelated reasons.

jglick:
Sorry, crossed wires—my last comment was in reference to `PostBuildScriptPluginTest`.

> Since that plugin's distribution has been suspended I don't think it is an error with `MockUpdateCenter`.

Or rather: this patch probably _causes_ `ScriptlerPluginTest` to fail, but rightly so—if actual users would not be able to install it from the update center, then by definition it is not “working”.

jglick:
@reviewbybees done

src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
```diff
@@ -183,8 +185,8 @@ public boolean installPlugins(final PluginSpec... specs) throws UnableToResolveD
             checkForUpdates();
         }
 
-        LOGGER.info("Installing plugins by direct upload: " + uploadPlugins);
         if (uploadPlugins) {
```
andresrc: It would be interesting to have a way to have both `uploadPlugins=false` and have the MUC disabled. Maybe an additional flag that is checked at the beginning of `mockUpdateCenter.ensureRunning()`

olivergondza: @andresrc, Why?

jglick: I think @andresrc had in mind tests which specifically check behavior of the actual update center.

PR #380, opened 2017-11-14
Execute ATH in parallel.

lvotypko:
Attempt to execute in parallel. I will do modification according result of execution this pull request.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,27 +1,27 @@
 // For ci.jenkins.io
 // https://github.com/jenkins-infra/documentation/blob/master/ci.adoc
 
-pipeline {
-    agent {
-        dockerfile {
-            dir 'src/main/resources/ath-container'
-            args '-v /var/run/docker.sock:/var/run/docker.sock'
-            label 'docker && highmem'
-        }
-    }
-    stages {
-        stage('Run ATH') {
-            steps {
-                sh '''
-                    eval $(./vnc.sh)
-                    ./run.sh firefox latest -Dmaven.test.failure.ignore=true -DforkCount=1 -B
-                    '''
+    def splits = splitTests count(10)
+    def branches = [:]
+    for (int i = 0; i < splits.size(); i++) {
+        int index = i;
+        branches["split${i}"] = {
+            stage("Run ATH - split${index}") {
+                node('docker && highmem'){
+                    checkout scm
+                    def image = docker.build('jenkins/ath', 'src/main/resources/ath-container')
+                    image.inside('-v /var/run/docker.sock:/var/run/docker.sock') {
+                        def exclusions = splits.get(index).join("\n");
+                        writeFile file: 'excludes.txt', text: exclusions
+                        sh '''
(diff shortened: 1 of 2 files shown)


jglick:
Unfortunately hard to “bootstrap” pending [JENKINS-47206](https://issues.jenkins-ci.org/browse/JENKINS-47206) or similar.

lvotypko:
Thank you, I fully agree - especially in case of ATH. I want to have successful build first before I ask for merging (and put it all in one commit - it is not necessary to have there all my way to the right script). I will add my voice for merging your pull request and add changes to mine version of JenkinsFile.

lvotypko:
3 hr 26 min, it seems working well. I will rebase it into one commit and solve conflicts. Resolving JENKINS-47206 would help a lot too.

jglick:
Ping me in IRC or whatever if you need help resolving the merge conflict. The `milestone` calls should be done as early as possible (prior to `node`); `realtimeJUnit` replaces `junit` and just wraps whatever is actually producing test results.

lvotypko:
@jglick 
I think I merged changes correctly, the build failed because of another reason (something wrong during collection of results from split0). Can you check it?

jglick:
The failure should have been fixed in #385—try merging with `master`.

Jenkinsfile
```diff
@@ -1,27 +1,31 @@
...
-            label 'docker && highmem'
-        }
+
+    for (int i = 0; i < (BUILD_NUMBER as int); i++) {
+      milestone()
     }
-    stages {
-        stage('Run ATH') {
-            steps {
-                script { // TODO pending proper solution for JENKINS-43353
```
jglick: recommend keeping the `TODO` comment there, since it is pretty cryptic

Jenkinsfile
```diff
@@ -1,27 +1,31 @@
...
-                    sh '''
-                        eval $(./vnc.sh)
-                        ./run.sh firefox latest -Dmaven.test.failure.ignore=true -DforkCount=1 -B
-                    '''
+    def splits = splitTests count(10)
+    def branches = [:]
+    for (int i = 0; i < splits.size(); i++) {
+        int index = i;
+        branches["split${i}"] = {
+            stage("Run ATH - split${index}") {
```
jglick: For now at least, use a more identifier-like `stage` name, since it makes its way into URLs when reporting test results using newer versions of `junit`. CC @abayer—would be better IMO to use the branch name than a stage label.

Or simply drop `stage` altogether. CC @svanoort & @i386 who were discussing this kind of thing recently.

Jenkinsfile
```diff
@@ -1,27 +1,31 @@
...
+                node('docker && highmem'){
+                    checkout scm
+                    def image = docker.build('jenkins/ath', 'src/main/resources/ath-container')
+                    image.inside('-v /var/run/docker.sock:/var/run/docker.sock') {
+                        def exclusions = splits.get(index).join("\n");
```
jglick: BTW `splits[index]` works too

PR #381, opened 2017-11-15
[JENKINS-48034] Enable configuration for HttpURLConnection timeouts

fcojfernandez:
See [JENKINS-48034](https://issues.jenkins-ci.org/browse/JENKINS-48034)

**Features:**

Enable the possibility of configure HttpURLConnection timeouts via environment variable. In case of not setting a property or environment variable they remain with the default value.

@reviewbybees 
@olivergondza this is the PR associated to the discussion in the JIRA issue

PR #382, opened 2017-11-15
[INFRA-1404] Use junit-attachments & junit-realtime-test-reporter plugins

jglick:
[INFRA-1404](https://issues.jenkins-ci.org/browse/INFRA-1404)

Enabled by https://github.com/jenkins-infra/jenkins-infra/pull/898. Would be improved by having https://github.com/jenkinsci/junit-realtime-test-reporter-plugin/pull/5 merged & released. Would be a merge conflict with #380, but easy to resolve.

@reviewbybees

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,16 +12,13 @@ pipeline {
     stages {
         stage('Run ATH') {
             steps {
-                sh '''
-                    eval $(./vnc.sh)
-                    ./run.sh firefox latest -Dmaven.test.failure.ignore=true -DforkCount=1 -B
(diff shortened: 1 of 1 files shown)


jglick:
Build 2 was hard-killed by @rtyler, not sure why. Build 1 had gone for a while before being hard-killed; unclear if https://github.com/jenkinsci/junit-realtime-test-reporter-plugin/pull/5 would have helped, or if we need a more complex implementation as per JENKINS-43276. Unfortunately that patch has not been reviewed, much less released, and I have no control over timing of updates in any case, so it is hard to tune the behavior based on experience.

jglick:
Also there are very few executors capable of running these PRs; this one appears to be waiting for two builds of #381. (The use of `milestone` in _this_ PR, if merged to `master`, would let us at least only be doing one build per PR at any given time.)

PR #386, opened 2017-11-30
[JENKINS-48314] https://issues.jenkins-ci.org/browse/JENKINS-48314

scherler:
[JENKINS-48314](https://issues.jenkins-ci.org/browse/JENKINS-48314)

Drop legacy line since it is removing the default source list. 
We define ubuntu:xenial so there is no need for that line anymore.

Alternative fix is to add `xenial-security` as well to the souce.list but IMHO the whole line does not make sense anymore. @raul-arabaolaza shares this observation.

@reviewbybees @cloudbees/team-astro

### DIFF
--- a/src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/GitContainer/Dockerfile
+++ b/src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/GitContainer/Dockerfile
@@ -5,8 +5,6 @@
 
 FROM ubuntu:xenial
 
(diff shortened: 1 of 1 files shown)


raul-arabaolaza:
Yeah, I see no reason to add an already existing repo to the sources *and remove all the other ones* 

So a :bee: on my side

batmat:
@reviewbybees done

scherler:
@reviewbybees done 

@olivergondza you can merge if you want

PR #394, opened 2018-01-05
customStartup should do a more conservative match

jglick:
Amends #86 by @armfergom (which already had to be edited in #179). Needed after https://github.com/jenkinsci/ssh-slaves-plugin/pull/32.

@reviewbybees


jglick:
The whole test suite [fails anyway](https://ci.jenkins.io/job/Core/job/acceptance-test-harness/job/PR-394/1/testReport/junit/plugins/SshSlavesPluginTest/customStartup/) for reasons I cannot easily diagnose (not reproducible locally):

```
java.net.ConnectException: Connection refused (Connection refused)
```

Maybe something weird with the networking setup in the CI builder?

PR #402, opened 2018-02-02
adding more detailed logging in the event that a test rule fails to apply

burnMyDread:
I have some tests running with the withPlugins test rule and a few others. I would really like a way to tell which of my test rules is failing when the tests fail to apply all the test rules. To that end I am submitting this PR to report which test rule failed and the throwable's cause. I didn't see a contributors guide for this, so if I missed it please let me know. @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/junit/JenkinsAcceptanceTestRule.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/junit/JenkinsAcceptanceTestRule.java
@@ -112,9 +112,8 @@ public class JenkinsAcceptanceTestRule implements MethodRule { // TODO should us
                     for (TestRule rule: rulesGroup) {
                         try {
                             body = rule.apply(body, description);
-                        } catch (Exception e) { //This is to note which rule is failing when a rule does fail as the evaluate method hinds this information
-                            System.out.println("Test Rule " + rule.getClass() + " failed to apply. Failed with: " + e.getCause());
-                            throw e;
+                        } catch (Exception e) {
+                            throw new TestRuleException(e, rule);
                         }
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/test/acceptance/junit/JenkinsAcceptanceTestRule.java
```diff
@@ -110,7 +110,12 @@ public int compare(Integer o1, Integer o2) {
 
                 for (Set<TestRule> rulesGroup: rules.values()) {
                     for (TestRule rule: rulesGroup) {
-                        body = rule.apply(body, description);
+                        try {
+                            body = rule.apply(body, description);
+                        } catch (Exception e) { //This is to note which rule is failing when a rule does fail as the evaluate method hinds this information
+                            System.out.println("Test Rule " + rule.getClass() + " failed to apply. Failed with: " + e.getCause());
+                            throw e;
```
olivergondza: I would be better to wrap to another exception indicating the rule class in message as this way the stack trace and the rule name are both reported but by separate mechanisms.

burnMyDread: That was my original plan but I didn't want to change the exception type as that was what is sent to the controller in the diagnose method. I can change it to a warped exception.

burnMyDread:
@olivergondza I move to a service exception model, please take a look.

olivergondza:
Merging amended with https://github.com/jenkinsci/acceptance-test-harness/commit/8c74b90b1b03c8aa10e3a5234e189a91cd090c49. Thanks

PR #404, opened 2018-02-08
Added utility methods for asserting build status with a given timeout

agentgonzo:
Be default, the shouldXyz methods use a hard-coded 120s timeout. (line 68)

@reviewbybees


olivergondza:
Is the convenience this adds worth the method number explosion? What is wrong with `waitUntilFinished(XXX).shouldSucceed();`?

agentgonzo:
I feel it it's useful. YMMV

PR #405, opened 2018-02-08
Change DropdownMenu selection.

lvotypko:
Fix constant failing test plugins.SubversionPluginTest.build_has_changes_and_repoBrowser
New area is not created by selection now and "yui-menu-button" does not exists for browser now.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/subversion/SubversionScm.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/subversion/SubversionScm.java
@@ -56,12 +56,8 @@ public class SubversionScm extends Scm {
     }
 
     public <T extends SvnRepositoryBrowser> T useRepositoryBrowser(final Class<T> type) {
-        String path = createPageArea("configs", new Runnable() {
-            @Override public void run() {
-                repositoryBrowser.selectDropdownMenu(type);
-            }
-        });
-        return this.newInstance(type, this, this.getPage().url(path));
+        repositoryBrowser.selectDropdownMenuAlt(type);
+        return this.newInstance(type, this, this.getPage().url(getPath("[1]")));
     }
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/test/acceptance/plugins/subversion/SubversionScm.java
```diff
@@ -56,12 +56,8 @@
...
 
     public <T extends SvnRepositoryBrowser> T useRepositoryBrowser(final Class<T> type) {
-        String path = createPageArea("configs", new Runnable() {
-            @Override public void run() {
-                repositoryBrowser.selectDropdownMenu(type);
-            }
-        });
-        return this.newInstance(type, this, this.getPage().url(path));
+        repositoryBrowser.selectDropdownMenuAlt(type);
+        return this.newInstance(type, this, this.getPage().url(getPath("[1]")));
```
olivergondza: This seems like a step backwards. `createPageArea` was added to solve the problems of locating the correct page area after it was created, why do we have to drop it?

lvotypko: because createPageArea expects that new element will appear after action - but in this case it does not happen - there is not new element, the required element was there even before in as a child of selection element.

lvotypko: https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/main/java/org/jenkinsci/test/acceptance/po/PageObject.java#L153

lvotypko: little correction - it does not require new element but there has to be more elements with given path then before.

PR #406, opened 2018-02-08
Add missing plugin dependency.

lvotypko:
Fix constant failure of test FindBugsPluginTest.should_set_pipeline_result

### DIFF
--- a/src/test/java/plugins/FindBugsPluginTest.java
+++ b/src/test/java/plugins/FindBugsPluginTest.java
@@ -177,6 +177,7 @@ public class FindBugsPluginTest extends AbstractAnalysisTest<FindBugsAction> {
      * Builds a pipeline project and checks if it is unstable.
      */
     @Test
+    @WithPlugins("workflow-aggregator")
     public void should_set_pipeline_result() {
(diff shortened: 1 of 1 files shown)


src/test/java/plugins/FindBugsPluginTest.java
```diff
@@ -177,6 +177,7 @@ public void should_set_result_to_unstable_if_warning_found() {
      * Builds a pipeline project and checks if it is unstable.
      */
     @Test
+    @WithPlugins("workflow-aggregator")
```
uhafner: Is it still ok to use workflow-aggregator? Didn't @jglick suggest to not use it anymore and make the dependencies explicit? Or is this guideline just for plug-ins and not ATH?

lvotypko: Hi, yeah you are right - I totally forgot! I will change it for something with less dependencies.

lvotypko:
Finally all plugins added. I will rebase it in one commit.

uhafner:
Thanks for finding out the required dependencies. This really is quite cumbersome to find the required workflow libs. (I have the same problem in some integration tests...)

PR #407, opened 2018-02-09
[JENKINS-49470] Fixing test assertion to take into account the sanitized url

vilacides:
[JENKINS-49470](https://issues.jenkins-ci.org/browse/JENKINS-49470)

Fixing a test that has been consistently failing due to not taking into account that user.url is sanitized

@olivergondza I believe that the best fix would actually be to sanitize the url in the test as well but since the tests depend on a 1.36 jenkins version, I do not have hudson.Util available to use.

@reviewbybees

### DIFF
--- a/src/test/java/core/InstallWizardTest.java
+++ b/src/test/java/core/InstallWizardTest.java
@@ -70,7 +70,7 @@ public class InstallWizardTest extends AbstractJUnitTest {
 
         // Check that the new user is logged in
         Login login = new Login(jenkins);
-        Assert.assertThat(login, loggedInAs("adminUser"));
+        Assert.assertThat(login, loggedInAs("adminuser"));
     }
 
     @Since("2.0")
@@ -102,6 +102,6 @@ public class InstallWizardTest extends AbstractJUnitTest {
 
         // Check that the new user is logged in
         Login login = new Login(jenkins);
-        Assert.assertThat(login, loggedInAs("adminUser"));
+        Assert.assertThat(login, loggedInAs("adminuser"));
     }
 }
--- a/src/test/java/core/InstallWizardTest.java
+++ b/src/test/java/core/InstallWizardTest.java
@@ -45,6 +45,11 @@ public class InstallWizardTest extends AbstractJUnitTest {
     @Inject
     public JenkinsController controller;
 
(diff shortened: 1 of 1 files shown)


olivergondza:
Agreed the case is not relevant for the test. I believe it would be useful to change `Matchers.loggedInAs(String)` to convert the argument to lowercase...

vilacides:
@olivergondza I explained myself poorly, that is what I meant. But in order to do that just like it's done in the header I would need to have hudson.Util available to use and in order to have that I need to upgrade the jenkins version that the acceptance test harness depends on which will require many other changes. Another option would be to actually copy the method, but maintainability wise I don't think it's a good idea since the test and the code evolve separately.
So at this point I am thinking about the following options:
* Leave the change as is
* Change the username test data to not have the camel case as it's not a relevant thing to the test

AND in any case, create a followup issue to have the argument conversion in the Matcher once it's available through hudson.Utils. This will give us coverage back. What are your thoughts?

raul-arabaolaza:
FWIW I believe the ideal change will be to fix `Matchers.loggedInAs(String)` but for what I have seen I also agree with @vilacides  in that is not going to be a trivial change and not strictly needed to fix this test.

So I fully support the idea of:
* Change the test data here so it can work even with the (now known issue) in `Matchers.loggedInAs(String)`
* Create a folow up ticket to discuss and find the best way to fix `Matchers.loggedInAs(String)`

olivergondza:
Oh, so this (https://github.com/jenkinsci/jenkins/blob/f4b78921e87c0a265c633674bc3ca3ec0458c19a/core/src/main/java/hudson/model/User.java#L225) is what we need to mimic here. Got it.

Given there is the `IdStrategy` involved that permit to change the case sensitivity, I do not think we should depend on the case here _at all_. I see two variants.

- Keep things lowercase here - including the user creation (https://github.com/jenkinsci/acceptance-test-harness/pull/407/files#diff-7dcfac1efdd27354135afb76e3fb8099R67).
- Adjust `Matchers.loggedInAs(String)` so it does not depend on the case-sensitivity of the username.

No. 1 is trivial and fully sufficient so I am ok with it.

vilacides:
Confirmed, after running the tests with my changes, they pass @olivergondza

PR #413, opened 2018-02-13
Avoid JENKINS-49026.

lvotypko:
There is a problem with optional dependencies deployed dynamically.

### DIFF
--- a/src/test/java/plugins/AbstractAnalysisTest.java
+++ b/src/test/java/plugins/AbstractAnalysisTest.java
@@ -149,7 +149,7 @@ public abstract class AbstractAnalysisTest<P extends AnalysisAction> extends Abs
     @Test @Issue("JENKINS-25501") @WithPlugins("email-ext")
     public void should_send_mail_with_expanded_tokens() {
         //avoid JENKINS-49026
-        jenkins.restart();
+        checkExtensionAreDeployed(null,"hudson.plugins.analysis.tokens.AbstractTokenMacro");
 
         setUpMailer();
 
@@ -169,6 +169,17 @@ public abstract class AbstractAnalysisTest<P extends AnalysisAction> extends Abs
 
         verifyReceivedMail(String.format("%s: FAILURE", title),
                 String.format("%s: %d-0-%d", title, getNumberOfWarnings(), getNumberOfWarnings()));
+    }
+
+
+    /**
+     * Restart Jenkins if there is not required extension - see JENKINS-49026
+     */
+    protected void checkExtensionAreDeployed(String extensionName, String extensionPointName){
+        String result = jenkins.runScript("Jenkins.instance.getExtensionList(" + extensionPointName + ".class)");
+        if(result.contains("[]") || (extensionName!=null && result.contains(extensionName))){
+            jenkins.restart();
+        }
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/test/java/plugins/AbstractAnalysisTest.java
```diff
@@ -148,6 +148,9 @@ protected DumbSlave createDockerAgent() {
     // TODO: we should have two builds so that the numbers are different
     @Test @Issue("JENKINS-25501") @WithPlugins("email-ext")
     public void should_send_mail_with_expanded_tokens() {
+        //avoid JENKINS-49026
+        jenkins.restart();
+
```
uhafner: Restarting Jenkins just because our Plugin Installation is broken seems to be not the correct approach: it slows down tests unnecessarily. E.g., I don't start the tests from an empty installation, I use a populated PLUGINS_DIR. In this case the restart is totally superfluous. 

Is there a way to check if the restart is required?

lvotypko: Yes I agree, but this can not be fixed in ATH - it is a jenkins core problem, plus this problem causes failure of several plugins - it is not one bug one failure. What about to create test especially for it? I am able to reproduce it now very easily, but I am not sure if it will not change by time - plugin dependencies are changing by the time, maybe creation of special plugin for it.... Any idea? Now I think that it is better to avoid this problem because purpose of these tests which are affected is not to test this. They should test something else and they cant now.

lvotypko: of course the best solution is to fix it in core, but I do not think that it will be done in short time - it is painful bug for ATH tests but not for user of Jenkins. It is easy to restart jenkins extension points implementation of optional dependencies are there because it is not the purpose of plugin - it is only nice thing to have. Some users are not even aware of missing them. This bug is there for a very long time and only one user complained about it (I think 3 years ago).

lvotypko: Any idea? Should I try to write test for it in PluginManagerTest? Any suggestions? Thank you!

olivergondza: I guess a quick and dirty solution may be making the restart only happen when the plugin - that fails to deploy dynamically - is missing.

src/test/java/plugins/AbstractAnalysisTest.java
```diff
@@ -171,6 +171,17 @@ public void should_send_mail_with_expanded_tokens() {
                 String.format("%s: %d-0-%d", title, getNumberOfWarnings(), getNumberOfWarnings()));
     }
 
+
+    /**
+     * Restart Jenkins if there is not required extension - see JENKINS-49026
+     */
+    protected void checkExtensionAreDeployed(String extensionName, String extensionPointName){
+        String result = jenkins.runScript("Jenkins.instance.getExtensionList(" + extensionPointName + ".class)");
+        if(result.contains("[]") || (extensionName!=null && result.contains(extensionName))){
```
uhafner: Shouldn't this be `!result.contains(extensionName)`

src/test/java/plugins/AbstractAnalysisTest.java
```diff
@@ -171,6 +171,17 @@ public void should_send_mail_with_expanded_tokens() {
                 String.format("%s: %d-0-%d", title, getNumberOfWarnings(), getNumberOfWarnings()));
     }
 
+
+    /**
+     * Restart Jenkins if there is not required extension - see JENKINS-49026
+     */
+    protected void checkExtensionAreDeployed(String extensionName, String extensionPointName){
```
uhafner: Add another overloaded method that does not use the first parameter? Using `null` as valid parameter is not a good programming style.

src/test/java/plugins/AbstractAnalysisTest.java
```diff
@@ -171,6 +171,17 @@ public void should_send_mail_with_expanded_tokens() {
...
+        String result = jenkins.runScript("Jenkins.instance.getExtensionList(" + extensionPointName + ".class)");
```
uhafner: Is ".class" required? When I run thin in the script console I get an exception!

lvotypko: it works both for me, so no problem to change it, if it causes problem with .class.

olivergondza:
@uhafner, are you ok with the changes?

PR #415, opened 2018-02-15
[INFRA-1489] Do not overwrite home folder when running container

raul-arabaolaza:
* The previous code was overwriting the /home/ath-user folder with the contents
of the ATH folder, resulting in .vnc folder, among other things probably,
being deleted and hence the container unable to run the ATH unless an
external browser was used

@reviewbybees


richbg:
:honeybee: lgtm

PR #416, opened 2018-02-16
[JENKINS-49599] Fix last-page.html and heap-dump.log junit attachement by moving them into a Junit Rule

romainverduci:
Move last-page.html and heap-dump.log generation to DiagnosticRule so they are taken into account by FailureDiagnostics when adding junit attachments
Adding screenshot in case of failure (screen recording is not available in headless mode so having a screenshot is better than nothing)

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/junit/DiagnosticRule.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/junit/DiagnosticRule.java
@@ -1,0 +1,72 @@
+package org.jenkinsci.test.acceptance.junit;
+
+import java.io.File;
+import java.io.IOException;
+import javax.inject.Inject;
+import org.apache.commons.io.FileUtils;
+import org.jenkinsci.test.acceptance.controller.JenkinsController;
+import org.jenkinsci.test.acceptance.po.CapybaraPortingLayerImpl;
+import org.junit.rules.TestWatcher;
+import org.junit.runner.Description;
+import org.openqa.selenium.NoSuchElementException;
+import org.openqa.selenium.OutputType;
+import org.openqa.selenium.TakesScreenshot;
+import org.openqa.selenium.WebDriver;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@GlobalRule
+public class DiagnosticRule extends TestWatcher {
+    private static final Logger logger = LoggerFactory.getLogger(DiagnosticRule.class);
+
+    @Inject
+    FailureDiagnostics diagnostics;
+    @Inject
+    JenkinsController controller;
+    @Inject
+    WebDriver driver;
+
+    @Override
+    protected void failed(Throwable t, Description description) {
+        takeScreenshot();
+
+        if (causedBy(t, NoSuchElementException.class)) {
+            writeHtmlPage();
+        }
+
+        try {
+            controller.diagnose(t);
+        } catch (IOException e) {
+            throw new Error(e);
+        }
+
+    }
+
+    private void takeScreenshot() {
+        try {
+            File file = diagnostics.touch("screenshot.png");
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/test/acceptance/junit/DiagnosticRule.java
```diff
@@ -0,0 +1,72 @@
...
+            controller.diagnose(t);
+        } catch (IOException e) {
+            throw new Error(e);
+        }
+
+    }
+
+    private void takeScreenshot() {
+        try {
+            File file = diagnostics.touch("screenshot.png");
```
ryanwren: If multiple tests fail, won't the screenshots overwrite each other? This seems to be writing to the same path.

If so, you can pass in the description as an argument when invoking this method in the failed method above. Then, include the class and method name from the description, into the filename. That will avoid overwriting the same file.

src/main/java/org/jenkinsci/test/acceptance/junit/JenkinsAcceptanceTestRule.java
```diff
@@ -62,15 +60,6 @@ public void evaluate() throws Throwable {
                     System.out.printf("Skipping %s%n", description.getDisplayName());
                     e.printStackTrace();
                     throw e;
-                } catch (Exception|AssertionError e) { // Errors and failures
```
ryanwren: After moving this logic out of this method into the failed method in the DiagnosticRule class - will screenshots still be taken on Assertion Errors? Or, only on Exceptions? I guess this is determined by what conditions causes the failed method to be invoked.

src/main/java/org/jenkinsci/test/acceptance/junit/DiagnosticRule.java
romainverduci: the touch() method is putting the file in a directory with the name of the test, as we take only one screenshot by failing test it should be ok.

ryanwren: @romainverduci okay perfect

src/main/java/org/jenkinsci/test/acceptance/junit/JenkinsAcceptanceTestRule.java
romainverduci: Just checked to be sure, the failed() method is called for both exceptions and AssertionError.

ryanwren: @romainverduci awesome

recena:
@romainverduci Could you attach an example of the screenshot?

romainverduci:
I update core.CreateItemTest#duplicate_item_name_displays_error to make it fail.
`        assertThat(driver, not(hasContent(JOB_CREATION_ERROR_MSG)));`
Basically the screenshot is the last state of the browser after the test failure.
On this was the test expect to not see the JOB_CREATION_ERROR_MSG, on the screenshot we can see that the message is there.
screenshot

vilacides:
LGTM @romainverduci

raul-arabaolaza:
:bee: Thanks for this @romainverduci

recena:
@romainverduci Thanks. I think the screenshot will be helpful only in particular cases where JS is not participating (dynamic elements, etc...).

🐝

olivergondza:
@recena, it is cheap to create and the heuristics to tell that it is needed or not can get wrong easily so I am fine taking that even if that might not be relevant.

recena:
@olivergondza What do you mean?

olivergondza:
I meant it as a response to your message[1]. I do not think it is a big deal to take it even when not really needed...

[1]
> I think the screenshot will be helpful only in particular cases where JS is not participating (dynamic elements, etc...).

recena:
@olivergodza if the user of this feature is aware about the limitations, is fine.

PR #418, opened 2018-02-26
Set SHARED_DOCKER_SERVICE=true for running ath inside docker container.

lvotypko:
SHARED_DOCKER_SERVICE=true should be set. Sending pull request and see the results in upstream Jenkins run.


lvotypko:
Result is good. There are several failures because of copied file access (between containers) but network communication seems fixed.

lvotypko:
I will fix the problem with file sharing/copying (between two containers) in separate pull request.

PR #419, opened 2018-03-02
[JENKINS-49841] Explicitely wait for the configure form to be loaded instead of waiting 1s

romainverduci:
Replace implicit wait of 1s by an explicit wait for the configure form to be present on the page to assert that the configure page is properly loaded.

Jira issue: https://issues.jenkins-ci.org/browse/JENKINS-49841

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/ConfigurablePageObject.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/ConfigurablePageObject.java
@@ -29,6 +29,7 @@ import java.util.concurrent.Callable;
 import com.google.inject.Injector;
 
 import groovy.lang.Closure;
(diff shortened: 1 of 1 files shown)


raul-arabaolaza:
@romainverduci AFAIUI this is waiting for the config form to be there (instead of waiting a fixed amount of time) but there are some js generated things that get rendered after the form is there so I am not sure if it would be better to check for something different like the save button to be clickable or something like that. WDYT?

romainverduci:
@raul-arabaolaza good point I will check that.

romainverduci:
@raul-arabaolaza I added a wait for the Save button, what do you think about having both ?
I wanted to use waitForCond method to wait to the element to be clickable but it's deprecated and probably waiting for the element to be visible is enough.

PR #424, opened 2018-03-15
Error message was changed.

lvotypko:


src/test/java/core/FormValidationTest.java
```diff
@@ -43,6 +43,6 @@ public void validate() throws Exception {
 
         c.numExecutors.set(-16);
         formValidation = c.numExecutors.getFormValidation();
-        assertThat(formValidation, reports(Kind.ERROR, "Not an integer"));
+        assertThat(formValidation, reports(Kind.ERROR, "Not a non-negative number"));
```
uhafner: When did it change? Shouldn't we provide support for LTS and latest? 

@olivergondza is there no LTS job available any more?

olivergondza: There is[1], though it is not running for ATH PRs IIRC. And yes, we should retain the compatibility here with comment of which version had the change so we know when to drop it.

[1] https://jenkins.ci.cloudbees.com/job/core/job/acceptance-test-harness-stable/

lvotypko: I will found where exactly was the change done (I mean commit) so we can decide if there is a need to count with both. Thank you!

lvotypko: this change is since jenkins 2.205. It was changed by this pull request https://github.com/jenkinsci/jenkins/pull/3141/. 
I will do change to support both now.

PR #425, opened 2018-03-16
Assure that user is logged.

lvotypko:
Visiting unprotected root action as the first url causes that kerberos authentication is not done. Assure first that user is logged in on the main page and after do check by "whoAmI".

### DIFF
--- a/src/test/java/plugins/KerberosSsoTest.java
+++ b/src/test/java/plugins/KerberosSsoTest.java
@@ -47,7 +47,9 @@ import org.jenkinsci.test.acceptance.po.GlobalPluginConfiguration;
 import org.jenkinsci.test.acceptance.po.GlobalSecurityConfig;
 import org.jenkinsci.test.acceptance.po.JenkinsConfig;
 import org.jenkinsci.test.acceptance.po.JenkinsDatabaseSecurityRealm;
+import org.jenkinsci.test.acceptance.po.Login;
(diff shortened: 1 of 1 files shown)


src/test/java/plugins/KerberosSsoTest.java
```diff
@@ -116,6 +119,11 @@ private void verifyTicketAuth(KerberosContainer kdc) throws IOException, Interru
 
         // Correctly negotiate in browser
         FirefoxDriver negotiatingDriver = getNegotiatingFirefox(kdc, tokenCache);
+
+        //assure first that user is logged before visiting unprotected root action
+        Assert.assertThat(new Login(jenkins), loggedInAs("user"));
```
olivergondza: This line seems wrong and confusing:

- It is asserting the user is logged in in the "default" browser - not the negotiating one.
- Accessing the `login` page triggers negotiation in "anonymous access" mode but that is not what we are testing here.

The fix, as I understand it, is happening on next line by accessing the dashboard (url without special treatment by the plugin). I do not think we need this line.

lvotypko: oh, you are right - this will fail! That line should be after visiting dashboard. It is not necessary. I put it there to have clear why to visit the dashboard first and to fail in case that user is not logged in (it is confusing to visit one url without check and immediately the next one). But comment can explain it as well.

PR #429, opened 2018-04-19
[INFRA-1574] Make the Jenkinsfile able to publish containers

raul-arabaolaza:
[INFRA-1574](https://issues.jenkins-ci.org/browse/INFRA-1574)

Add appropriate code to Jenkinsfile so docker images for ath container can be published automatically

@reviewbybees

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,9 +1,33 @@
 // For ci.jenkins.io
 // https://github.com/jenkins-infra/documentation/blob/master/ci.adoc
 
+def imageName = 'jenkinsciinfra/ath'
+
+def container
+/* Assuming we're not inside of a pull request or multibranch pipeline */
+if (!(env.CHANGE_ID || env.BRANCH_NAME)) {
+    node('docker') {
+        stage('Prepare Container') {
+            timestamps {
+                checkout scm
+                sh 'git rev-parse HEAD > GIT_COMMIT'
+                shortCommit = readFile('GIT_COMMIT').take(6)
+                def imageTag = "${env.BUILD_ID}-build${shortCommit}"
+                echo "Creating the container ${imageName}:${imageTag}"
+                container = docker.build("${imageName}:${imageTag}", 'src/main/resources/ath-container')
+            }
+        }
+
+        stage('Publish container') {
+            infra.withDockerCredentials {
+                timestamps { container.push() }
+            }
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
Why don't we just use DockerHub auto-build?

raul-arabaolaza:
@oleg-nenashev I am following the instructions in INFRA-1574, no idea if there are other alternatives so any suggestion is welcome

Jenkinsfile
```diff
@@ -1,17 +1,41 @@
...
     def splits = splitTests count(10)
     def branches = [:]
     for (int i = 0; i < splits.size(); i++) {
         int index = i;
         branches["split${i}"] = {
             stage("Run ATH - split${index}") {
-                node('docker && highmem'){
+                node('docker && highmem') {
                     checkout scm
                     def image = docker.build('jenkins/ath', 'src/main/resources/ath-container')
```
olblak: In order to remove confusion, It would  be better to specify a tag different than what already exist on hub.docker.com.
And avoid the default 'latest'.

olblak: Is it needed to build this docker image N times? 
Wouldn't it be better to move this operation outside the for loop?

olblak: I mean, moving the build step outside the loop and just run a specific docker image inside the loop

olblak:
@oleg-nenashev you can specify a path for the Dockerfile and the Jenkinsfile still need to be adapted, docker build step need to be replace by a docker run

oleg-nenashev:
> docker build step need to be replace by a docker run

Yup. Once there is a `Dockerfile` ready, I will get it posted as a first snapshot.
@raul-arabaolaza Do you want me to just post https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/main/resources/ath-container/Dockerfile ? Or do you want any patches to be applied before it?

raul-arabaolaza:
@oleg-nenashev The current content is fine for me, WDYT @olivergondza ?

olivergondza:
@raul-arabaolaza, As I understand https://issues.jenkins-ci.org/browse/INFRA-1574 we agreed not to mix this *test execution* `Jenkinsfile` with image publishing logic for several reasons.

raul-arabaolaza:
So, I am closing this to avoid confussion

PR #433, opened 2018-05-23
[JENKINS-51489] Introducing pre configured plugins mode

raul-arabaolaza:
[JENKINS-51489](https://issues.jenkins-ci.org/browse/JENKINS-51489)

See the linked ticket, when running against an existing environment with a preconfigured set of plugins, for example, a war file generated from [CWP](https://github.com/jenkinsci/custom-war-packager)  or an existing instance seems reasonable for the ATH to not try to install/update any plugins but only validate the existing ones to check if they are enough for the set of tests to run.

This PR introduces a new `pre-configured plugins mode` which does exactly that if you specify a `pluginEvaluationOutcome` property via the groovy wiring script the mode is activated, depending on the value (see the new docs) a test whose dependencies are not satisfied may directly fail or be skipped without the ATH trying to install/update any plugin

Note that this new mode is opt-in so no change in default behaviour is introduced 

@reviewbybees @oleg-nenashev @jglick 

cc @olivergondza

### DIFF
--- a/docs/SUT-VERSIONS.md
+++ b/docs/SUT-VERSIONS.md
@@ -58,16 +58,13 @@ executed during your tests).
 Note that this option is not yet supported by all [Jenkins controllers](CONTROLLER.md), 
 so e.g. use `TYPE=winstone` in order to get the `PLUGINS_DIR` option working.
 
-### Running in pre configured plugin mode
-
-In some cases you may be running the ATH against a pre configured instance, for example you can provide a war file that
-somehow, for example by being created by the [CWP](https://github.com/jenkinsci/custom-war-packager),
-contains an already pre configured set of plugins, or against an existing environment by using some specific Controller.
-In this case you may want to ATH to not manage plugins installations at all but only validate that the existing configured
+In some cases, when you are providing the full plugins colection, you are running the ATH against a war with a pre configured
+set of plugins (for example by being created by the [CWP](https://github.com/jenkinsci/custom-war-packager))
+or running against an already existing environment, you may want the ATH to not manage plugins at all but only validate that the existing configured
 plugins are enough to run the tests.
 
-You can activate this mode with the property `pluginEvaluationOutcome`, possible values for this property are:
+You can activate this `pre configured plugins` mode with the System property `pluginEvaluationOutcome`, possible values for this property are:
 * `"skipOnInvalid"` Which means the test will be skipped if the installed plugins are not enough to cover the test requisites
 * `"failOnInvalid"` Which means the test will fail if the installed plugins are not enough to cover the test requisites
 
-This property has to be provided via [the groovy wiring script](WIRING.md) and this mode is disabled if the property is not specified
+Note this mode is disabled if the property is not specified
--- a/docs/SUT-VERSIONS.md
+++ b/docs/SUT-VERSIONS.md
@@ -60,11 +60,13 @@ so e.g. use `TYPE=winstone` in order to get the `PLUGINS_DIR` option working.
 
 ### Running in pre configured plugin mode
 
-You can provide a war file that somehow, for example by being created by the [CWP](https://github.com/jenkinsci/custom-war-packager),
-contains an already pre configured set of plugins, in this case you may want to ATH to not manage plugins installations at all
-but only validate that the existing configured plugins are enough to run the tests.
+In some cases you may be running the ATH against a pre configured instance, for example you can provide a war file that
+somehow, for example by being created by the [CWP](https://github.com/jenkinsci/custom-war-packager),
+contains an already pre configured set of plugins, or against an existing environment by using some specific Controller.
+In this case you may want to ATH to not manage plugins installations at all but only validate that the existing configured
+plugins are enough to run the tests.
 
-You can activate this mode by adding a CONFIG file with the property `pluginEvaluationOutcome`, possible values for this property are:
+You can activate this mode with the property `pluginEvaluationOutcome`, possible values for this property are:
 * `"skipOnInvalid"` Which means the test will be skipped if the installed plugins are not enough to cover the test requisites
 * `"failOnInvalid"` Which means the test will fail if the installed plugins are not enough to cover the test requisites
 
--- a/docs/SUT-VERSIONS.md
+++ b/docs/SUT-VERSIONS.md
@@ -57,3 +57,15 @@ executed during your tests).
 
 Note that this option is not yet supported by all [Jenkins controllers](CONTROLLER.md), 
 so e.g. use `TYPE=winstone` in order to get the `PLUGINS_DIR` option working.
+
+### Running in pre configured plugin mode
+
+You can provide a war file that somehow, for example by being created by the [CWP](https://github.com/jenkinsci/custom-war-packager),
+contains an already pre configured set of plugins, in this case you may want to ATH to not manage plugins installations at all
+but only validate that the existing configured plugins are enough to run the tests.
+
+You can activate this mode by adding a CONFIG file with the property `pluginEvaluationOutcome`, possible values for this property are:
+* `"skipOnInvalid"` Which means the test will be skipped if the installed plugins are not enough to cover the test requisites
+* `"failOnInvalid"` Which means the test will fail if the installed plugins are not enough to cover the test requisites
+
+This property has to be provided via [the groovy wiring script](WIRING.md) and this mode is disabled if the property is not specified
--- a/src/main/java/org/jenkinsci/test/acceptance/junit/WithPlugins.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/junit/WithPlugins.java
@@ -45,7 +45,7 @@ import java.io.IOException;
  *
  * The latter example declares that running the test with older version is pointless, typically because of missing feature.
  *
- * The annotation guarantees that the plugin is installed in required or later version.
+ * In normal mode the annotation guarantees that the plugin is installed in required or later version.
  * If the plugin is already installed but not in correct version then
  * the environment variable NEVER_REPLACE_EXISTING_PLUGINS is evaluated:
  * <ul>
@@ -53,7 +53,21 @@ import java.io.IOException;
  *     <li>if the environment variable is undefined then the installed version
  *     of the plugin is overwritten with the latest version of the plugin. If
  *     required version is not available in update center, the test will fail.</li>
-*  </ul>
+ *  </ul>
+ *
+ * There is also a pre configured plugins mode, running in this mode means that the ATH is using a war file that (somehow)
+ * has already preconfigured all the plugins that are to be tested, in that case the ATH only validates that the
+ * pre configured plugin universe is enough to run the tests and don't try to modify the existing plugins in any way.
+ *
+ * If the existing plugin configuration is not enough to run the test the end result depends on the configured value
+ * for the configuration property pluginEvaluationOutcome
+ * <ul>
+ *     <li>failOnInvalid means the test is to fail if the plugin configuration is not valid for the test</li>
+ *     <li>skipOnInvalid means the test is to be skipped if the plugin configuration is not valid for the test</li>
+ * </ul>
+ *
+ * Pre configured plugins mode is activated if and only if pluginEvaluationOutcome has a not null value, by default is
+ * not active
  *
  * @author Kohsuke Kawaguchi
  */
@@ -74,6 +88,8 @@ public @interface WithPlugins {
     class RuleImpl implements TestRule {
 
         private static final Logger LOGGER = Logger.getLogger(WithPlugins.class.getName());
+        private static final String FAIL_ON_INVALID = "failOnInvalid";
+        private static final String SKIP_ON_INVALID = "skipOnInvalid";
 
         @Inject
         Injector injector;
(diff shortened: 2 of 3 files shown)


raul-arabaolaza:
No new failures introduced by this PR it seems

olivergondza:
Let me see if I understand this: you are interested testing a core+plugins bundle and in order to keep the testing predictable/reliable,you wish ATH not to touch any plugins. Additionally, you would like to configure whether tests should fail or be skipped in case it does not have its plugin dependencies satisfied.

I am wondering, are you after prohibiting updates, installals (of plugins not in bundle) or both?

olivergondza:
Note there is a similar feature to inject plugins into ATH execution (which I presume does not work for you), so I would like to make sure enhancements contributed here are usabe there as well: https://github.com/jenkinsci/acceptance-test-harness/blob/master/docs/SUT-VERSIONS.md#provide-full-plugins-collection

raul-arabaolaza:
> Let me see if I understand this: you are interested testing a core+plugins bundle and in order to keep the testing predictable/reliable,you wish ATH not to touch any plugins. Additionally, you would like to configure whether tests should fail or be skipped in case it does not have its plugin dependencies satisfied.

Correct, I want the ability to run the ATH with a precondition that can be described as "All needed plugins are already installed" in this case what I want the ATH to do is validate if that precondition is valid before running the tests but do nothing to meet the requeriments if not.

> I am wondering, are you after prohibiting updates, installals (of plugins not in bundle) or both?

In this case I want do prohibit both install and updates, the graph of bundled plugins has to be coherent and enough to run the tests, no new installations or updates of plugins.

> Note there is a similar feature to inject plugins into ATH execution (which I presume does not work for you), so I would like to make sure enhancements contributed here are usabe there as well: https://github.com/jenkinsci/acceptance-test-harness/blob/master/docs/SUT-VERSIONS.md#provide-full-plugins-collection

In this new mode, `LOCAL_JARS` or  `LOCAL_SNAPSHOTS` are not going to work as `WithPlugins` is modified to never use the UC to install or update plugins, however `PLUGINS_DIR` can be considered as a way to achieve a preconfigured environment. However using `PLUGINS_DIR` alone neither inhibits the ATH capabilities to install/update plugins nor validates if the current installed plugins are enough to run the desired tests.

olivergondza:
> In this new mode, LOCAL_JARS or LOCAL_SNAPSHOTS are not going to work as WithPlugins is modified to never use the UC to install or update plugins, however PLUGINS_DIR can be considered as a way to achieve a pre-configured environment. However, using PLUGINS_DIR alone neither inhibits the ATH capabilities to install/update plugins nor validates if the currently installed plugins are enough to run the desired tests.

Right, `PLUGINS_DIR` is exactly what I meant.

---

The thing is, there already is a way to prevent it to update plugins that are installed (`NEVER_REPLACE_EXISTING_PLUGINS`) and your approach is tangential to it. I find the whole business of 2 modes and 2 variables (each passed a different mechanism with separate implementation) to control plugin evaluation increasingly harder and harder to grasp. Though, I like you approach as it is lot more general than what ATH does right now.

I am thinking about 2 java properties here (I am open for suggestions of the names):

`whenPluginInstallRequired` | `whenPluginUpdateRequired`
------------------------|--------------------------
`install` | `update`
`skip` | `skip`
`fail` | `fail`

The default of `install+update` would be for upstream CI. `skip+skip`/`fail+fail` would cover your use-cases. `NEVER_REPLACE_EXISTING_PLUGINS` can be deprecated configuring the variables to `install+skip` without breaking existing users of it and leveraging your new implementation. WDYT?

raul-arabaolaza:
> The thing is, there already is a way to prevent it to update plugins that are installed (`NEVER_REPLACE_EXISTING_PLUGINS`) and your approach is tangential to it.

Not really, that property prevents to update an already existing plugin for which a new version is requested explicitly via `WithPlugins` that is, if I have a test with:

```
@Test
@WithPlugins("git-client@1.7.0")
public void whatever {
....}
```

And your environment has an older version installed the update is prevented (and the test is always skipped) but it does not prevent an update of `git-client` if you have something like:

```
@Test
@WithPlugins("git@3.9.0")
public void whatever {
....}
```

And `git3.9.0` depends on a higher version of `git-client` than the already installed.

_Well, actually no update is done regardless of the value of `neverReplaceExistingPlugins` if the plugin is bundled on the war file due to [JENKINS-36441](https://issues.jenkins-ci.org/browse/JENKINS-36441) but that is irrelevant in our use case :P_

So, the problem with your proposal as I I understand it is that allowing to install plugins but not allowing to update them really does not guarantee that no updates are performed because of a transitive dependency, and that is managed by the UC no the ATH. In this sense, I also believe the actual neverUpdatePlugins property is useless if you want to test a deterministic environment

But I am totally open to discuss  :)

olivergondza:
Good point, we seem to have broken `NEVER_REPLACE_EXISTING_PLUGINS` guarantee for transitive dependencies when we stopped with custom plugin resolution.

Alright than, since that is broken and your proposal provides more meaningful guarantees even for `PLUGINS_DIR` use-case, I suggest to discontinue `NEVER_REPLACE_EXISTING_PLUGINS` altogether.

CC @lyrch

raul-arabaolaza:
@olivergondza I do not have much experience discontinuing things in the ATH, should I just deprecate it and put a warn if used or going the full path and delete it?

docs/SUT-VERSIONS.md
```diff
@@ -56,4 +56,18 @@ as expected (and it is also a good way to speed up your tests since no slow plug
 executed during your tests).
 
 Note that this option is not yet supported by all [Jenkins controllers](CONTROLLER.md), 
-so e.g. use `TYPE=winstone` in order to get the `PLUGINS_DIR` option working.
\ No newline at end of file
+so e.g. use `TYPE=winstone` in order to get the `PLUGINS_DIR` option working.
+
+### Running in pre configured plugin mode
```
olivergondza: I suggest to squash this to previous section covering both `PLUGIN_DIR` and the bundle use-case.

src/main/java/org/jenkinsci/test/acceptance/junit/WithPlugins.java
```diff
@@ -45,15 +45,29 @@
...
  * the environment variable NEVER_REPLACE_EXISTING_PLUGINS is evaluated:
  * <ul>
  *     <li>if the environment variable is set then the test will be skipped.</li>
  *     <li>if the environment variable is undefined then the installed version
  *     of the plugin is overwritten with the latest version of the plugin. If
  *     required version is not available in update center, the test will fail.</li>
-*  </ul>
+ *  </ul>
+ *
+ * There is also a pre configured plugins mode, running in this mode means that the ATH is using a war file that (somehow)
```
olivergondza: Once we kill `NEVER_REPLACE_EXISTING_PLUGINS` this will be the "normal" mode

src/main/java/org/jenkinsci/test/acceptance/junit/WithPlugins.java
```diff
@@ -84,6 +100,9 @@
         @Inject(optional=true) @Named("neverReplaceExistingPlugins")
         boolean neverReplaceExistingPlugins;
 
+        @Inject(optional = true) @Named("pluginEvaluationOutcome")
```
olivergondza: Can this be provided by good old `-D...` properties? I would like to stay away from the config *.groovy binding insanity.

src/main/java/org/jenkinsci/test/acceptance/junit/WithPlugins.java
```diff
@@ -84,6 +100,9 @@
         @Inject(optional=true) @Named("neverReplaceExistingPlugins")
```
olivergondza: We can kill all traces of this.

src/main/java/org/jenkinsci/test/acceptance/junit/WithPlugins.java
raul-arabaolaza: If we make this the normal mode it will break all existing ATH runs out there but the essentials ones and maybe some of those using `PLUGINS_DIR`, better to make this opt-in for the moment, IMO

raul-arabaolaza:
@olivergondza Feedback addressed

src/main/java/org/jenkinsci/test/acceptance/junit/WithPlugins.java
olivergondza: Sure, what I meant is I do not like distinguishing the *modes*. It is just an option that will cause test to skip/fail under certain situation. We may think of at as modes of execution but then there is not much to say about the normal mode. I do not feel strongly here.

src/main/java/org/jenkinsci/test/acceptance/junit/WithPlugins.java
```diff
@@ -98,9 +98,6 @@
         @Inject @Named("ExercisedPluginReporter")
         ExercisedPluginsReporter pluginReporter;
 
-        @Inject(optional=true) @Named("neverReplaceExistingPlugins")
-        boolean neverReplaceExistingPlugins;
```
olivergondza: Note this is used in `FallbackConfig`'s provider

raul-arabaolaza: Shame on me, my search keyword was wrong, fixed

raul-arabaolaza:
Investigating if the failures are related

raul-arabaolaza:
Forced a push to retrigger the CI build, let's wait for the results,

raul-arabaolaza:
Giving the same results than `master`, @olivergondza I believe this can be merged now

raul-arabaolaza:
@olivergondza Ping, please :)

olivergondza:
@jglick, can we go on without you?

src/main/java/org/jenkinsci/test/acceptance/junit/WithPlugins.java
```diff
@@ -45,15 +45,29 @@
...
+ *
+ * There is also a pre configured plugins mode, running in this mode means that the ATH is using a war file that (somehow)
+ * has already preconfigured all the plugins that are to be tested, in that case the ATH only validates that the
+ * pre configured plugin universe is enough to run the tests and don't try to modify the existing plugins in any way.
+ *
+ * If the existing plugin configuration is not enough to run the test the end result depends on the configured value
+ * for the configuration property pluginEvaluationOutcome
+ * <ul>
+ *     <li>failOnInvalid means the test is to fail if the plugin configuration is not valid for the test</li>
+ *     <li>skipOnInvalid means the test is to be skipped if the plugin configuration is not valid for the test</li>
```
jglick: Already documented in `SUT-VERSIONS.md`, best to not repeat the details here.

src/main/java/org/jenkinsci/test/acceptance/junit/WithPlugins.java
```diff
@@ -182,5 +208,14 @@ private void installPlugins(List<PluginSpec> install) {
                 }
             };
         }
+
+        private void handleInvalidState(String pluginEvaluationOutcome, PluginSpec spec, Description d) {
+            final String format = String.format("%s plugin is required by test %s but it is not installed in a valid version and ATH is running in preconfigured mode", spec, d.getDisplayName());
+            if (pluginEvaluationOutcome.equals(FAIL_ON_INVALID)) {
+                throw new IllegalStateException(format);
```
jglick: `AssertionError` is the normal throwable type for test failures.

src/main/java/org/jenkinsci/test/acceptance/junit/WithPlugins.java
```diff
@@ -182,5 +208,14 @@ private void installPlugins(List<PluginSpec> install) {
...
+            } else if (pluginEvaluationOutcome.equals(SKIP_ON_INVALID)) {
+                throw new AssumptionViolatedException(format);
+            }
```
jglick:
```java
} else {
    assert false : "unrecognized pluginEvaluationOutcome=" + pluginEvaluationOutcome;
}
```

src/main/java/org/jenkinsci/test/acceptance/junit/WithPlugins.java
```diff
@@ -182,5 +208,14 @@ private void installPlugins(List<PluginSpec> install) {
                 }
             };
         }
+
+        private void handleInvalidState(String pluginEvaluationOutcome, PluginSpec spec, Description d) {
+            final String format = String.format("%s plugin is required by test %s but it is not installed in a valid version and ATH is running in preconfigured mode", spec, d.getDisplayName());
```
jglick: To assist debugging, is the actual plugin version easily available here?

src/main/java/org/jenkinsci/test/acceptance/junit/WithPlugins.java
raul-arabaolaza: Fixed

src/main/java/org/jenkinsci/test/acceptance/junit/WithPlugins.java
raul-arabaolaza: Done

raul-arabaolaza:
Addressed feedback from @jglick. Now all is approved and ready to merge :)

PR #434, opened 2018-05-23
Update check for Install Wizard completion to look for the target button by ID instead of the text label

mikecirioli:
The label for button used to complete the install wizard has changed in the UI, causing tests that
 depend on this method to fail.  The button label is now "**Save and Continue**" (formerly "**Save and Finish**") - just updating the string that is being searched for in the test means we will be at risk of hitting this problem again the next time it changes.  

Instead the check needs to be fixed to look for the button by its ID, which should be much less likely to change.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/WizardCreateAdminUser.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/WizardCreateAdminUser.java
@@ -63,7 +63,7 @@ public class WizardCreateAdminUser extends PageObject {
         }
 
         driver.switchTo().defaultContent();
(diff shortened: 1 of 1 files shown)


uhafner:
When did it change? The tests should be compatible with LTS and latest...

BTW: rather then checking the label you should use IDs.

olivergondza:
Agree with @uhafner, This will make the tests to work with new version of Jenkins breaking the old one. The fix should support both.

uhafner:
Has the button no ID?

olivergondza:
Closing as https://github.com/jenkinsci/acceptance-test-harness/pull/435 was merged.

PR #435, opened 2018-05-28
Look up buttons based on css classes rather than label

Vlatombe:
This is an alternate impl, to #434 using css classes to look up the buttons. The buttons themselves don't have IDs so we don't have anything better for now.

/cc @mikecirioli

@olivergondza @uhafner

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/WizardCreateAdminUser.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/WizardCreateAdminUser.java
@@ -29,6 +29,7 @@ import static org.jenkinsci.test.acceptance.Matchers.hasContent;
 import org.jenkinsci.test.acceptance.controller.JenkinsController;
 
 import com.google.inject.Inject;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/WizardCreateAdminUser.java
```diff
@@ -73,8 +73,9 @@ public void shouldCreateTheUserSuccessfully() {
     }
 
     public void wizardShouldFinishSuccessfully() {
-        waitFor(by.button(START_BUTTON_TEXT));
-        clickButton(START_BUTTON_TEXT);
+        By installDoneButtonSelector = by.button(".btn-primary.install-done");
```
olivergondza: Should this not be `by.css` as well?

Vlatombe: `by.button` works, but `by.css` is simpler indeed.

PR #436, opened 2018-06-04
Fix tests for cifs publisher

olivergondza:


lvotypko:
I agree, due to problems with docker cp, see [1], it is better not to use it if it is not necessary. In this case it can be easy done another way.

[1] https://github.com/jenkinsci/docker-fixtures/pull/17

olivergondza:
This is still a work in progress as this has uncovered what appears to be a real regression in the plugin.

olivergondza:
Filing the discovered issue as https://issues.jenkins-ci.org/browse/JENKINS-51852

olivergondza:
Their plugin version is fairly old...

PR #437, opened 2018-06-05
Infer some variables automatically when run from container

olivergondza:
@raul-arabaolaza


raul-arabaolaza:
LGTM, I was worried about people using Dind instead of docker on docker, but you can always override SHARED_DOCKER_SERVICE from the docker run command.

raul-arabaolaza:
BTW I am working on a way to make this image usable even if you run it as root, for CI scenarios

PR #441, opened 2018-07-05
Make HttpURLConnection timeouts configurable.

agentgonzo:
The default timeouts may not be for everyone

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/utils/IOUtil.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/utils/IOUtil.java
@@ -35,6 +35,9 @@ public class IOUtil {
 
     private final static ElasticTime time = new ElasticTime();
 
+    private static int connectTimeoutInMs = 10_000;
+    private static int readTimeoutInMs = 10_000;
+
     /**
      * Get First existing file or directory.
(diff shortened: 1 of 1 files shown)


jtnord:
This is really just because Jenkins may not start up in the default timeout.
Whilst making this configurable is OK, I think the startup timeout for Jenkins should be different to anything else.  The init method should probably use something other than `getVersion()` to check that the system is a Jenkins (and possibly poll)

```
openConnection:60, IOUtil (org.jenkinsci.test.acceptance.utils)
getVersion:70, Jenkins (org.jenkinsci.test.acceptance.po)
<init>:38, Jenkins (org.jenkinsci.test.acceptance.po)
<init>:50, Jenkins (org.jenkinsci.test.acceptance.po)
createJenkins:261, FallbackConfig (org.jenkinsci.test.acceptance)
```

agentgonzo:
@jtnord - the startup timeout has been subsequently done in #442. I'll leave this here because I don't think it makes things any worse

olivergondza:
I prefer to do this through a system property. I do not like the aspect how one test can change this and modify the environment for other tests.

agentgonzo:
Sounds reasonable

PR #444, opened 2018-07-12
Fix kerberos for ATH inside container

lvotypko:
Tests for kerberos ticket are failing when ATH is running inside docker. The reason is that they were written for localhost:<port> (communication between host and container) which is not the way how one container can communicate with another.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/KerberosContainer.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/KerberosContainer.java
@@ -23,12 +23,15 @@
  */
 package org.jenkinsci.test.acceptance.docker.fixtures;
 
+import org.apache.commons.io.IOUtils;
 import org.jenkinsci.test.acceptance.docker.Docker;
 import org.jenkinsci.test.acceptance.docker.DockerFixture;
 import org.jenkinsci.test.acceptance.docker.DynamicDockerContainer;
 import org.jenkinsci.utils.process.CommandBuilder;
 
+
 import java.io.File;
+import java.io.FileOutputStream;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.net.InetAddress;
@@ -66,17 +69,44 @@ public class KerberosContainer extends DynamicDockerContainer {
         }
     }
 
+    public String getIpAddress(){
+        String address = null;
+        try {
+            address = Docker.cmd(new String[]{"inspect"}).add(getCid()).popen().asText();
+            address = address.split("IPAddress")[2].split(":")[1].split(",",2)[0].trim().replaceAll("\"", "").replaceAll(",","");
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+        return address;
+    }
+
     private File targetDir = null;
     private File loginConf = null;
     private File krb5Conf = null;
 
     public File populateTargetDir(File target) {
+        try {
+            int count = 0;
+            while(!isKerberosRunning()){
+                if(count > 30){
+                    throw new Error("Kerberos is still not running after 30 seconds.");
+                }
+                Thread.sleep(1000);
+                count++;
+            }
+        } catch (Exception e) {
+            throw new Error(e);
+        }
         // No need to do this twice
         if (targetDir == null) {
             targetDir = target;
+            File keytabFile = new File(targetDir, "keytab");
+            keytabFile.mkdirs();
 
             // Get the keytabs out of the container
-            cp("/target/keytab/", target.getAbsolutePath());
+            copyContentOfFile("/target/keytab/user", new File(keytabFile,"user").getAbsolutePath());
+            copyContentOfFile("/target/keytab/service", new File(keytabFile,"service").getAbsolutePath());
+
             if (!new File(targetDir, "keytab/service").exists()) throw new AssertionError("Service keytab not created");
             if (!new File(targetDir, "keytab/user").exists()) throw new AssertionError("User keytab not created");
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/KerberosContainer.java
```diff
@@ -66,17 +69,44 @@
         }
     }
 
+    public String getIpAddress(){
+        String address = null;
+        try {
+            address = Docker.cmd(new String[]{"inspect"}).add(getCid()).popen().asText();
+            address = address.split("IPAddress")[2].split(":")[1].split(",",2)[0].trim().replaceAll("\"", "").replaceAll(",","");
```
olivergondza: Why not use `DockerContainer#inspect()` and select the value through `JsonNode`?

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/KerberosContainer.java
```diff
@@ -66,17 +69,44 @@
...
+        } catch (Exception e) {
+            e.printStackTrace();
```
olivergondza: Better fail then propagate incorrect IP address to fail in some later point.

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/KerberosContainer.java
```diff
@@ -104,6 +135,21 @@ public File populateTargetDir(File target) {
...
+    }
+
+
+    public void copyContentOfFile(String from, String to){
+        try {
+            IOUtils.copy(Docker.cmd("exec").add(getCid()).add ("base64", from).popen().getInputStream(), new FileOutputStream(new File("/tmp/coded")));
+            ProcessBuilder builder = new ProcessBuilder();
+            builder.command("base64", "-d", "/tmp/coded").redirectOutput(new File(to)).start().waitFor();
+        } catch (Exception e) {
+            e.printStackTrace();
```
olivergondza: Ditto

olivergondza:
Note the relevant tests are failing - presumably for some other reason.

lvotypko:
I see the tests are still failing from different reason now. I will check what is the cause now, so we do not change one failure for another.

rtyler:
Somehow this last Pipeline run generated 17GB of data on disk. That's a bit of a problem :-1:

lvotypko:
@rtyler , I do not think that it is problem of my fix. I do not have any trouble to run it in our Jenkins (I have the same pipeline) with last LTS of Jenkins. But I have a similar problem with last versions of Jenkins. And according to log which was generated (I did not read whole, ours had 8GB and then I killed it) it seems that problem was somewhere before my fix was even executed - in downloading plugins and dependencies (but I run only the test for Kerberos in this case, so I have a log of Jenkins in  console log of pipeline) - maybe it was generated for all tests in such big amount, but it was in only in jenkins log - not in the console log. If you want, I can check mime more and analyze it. Can you detect the 17GB log in your Jenkins and where was created such a big amount of data - console log of pipeline job, of jenkins instnace log which was running for tests? I will try to reproduce the problem on our instance, I believe that my commits did not cause it, more rather I hit some another problem by it.
Thanks!

lvotypko:
@rtyler, I send you a few e-mails about investigation. I do not want to spam this pull request with investigation of this issue, since it is probably not connected with it. But I suggest not to merge it until we find out what causes it - just to be sure.

rtyler:
It looks like the [5th build](https://ci.jenkins.io/job/Core/job/acceptance-test-harness/view/change-requests/job/PR-444/5/) didn't generate an excessive amount of disk usage.

It does appear that the issue was due to a failure in ATH to find dependencies. Some of the plugin tests have junit-attachment logs spammed with logs like the following:

```
master56949|WARNING: Could not find dependency jdk-tool of workflow-support
master56949|Jul 18, 2018 4:22:43 PM hudson.model.UpdateSite$Plugin getNeededDepende
ncies
```

lvotypko:
it is exactly what happened to me locally - but I got into the cycle (still not to sure why) - so my log was 8GB even for one test. Unfortunately I did not reproduce it (everything works now) and it is hard to tell what exactly caused it. Only thing which I can tell is that it is not caused by this pull request - because the problem is somewhere before the changed functionality was even performed.
 
You are right that Jenkins could not find dependencies (that log is not from ATH but copied log from Jenkins which is tested by ATH). The only explanation I was able to think is somehow corrupted update-cite repository json. It is obtained from real Jenkins update repository and after cashed for every test - (so this is a reason why this problem had all tests). Somehow in this cashed json did not have dependency for jdk-tool and command-launcher. But without seeing that json (which is now impossible because we have no job running with this issue), it is only my theory. I look at this like only explanation but very low probably (that is the reason why I am not sure) - according to log it was not empty json (so network failure is not probably) - because Jenkins would not find any dependency - but it found for example folder plugin, but not jdk-tool. Another reason is that it was corrupted, but it had to be corrupted in real update center of Jenkins, which is not very probably (at least without anybody noticing it). 

I do not want to freeze this pull request so I will create jira for it where I describe what happlened, give short example of logs. If it happens again we know what it is and what we need to get further with investigation. 

Thank you!

lvotypko:
https://issues.jenkins-ci.org/browse/JENKINS-52722

lvotypko:
@rtyler, do you agree with this approach of solving the issue (wait until it happens again and check cashed update center of ATH) and resolution that this issue is not connected with changes in this pull request?

lvotypko:
@olivergondza, do you agree with merging this? My changes which ended with 17GB occupied disk space was already solved in another pull request and I accepted its version. So now I do not see any reason why not merging it, do you have the same opinion?

PR #446, opened 2018-07-19
UI for adding user changed.

lvotypko:
UI for adding user in matrix authentication plugin has changed. This fix should work for Jenkins version 2.60+ and plugin version 2.3+. I do not think that we need backward compatibility for previous versions of plugins or Jenkins (I do not like to have a lot of if/else for due to jenkins/plugin version), anyone with different opinion?


lvotypko:
there are changes in login page, so first wait for pull request https://github.com/jenkinsci/acceptance-test-harness/pull/447

lvotypko:
after merging master with last pull request (fixing tests for new UI of login), the MatrixAuthPluginTest is finally passing.

daniel-beck:
I'm sorry for having missed this. Thanks for taking care of it!

romainverduci:
@lvotypko Some tests will break (and are actually breaking) if using a version < 2.3 of matrix auth plugin. I'm not sure forcing the upgrade of a plugin to make a test green is a good motivation. What's your opinion on that ?

PR #447, opened 2018-07-30
[JENKINS-52779] Fix Login#doLogin with the new Login page

romainverduci:
[Issue detail](https://issues.jenkins-ci.org/browse/JENKINS-52779)

Login was relying on https://plugins.jenkins.io/form-element-path and path selectors.
This is not working anymore with the new Login page introduced by https://github.com/jenkinsci/jenkins/commit/90670e05c17b34270cda9436eaf877cc72a341ec.

As the form elements attributes has not been changed by the new Login page, use By.name selectors so it's working with both new and old Login pages.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Login.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Login.java
@@ -1,6 +1,7 @@
 package org.jenkinsci.test.acceptance.po;
 
 import org.junit.Assert;
+import org.openqa.selenium.By;
 
 import static org.jenkinsci.test.acceptance.Matchers.hasInvalidLoginInformation;
 import static org.jenkinsci.test.acceptance.Matchers.loggedInAs;
@@ -12,9 +13,9 @@ import static org.jenkinsci.test.acceptance.Matchers.loggedInAs;
  */
 public class Login extends PageObject {
 
-    private Control cUser = control("/j_username");
-    private Control cPassword = control("/j_password");
-    private Control cLogin = control("/Submit");
+    private Control cUser = control(By.name("j_username"));
+    private Control cPassword = control(By.name("j_password"));
+    private Control cLogin = control(By.name("Submit"));
 
     public Login(Jenkins jenkins) {
         super(jenkins.injector, jenkins.url("login"));
@@ -40,7 +41,7 @@ public class Login extends PageObject {
     public Login doLoginDespiteNoPaths(String user, String password){
         driver.findElement(by.name("j_username")).sendKeys(user);
         driver.findElement(by.name("j_password")).sendKeys(password);
-        clickButton("log in");
+        driver.findElement(by.name("Submit")).click();
         return this;
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/Login.java
```diff
@@ -1,6 +1,7 @@
 package org.jenkinsci.test.acceptance.po;
 
 import org.junit.Assert;
+import org.openqa.selenium.By;
```
raul-arabaolaza: Not needed, the already present `by` object has a `name` method, see [ByFactory#L113](https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/main/java/org/jenkinsci/test/acceptance/ByFactory.java#L113

src/main/java/org/jenkinsci/test/acceptance/po/Login.java
```diff
@@ -12,9 +13,9 @@
  */
 public class Login extends PageObject {
 
-    private Control cUser = control("/j_username");
-    private Control cPassword = control("/j_password");
-    private Control cLogin = control("/Submit");
+    private Control cUser = control(By.name("j_username"));
```
raul-arabaolaza: Use `by.name`

src/main/java/org/jenkinsci/test/acceptance/po/Login.java
```diff
@@ -40,7 +41,7 @@ public Login doLogin(String user, String password){
     public Login doLoginDespiteNoPaths(String user, String password){
         driver.findElement(by.name("j_username")).sendKeys(user);
         driver.findElement(by.name("j_password")).sendKeys(password);
-        clickButton("log in");
+        driver.findElement(by.name("Submit")).click();
```
raul-arabaolaza: I tend to use `waitFor` to shield against performance issues instead of `findElement`

src/main/java/org/jenkinsci/test/acceptance/po/Login.java
romainverduci: Good catch

src/main/java/org/jenkinsci/test/acceptance/po/Login.java
romainverduci: Well the goal was just to avoid finding element by text but by name and do what clickButton is doing (would have done clickButton(by.name("...") if possible): https://github.com/jenkinsci/acceptance-test-harness/blob/5e551aa8b82398accdd45366add77bbaa08a98b5/src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java#L113

I tend to think that the wait for should be done at the page level and not for each element, no real strong opinion on that, open to other feedback, I can add it if needed.

olivergondza: I suggest not to pollute the code with unnecessary waits. Let's wait until it turns necessary.

PR #449, opened 2018-08-14
UI change in adding parser.

lvotypko:
In global config was added advanced button for displaying parser's section.


olivergondza:
Do we care for older versions?

uhafner:
No, support for the old version is not needed.

PR #450, opened 2018-08-14
Skip tests which require adding a user with matrix-auth plugin if version < 2.3

romainverduci:
As https://github.com/jenkinsci/acceptance-test-harness/pull/446 is introducing a non backward compatible change with matrix-auth plugin < 2.3, skipping the tests accordingly so if we run them with an older version, they won't fail.

@reviewbybees

### DIFF
--- a/src/test/java/plugins/JobDslPluginTest.java
+++ b/src/test/java/plugins/JobDslPluginTest.java
@@ -81,7 +81,7 @@ public class JobDslPluginTest extends AbstractJUnitTest {
      * Verifies that all configurations, done on the job configuration page,
      * are saved correctly.
      */
-    @Test @WithPlugins({"matrix-auth","mock-security-realm"})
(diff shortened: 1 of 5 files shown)


bmunozm:
🐝

olivergondza:
@lvotypko, I guess we need that compatibility...

olivergondza:
I do not feel strongly, I have verified we internally do not depend on that. User starts shows about a half of the users uses alder version: I do not feel strongly, according to the user stats

romainverduci:
After having a look at the code, implementing backward compatibility would be really ugly, I'm fine with skipping the tests until the upgrade of matrix-auth.

romainverduci:
@olivergondza The build is passing with the same failing tests as master.
If everyone is ok not having backward compatibility, can we merge this PR ?
In addition, could you please release the project so we can benefit from this PR and the fix for the new Login page ?

Thanks for you time :)

PR #451, opened 2018-08-15
Fix global publisher tests

lvotypko:
There are a few mistakes which were done when using docker cp was removed.


src/test/java/plugins/GlobalPublishPluginTest.java
```diff
@@ -485,7 +486,7 @@ public void publish_slave_resourses() throws IOException, InterruptedException,
         return new Matcher<DockerContainer>("container with file '%s' in it", path) {
```
olivergondza: The message now does not make sense

PR #454, opened 2018-09-17
Define id for clear identification in form.

lvotypko:
After merge of pull request #395 some tests which use select are broken (now the select does not search for option which contains given string, it has to match exactly). The problem is mostly in credentials select since there is used id (generated hash) or combination user name with /***** (and tests use for identification only username). I think that it is better to start use id of credentials since it is the only one unique identification for id. The most easy way how to do it is to use already implemented functionality and define which id should be used for credentials. 
Some of tests which use credentials are not broken, because they use their own implementation for identifying credentials. I let those test like they are and modify only the broken ones. I will modify them the same way after suggested change is approved.


lvotypko:
It seems working fine after merge.

PR #455, opened 2018-10-22
Prototype JDK 11 based build

olivergondza:
For local test against java 11: `env INTERACTIVE=true JENKINS_OPTS="--enable-future-java" VERSION_OVERRIDES="workflow-support=2.23-20181122.094059-1" ./run.sh firefox 2.150 -Dtest=...`

* JENKINS_JAVA_OPTS for JVM
  * Oleg: most likely - https://jenkins.io/blog/2018/06/17/running-jenkins-with-java10-11/#java-11


ath-container.sh
```diff
@@ -6,8 +6,9 @@ trap 's=$?; echo "$0: Error $s on line "$LINENO": $BASH_COMMAND"; exit $s' ERR
 uid=$(id -u)
 gid=$(id -g)
 tag="jenkins/ath"
+java_version="${java_version:-8}"
 
-docker build --build-arg=uid="$uid" --build-arg=gid="$gid" src/main/resources/ath-container -t "$tag"
+docker build --build-arg=uid="$uid" --build-arg=gid="$gid" --build-arg=java_version="$java_version" src/main/resources/ath-container -t "$tag"
```
oleg-nenashev: In PCT there was a decision to bubdle both Java versions into a single container: https://github.com/jenkinsci/plugin-compat-tester/pull/86 . It helps to ship an official PCT image which can be configured to use different versions depending on options. It would also allow running ATH engine in Java 8 while using Java 8 only for Jenkins itself

CC @raul-arabaolaza who was experimenting with ATH images iirc

olivergondza: Yeah, we would have to mimic that as I doubt we can reuse that container. I prefer to have ATH running on Java 11 and for now I am making progress so this is a good alternative in case I get stuck.

oleg-nenashev: I suspect that Selenium upgrade by @lvotypko may become a prerequisite. But it is merged in https://github.com/jenkinsci/acceptance-test-harness/pull/457

olivergondza: @oleg-nenashev, exaclty

MRamonLeon:
BTW, the construction of the docker image in the _Jenkinsfile_ doesn't use the _ath-container_ script, WDYT about syncing both buildings and executions (--shm-size 2g, VERSION_OVERRIDES, enable-future-java, ...) @olivergondza ? Thank you for this amazing work.

olivergondza:
@MRamonLeon, I was bitten by that shortly after you wrote it: https://github.com/jenkinsci/acceptance-test-harness/commit/46fc246317e07edc3345aec1a8dc87daec70f316

ath-container.sh
raul-arabaolaza: @oleg-nenashev @olivergondza FTR I am strongly against moving the `latest` tag of `jenkins/ath` docker image to java 11. That may have a strong impact on third parties and break a lot of builds as not everybody may be using latest version of the ATH sources ready for java 11 but they may be using latest version of the docker image. At least I would ask for keeping both versions by building and publishing a `java8` version of the image or including both JDKs in the image and let users decide which one to use like in the PCT example as suggested by Oleg

olivergondza:
Alright, I am merging this. Thank you everyone for your help!

PR #456, opened 2018-10-28
[JENKINS-54295] Fix docker image

raul-arabaolaza:
[JENKINS-54295](https://issues.jenkins-ci.org/browse/JENKINS-54295) This version was working perfectly for the 1.62 tag

* DO NOT MERGE YET, I HAVE CREATED THIS PR TO VALIDATE MY THEORY*

### DIFF
--- a/src/main/resources/ath-container/Dockerfile
+++ b/src/main/resources/ath-container/Dockerfile
@@ -20,7 +20,7 @@ RUN apt-get clean && \
         vnc4server
 
 # All we need is a statically linked client library - no need to install daemon deps: https://get.docker.com/builds/
-RUN curl -fsSLO curl -fsSL https://get.docker.com -o get-docker.sh && sh get-docker.sh
+RUN curl -fsSL https://get.docker.com -o get-docker.sh && sh get-docker.sh
 ENV SHARED_DOCKER_SERVICE true
 
 # Allow injecting uid and git to match directory ownership
--- a/src/main/resources/ath-container/Dockerfile
+++ b/src/main/resources/ath-container/Dockerfile
@@ -20,7 +20,8 @@ RUN apt-get clean && \
         vnc4server
 
 # All we need is a statically linked client library - no need to install daemon deps: https://get.docker.com/builds/
-RUN curl -fsSL https://get.docker.com -o get-docker.sh && sh get-docker.sh
+RUN curl -fsSLO https://download.docker.com/linux/static/stable/x86_64/docker-18.06.1-ce.tgz && \
+    tar --strip-components=1 -xvzf docker-18.06.1-ce.tgz -C /usr/local/bin
 ENV SHARED_DOCKER_SERVICE true
 
 # Allow injecting uid and git to match directory ownership
--- a/src/main/resources/ath-container/Dockerfile
+++ b/src/main/resources/ath-container/Dockerfile
@@ -20,8 +20,8 @@ RUN apt-get clean && \
         vnc4server
 
 # All we need is a statically linked client library - no need to install daemon deps: https://get.docker.com/builds/
-RUN curl -fsSLO https://download.docker.com/linux/static/stable/x86_64/docker-18.06.1-ce.tgz && \
-    tar --strip-components=1 -xvzf docker-18.06.1-ce.tgz -C /usr/local/bin
+RUN curl -fsSLO https://download.docker.com/linux/static/stable/x86_64/docker-17.03.2-ce.tgz && \
+    tar --strip-components=1 -xvzf docker-17.03.2-ce.tgz -C /usr/local/bin
 ENV SHARED_DOCKER_SERVICE true
 
 # Allow injecting uid and git to match directory ownership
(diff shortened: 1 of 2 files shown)


olivergondza:
@lvotypko, will this work as a temporary measure?

raul-arabaolaza:
@olivergondza This PR is not yet complete, I need more things to investigate given the error. We already have temporary measures in core that can be easily applied here by using `jenkins/ath:acceptance-test-harness-1.62` instead of `latest`. 
Still working on fix the master build, I will ping you when this works

lvotypko:
Hi, I am working on upgrade to selenium 3, because we can not be on current selenium version forever. But it takes me more time then I expected, so I am fine with setting the version of Firefox as temporary fix (I think it should be something lesser then 58, but not sure if I remember it well) and consider it as permanent setting. Because sometime it is hard to make it work for any browser version. 
error in build is caused wrong url for docker binaries. Try to change it for this one https://download.docker.com/linux/static/stable/x86_64/docker-18.06.1-ce.tgz

raul-arabaolaza:
Temporarily clossing to try to trigger a new PR builder run as the build got a temporarily unavailable error from debian apt repos

raul-arabaolaza:
The current failure is due to a breaking change introduced in the last `openjdk8-jdk` debian package (see ticket for details). My PR includes a workaround for that but as I am not a trusted source is not evaluated.
```
Loading trusted files from base branch master at 3618c0c3c2aa5b6318b2d990c3181fd93fc37972 rather than a52ba2b8403220b3d85725d2a2c63eaf15fb10ac
Obtained Jenkinsfile from 3618c0c3c2aa5b6318b2d990c3181fd93fc37972
```

@olivergondza or @lvotypko can you run my changes to check the validity? It worked on my local runs but I have no way to try them in the PR Builder

lvotypko:
@raul-arabaolaza, I use the same workaround for selenium3 and it works on our machines. I will try to run your branch too.

raul-arabaolaza:
@lvotypko any news?

lvotypko:
Hello, I realized in my pull request that there is a problem with version of docker, so we need binaries closer to previous version (there is a problem with sharing docker.sock). In my pull request it worked with docker-17.03.2. But your test failed even before. I will try to run it from Jenkinsfile (I do not use it because we have a different enviromnet and it needs a few modifications) and find out what is wrong. I had to use more memory with selenium 3. It is possible that current memory given by default is not enough even for selenium 2. But it is only theory.

lvotypko:
I tried it several time locally and have no problem with jvm crash. But you will need to change docker version too.
```
RUN curl -fsSLO https://download.docker.com/linux/static/stable/x86_64/docker-17.03.2-ce.tgz && \
    tar --strip-components=1 -xvzf docker-17.03.2-ce.tgz -C /usr/local/bin
```
Try to commit it and we will see, if it fails again.

src/main/resources/ath-container/Dockerfile
```diff
@@ -13,15 +13,15 @@ RUN apt-get clean && \
         git \
         imagemagick \
         iptables \
-        firefox-esr \
+        firefox-esr=52.9.0esr-1~deb9u1 \
```
jglick: FWIW I have had serious problems specifying exact versions in `apt` commands in `Dockerfile`s: after security fixes are released, vulnerable versions are _deleted_ from the repositories, and new `docker build`s start failing.

raul-arabaolaza: @jglick  This is only needed until #457 gets merged, I am not very fond to do big changes in the Dockerfile until then... Just the minimum to get master working again.

raul-arabaolaza:
@lvotypko Changed the docker version as you suggested

lvotypko:
Hello, 
for some reason in upstream log there is: 
 mvn test -Dmaven.test.failure.ignore=true -DforkCount=1 -B

but it should be:
 mvn test -Dmaven.test.failure.ignore=true -DforkCount=1 -DargLine="-Djdk.net.URLClassPath.disableClassPathURLCheck=true" -B

When I run your branch in my environment, I have desired 
-DargLine="-Djdk.net.URLClassPath.disableClassPathURLCheck=true"

I do not understand why it is not here... Any idea?

lvotypko:
I think I have an idea why it happens. But it is only theory... if the build is started with master Jenkinsfile and after the change is merged, it is running with Jenkinsfile on master not with Jenkinsfile in your branch. But it is only theory, otherwise I am not able to explain this behavior.

raul-arabaolaza:
That is a security feature, changes in Jenkinsfiles done by PR are only used if the owner of the PR is a trusted source, which for the ATH I am not :) That is the reason the pr builder for this change is not accurate and I am requesting you to run my changes

lvotypko:
Ok, now we can merge the commit. We can merge the commit so we have at least some results then #457  is merged. I hope it will be sooner then new version of Firefox. Thank you!

raul-arabaolaza:
Thanks!

PR #457, opened 2018-11-02
Selenium 3

lvotypko:
My builds have seemed stable for some time, open this pull request to check how it works in upstream builds.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,7 +14,7 @@
                 node('docker && highmem'){
                     checkout scm
                     def image = docker.build('jenkins/ath', 'src/main/resources/ath-container')
-                    image.inside('-v /var/run/docker.sock:/var/run/docker.sock') {
+                    image.inside('-v /var/run/docker.sock:/var/run/docker.sock --shm-size 2g') {
                         def exclusions = splits.get(index).join("\n");
                         writeFile file: 'excludes.txt', text: exclusions
                         realtimeJUnit(testResults: 'target/surefire-reports/TEST-*.xml', testDataPublishers: [[$class: 'AttachmentPublisher']]) {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -20,7 +20,7 @@
                         realtimeJUnit(testResults: 'target/surefire-reports/TEST-*.xml', testDataPublishers: [[$class: 'AttachmentPublisher']]) {
                             sh '''
                                 eval $(./vnc.sh)
-                                ./run.sh firefox latest -Dmaven.test.failure.ignore=true -DforkCount=1 -B
+                                ./run.sh firefox latest -DargLine="-Djdk.net.URLClassPath.disableClassPathURLCheck=true" -Dmaven.test.failure.ignore=true -DforkCount=1 -B
                             '''
                         }
                     }
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Jenkins.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Jenkins.java
@@ -3,10 +3,13 @@ package org.jenkinsci.test.acceptance.po;
 import java.io.IOException;
 import java.net.URL;
 import java.net.URLConnection;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 
+import com.google.common.base.Throwables;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.ObjectUtils;
 import org.jenkinsci.test.acceptance.controller.JenkinsController;
@@ -18,6 +21,7 @@ import com.google.common.base.Function;
 import com.google.inject.Injector;
 
 import hudson.util.VersionNumber;
+import org.openqa.selenium.WebDriverException;
 
 /**
  * Top-level object that acts as an entry point to various systems.
@@ -161,12 +165,18 @@ public class Jenkins extends Node implements Container {
         visit("restart");
         clickButton("Yes");
 
+        waitForLoad(JenkinsController.STARTUP_TIMEOUT);
+    }
+
+    public void waitForLoad(int seconds){
+        List<Class<? extends Throwable>> ignoring = new ArrayList<Class<? extends Throwable>>();
+        ignoring.add(AssertionError.class);
+        ignoring.add(NoSuchElementException.class);
+        ignoring.add(WebDriverException.class);
+        //Ignore WebDriverException during restart.
         // Poll until we have the real page
-        waitFor(driver).withTimeout(JenkinsController.STARTUP_TIMEOUT, TimeUnit.SECONDS)
-                .ignoring(
-                        AssertionError.class, // Still waiting
-                        NoSuchElementException.class // No page served at all
-                )
+        waitFor(driver).withTimeout(seconds, TimeUnit.SECONDS)
+                .ignoreAll(ignoring)
                 .until((Function<WebDriver, Boolean>) driver -> {
                     visit(driver.getCurrentUrl()); // the page sometimes does not reload (fast enough)
                     getJson("tree=nodeName"); // HudsonIsRestarting will serve a 503 to the index page, and will refuse api/json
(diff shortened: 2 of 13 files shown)


lvotypko:
it seems working. Now I check the failures if there is a regression in comparing with selenium 2.

Jenkinsfile
```diff
@@ -14,13 +14,13 @@
...
                     def image = docker.build('jenkins/ath', 'src/main/resources/ath-container')
-                    image.inside('-v /var/run/docker.sock:/var/run/docker.sock') {
+                    image.inside('-v /var/run/docker.sock:/var/run/docker.sock --shm-size 2g') {
                         def exclusions = splits.get(index).join("\n");
                         writeFile file: 'excludes.txt', text: exclusions
                         realtimeJUnit(testResults: 'target/surefire-reports/TEST-*.xml', testDataPublishers: [[$class: 'AttachmentPublisher']]) {
                             sh '''
                                 eval $(./vnc.sh)
-                                ./run.sh firefox latest -Dmaven.test.failure.ignore=true -DforkCount=1 -B
+                                ./run.sh firefox latest -DargLine="-Djdk.net.URLClassPath.disableClassPathURLCheck=true" -Dmaven.test.failure.ignore=true -DforkCount=1 -B
```
raul-arabaolaza: @jglick Points in #456 that we can just update surefire version to get the jdk fix see [here](https://github.com/jenkinsci/plugin-pom/pull/130) That seems better than having a temporary patch using a system property in this case

src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
```diff
@@ -95,23 +93,7 @@ protected final WebDriver visit(URL url) {
...
-            isStickyButton = (boolean)je.executeScript("return arguments[0].contains(arguments[1])", stickyContainer, e);
-        }
-        if (isStickyButton) {
-            Actions builder = new Actions(driver);
-            builder.moveToElement(e).click(e);
-            builder.perform();
-        } else {
-            e.click();
-        }
+        e.click();
```
raul-arabaolaza: Love this :)

lvotypko:
The test results are the same as with selenium 2 ( there is more failure but it is not caused by selenium 3, but higher browser version and bug in java - some tests run build with maven too). So now it is time to discuss changes. My changes are created from #360. I took the rtyler's branch and fix things which did not work the same as for selenium 3 and did a few workarounds for bugs in current version of geckodriver and marionette.

Jenkinsfile
lvotypko: Great, it is really better.

src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
lvotypko: Not sure why it was done since it is not my change, but I guess he had the reason, so I let it this way. But it probably caused the issue fixed in this commit https://github.com/jenkinsci/acceptance-test-harness/pull/457/commits/b45a4251583f5be64a1a8ac5d0754aef6acd3638

I will try to return it and disable my fix if it will be work old way without causing any other issue. Before it was hard to judge because there failed more then half tests. Now it should be visible, if it does not break anything, it should be better this way, thank you for point it.

lvotypko: I tried it and it seems there is a problem with old implementation (which is the reason why it was removed), all tests fails with:

`The element reference of <button id="yui-gen7-button" type="button"> is stale; either the element is no longer attached to the DOM, it is not in the current frame context, or the document has been refreshed`

I look at it more to decide if there is an easy way how to fix it there (we can always let there my fix in Scroller). Or you have any suggestions?

lvotypko: Oh sorry, my bad, I did mistake in returning it to the previous state. Now the tests are running.

src/main/java/org/jenkinsci/test/acceptance/po/Jenkins.java
```diff
@@ -161,19 +165,22 @@ public void restart() {
...
 
+        waitForLoad(JenkinsController.STARTUP_TIMEOUT);
+    }
+
+    public void waitForLoad(int seconds){
+        List<Class<? extends Throwable>> ignoring = new ArrayList<Class<? extends Throwable>>();
+        ignoring.add(AssertionError.class);
+        ignoring.add(NoSuchElementException.class);
+        ignoring.add(WebDriverException.class);
+        //Ignore WebDriverException during restart.
```
olivergondza: Are we ok to swallow that in non-restart scenarios, too?

src/main/resources/ath-container/Dockerfile
```diff
@@ -20,10 +20,15 @@ RUN apt-get clean && \
...
 
 # All we need is a statically linked client library - no need to install daemon deps: https://get.docker.com/builds/
-RUN curl -fsSLO https://get.docker.com/builds/Linux/x86_64/docker-17.04.0-ce.tgz && \
-    tar --strip-components=1 -xvzf docker-17.04.0-ce.tgz -C /usr/local/bin
+RUN curl -fsSLO https://download.docker.com/linux/static/stable/x86_64/docker-17.03.2-ce.tgz && \
+    tar --strip-components=1 -xvzf docker-17.03.2-ce.tgz -C /usr/local/bin
+
 ENV SHARED_DOCKER_SERVICE true
 
+# Selenium needs a geckodriver in order to work properly
```
olivergondza: In general or for firefox only? Why it is not installed through `apt-get`? I presume there are reasons but they are not obvious to me - so please document them.

src/main/java/org/jenkinsci/test/acceptance/selenium/Scroller.java
```diff
@@ -73,13 +73,23 @@ public void beforeClickOn(WebElement element, WebDriver driver) {
...
     }
 
     private void scrollIntoView(WebElement e, WebDriver driver) {
         int eYCoord = e.getLocation().getY();
         int eXCoord = e.getLocation().getX();
-        ((JavascriptExecutor)driver).executeScript(scrollJs, eYCoord, eXCoord);
+        String id = e.getAttribute("id");
+
+        WebElement response = (WebElement) ((JavascriptExecutor)driver).executeScript(scrollJs, eYCoord, eXCoord, id);
+        if(response!=null && response.getAttribute("style").contains("bottom: 0px")){
```
olivergondza: Since we have to add custom handling for some elements to our scrolling logic, I suggest to either keep it either fully in javascript land or the java land. I speculate the reason for the split is that sleeping in JS is tricky. Here are some suggestion I consider cleaner:

- Return from JS a clear indicator we have hit a specific condition to handle (so java code do not have to inspect the CSS attached).
- Or provided this is a temporary condition, return `true` if we scrolled and `false` if we hit the special condition so the java code will sleep and _retry_ the scrolling logic until it succeeds.

src/main/resources/ath-container/Dockerfile
lvotypko: I think it is necessary only for Firefox.
I am not sure if it is available from apt-get - it was from rtyler, but it is worthy to check.

src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
lvotypko: After returning to the previous state (with sticky buttons detection), there are a lot of more failure, usually with visiting page after saving configuration. But it does not seems as the failure of clicking on save button, but rather that something unexpected happens after clicking. So I am not sure if it is connected with it, but that is the only change. 

@rtyler Do you know why this was done?

src/main/resources/ath-container/Dockerfile
lvotypko: It is not possible from apt-get.

src/main/java/org/jenkinsci/test/acceptance/selenium/Scroller.java
lvotypko: I moved it into JS which return true if it is successful scrolled and false if we hit the footer overlap condition. In this case there is retry until it succeeds with timeout 5 seconds.

olivergondza:
The tests now works with almost arbitrary firefox version I have installed without the need to use docker for local development. This is awesome! :tada::fireworks: :tada:

oleg-nenashev:
+1 to what @olivergondza says! It should be a great improvement for ATH. Finally we will be able to test against Firefox versions which are actually used by Jenkins users

PR #459, opened 2018-11-19
ath-container, remove pinned version for firefox-esr

olblak:
I don't know why firefox-esr was pinned down to that specific version
but what I know is that the docker image doesn't build [anymore](https://ci.jenkins.io/job/Core/job/acceptance-test-harness/view/change-requests/job/PR-458/2/console) .
A small comment would been usefull. 
@raul-arabaolaza apparently you recently made that change 
`2018-10-28 19:16:55 +0100 16)         firefox-esr=52.9.0esr-1~deb9u1 \`

### DIFF
--- a/src/main/resources/ath-container/Dockerfile
+++ b/src/main/resources/ath-container/Dockerfile
@@ -13,7 +13,7 @@ RUN apt-get clean && \
         git \
         imagemagick \
         iptables \
-        firefox-esr=52.9.0esr-1~deb9u1 \
+        firefox-esr \
         maven \
         openjdk-8-jdk \
         unzip \


raul-arabaolaza:
@olblak You have all the information in #456 long history short the debian esr package was updated to a version no longer compatible with the selenium version used by the ATH. With the result of any ATH run using the `latest` tag of the docker image failing all tests for more than one month and impacting any third party using it (or even core builds)

It [was working](https://ci.jenkins.io/job/Core/job/acceptance-test-harness/job/master/104/) when I made the change so my guess is that now the version used is no longer on the debian package. 

FTR I am strongly against reverting the change as that would imply publishing a `latest` tag that is not able to run the ath tests. Sure the docker build may be broken now but at least we are not breaking any build by publishing a docker image unable to run the tests.

The definitive solution is IMO in #457 which will update the selenium version of the ATH to be able to run with the ESR firefox version in the debian repos

olivergondza:
I tend to agree with Raul. @olblak, is this causing any problems apart from those we know of?

lvotypko:
#457 is merged.

PR #460, opened 2018-11-30
Display maven and JDK versions from `run.sh`

batmat:
Can we useful as we're adding JDK 11 support and people use this for various JDK versions and be able to check they're actually running what they think.

cc @jenkinsci/java11-support @olivergondza

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -94,4 +94,4 @@ shift 2
 
 set -x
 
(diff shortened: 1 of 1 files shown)


olivergondza:
Better than https://github.com/jenkinsci/acceptance-test-harness/pull/455/commits/d38ecb91d970a7381eb446c0401321c1317946c2 in jdk11 branch.

PR #461, opened 2018-11-30
Debug gradle plugin.

lvotypko:
This pull request should not be merged. It is only for debugging gradle test failures in upstream build, because I am not able to reproduce it.

### DIFF
--- a/src/test/java/plugins/GradlePluginTest.java
+++ b/src/test/java/plugins/GradlePluginTest.java
@@ -23,7 +23,6 @@
  */
 package plugins;
 
-import org.apache.commons.io.IOUtils;
 import org.jenkinsci.test.acceptance.junit.AbstractJUnitTest;
 import org.jenkinsci.test.acceptance.junit.WithPlugins;
 import org.jenkinsci.test.acceptance.plugins.gradle.GradleInstallation;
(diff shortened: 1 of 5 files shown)


src/test/java/plugins/GradlePluginTest.java
```diff
@@ -35,6 +36,8 @@
 import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
 
+import java.io.InputStream;
```
olivergondza: imports are unused now

lvotypko:
Now it seems working, Unfortunately due to history of builds with one test class it executed almost all tests in every split. 
But it does not cause any regression in other tests and gradle is fixed.

src/test/java/plugins/GradlePluginTest.java
```diff
@@ -54,6 +57,8 @@
 
     /**
      * Verify the execution of a basic gradle build script.
+     *
+     * Prevent to run gradle daemon because attempt to clean processes after build put daemon into defunct mode and build timeout.
```
lvotypko: The cause of strange behavior is gradle daemon process. It stayed in defunct mode after process killer tried to clean processes in the end of a build. I was not able to find what is the cause of this behavior of gradle daemon (or some reported bug which describes exactly the same situation). Since it does not happen in my environment, I think that it can be connected with docker in combination with specific environment of its host. 
I did not want to spend hours by searching because there is no need to run gradle daemon for ATH testing.
So I set --no-daemon switch for tests. 
@olivergondza are you alright with documentation for understanding why gradle daemon is not used?

PR #470, opened 2019-01-03
 [INFRA-1953] Fix run ATH on java 11 via Jenkinsfile

MRamonLeon:
See [Make pipeline-library#runATH step runs with old ATH sources and images
](https://issues.jenkins-ci.org/browse/INFRA-1953)

While adding backward compatibility for pipeline-library#runATH step, I realized the steps to run the ATH via its Jenkinsfile are not right. We need to call the new set-script.sh to change the java in use and the java_version build-arg is not needed anymore.

If you look at the builds they all are using Java8 while they should be using Java11 in one of its stages. For example: https://ci.jenkins.io/job/Core/job/acceptance-test-harness/job/PR-467/3/execution/node/312/log/?consoleFull

@jenkinsci/java11-support @raul-arabaolaza @olivergondza

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,7 +15,7 @@ for (int j in [8, 11]) {
             stage(name) {
                 node('docker && highmem') {
                     checkout scm
-                    def image = docker.build('jenkins/ath', "--build-arg=java_version=$javaVersion src/main/resources/ath-container")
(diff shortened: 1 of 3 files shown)


olivergondza:
I am merging this right away as the PR build is using master Jenkinsfile anyway. @MRamonLeon, thanks for noticing the problem and fixing it!

PR #473, opened 2019-01-04
[JENKINS-55406] Remove machine provider support

olivergondza:
I doubt this is used at all these days.

TODO:

- `SlaveProvider` and `SlaveController` abstractions are now useless.
- Eliminate the dependencies we no longer need after this.


olivergondza:
@raul-arabaolaza, that would be much appreciated. The code was contributed by cloudbees but I never heard of anyone actually using it so if you can confirm you guys do not need it, it would be great.

src/main/java/org/jenkinsci/test/acceptance/machine/JenkinsProvider.java
```diff
@@ -1,10 +0,0 @@
-package org.jenkinsci.test.acceptance.machine;
```
jtnord: this is used by a commercial test suite.

raul-arabaolaza:
And I have confirmed another usage of `JenkinsProvider` apart of the one noted by @jtnord so I am requesting changes to block this until I can suggest a safe path

olivergondza:
Does the custom `JenkinsProvider` depends on ability to spin machines in some sort of cloud or is it just local jenkins(es).

olivergondza:
Right. My main motivation here is to get rid of jclouds that has high potential of dependency clashing when ATH used as a dependency to Jenkins plugin. There seem to be a way to make the cut without impacting you.

@raul-arabaolaza, will you restore the classes you know you need or should I attempt to do it?

raul-arabaolaza:
@olivergondza I can do that if you give me some time, I am full at this moment

raul-arabaolaza:
FTR I expect to be able to have some time next week, sorry for the big delay, I need to do this on my free time and I have some personal issues limiting that at this moment.

raul-arabaolaza:
Sorry, some emergency work is preventing me to progress here, I have not forgot about this PR, I promise

olivergondza:
@raul-arabaolaza, Do you happen to have good news for us? :D

olivergondza:
@raul-arabaolaza, any update?

raul-arabaolaza:
@olivergondza So sorry, I completely forgot about this, I will make sure to review during this week

olivergondza:
@amuniz, @MRamonLeon, perhaps some of you might know?

amuniz:
I'm checking...

amuniz:
So, `JenkinsProvider` is definitely used. However I don't see any usage of other APIs (like `Machine` or `MachineProvider`). Removing all specific implementation is ok for sure (`EC2Provider`, etc). APIs need a deeper look, @raul-arabaolaza tells me he is going to do it (for real this time) :smile: 

Thanks Raul!

raul-arabaolaza:
I confirm what @amuniz has said, only `JenkinsProvider` is needed AFAICT. Right now I have restored just that file back and everything compiles fines, I am going to run some tests just to be sure it works.

raul-arabaolaza:
The good point is that given `JenkinsProvider` is just an standard Guice `Provider` even if removed from here we can restore it on our proprietary ATHs without a big impact as suggested by @amuniz so a little bit of extra work but nothing too serious.

olivergondza:
Thank for the review guys! I am keeping the `JenkinsProvider` in place, unused, deprecated. Will go on and merge unless someone complains.

PR #478, opened 2019-02-11
[JENKINS-56018] ATH Docker image should not be using Module patches starting from 2.163 

MRamonLeon:
See [[JENKINS-56018] ATH Docker image should not be using Module patches starting from 2.163](https://issues.jenkins-ci.org/browse/JENKINS-56018).

As of 2.163 the modules are not needed anymore. We change the documentation to reflect that, but also we write down how you need to run ATH in previous versions, as ATH could be used with whatever version.

To make it easy, we stay downloading the jaxb libraries to let just specify them in Jenkins < 2.163. In other way, the user will need to download them by him/herself.

**WARNING**: To be merged when Jenkins 2.164 is out.


docs/JAVA11.md
```diff
@@ -4,18 +4,33 @@ To run the tests using a Java 11 virtual machine:
...
-1. In the container shell, set up the vnc and run the tests with Java11 support:
+1. In the container shell, set up the vnc:
 
    ```bash
    ath-user@1803848e337f:~/ath-sources$ eval $(./vnc.sh)
    ```
 
+1. And run the tests with Java11 support:
+
+   Most likely, you will need to add some modules left apart from Java 9, like `JAVA_OPTS="--add-modules java.sql"` used by _pipeline-maven-plugin_ for example. It depends on the Jenkins version and the plugins to test. You can find the most updated documentation on how to run Jenkins on Java 11 at: https://jenkins.io/doc/administration/requirements/jenkins-on-java-11. Basically:
```
oleg-nenashev: Is it true for ATH? If yes, we may need to fix this bug? It would be a platform classloader issue in the test suite

docs/JAVA11.md
```diff
@@ -4,18 +4,33 @@ To run the tests using a Java 11 virtual machine:
...
+
+   For Jenkins [2.155, 2.163), add modules, _enable-future-java_ and dependencies from tests:
```
batmat: I would simply strongly recommend avoiding these versions on Java 11. Only "experts" should try them (or we put this in some addendum, not in the main docs flow IMO).

docs/JAVA11.md
```diff
@@ -4,18 +4,33 @@ To run the tests using a Java 11 virtual machine:
...
+   ```bash
+   ath-user@1803848e337f:~/ath-sources$ env JAVA_OPTS=\"${JAVA_OPTS} -p /home/ath-user/jdk11-libs/jaxb-api.jar:/home/ath-user/jdk11-libs/javax.activation.jar --add-modules java.xml.bind,java.activation -cp /home/ath-user/jdk11-libs/jaxb-impl.jar:/home/ath-user/jdk11-libs/jaxb-core.jar\" JENKINS_OPTS=\"--enable-future-java\" ./run.sh ...
```
batmat: Following my previous comment, I would say we should now put this only as an addendum or so, and not use `--enable-future-java` (or possibly simpler `-e JENKINS_ENABLE_FUTURE_JAVA=true`.

MRamonLeon:
Regarding @batmat comments, I think the documentation about how to run jenkins on java 11 is short enough to keep the instructions depending on the Jenkins version. There are already people testing Jenkins < 2.164 on java11 and this kind of documentation could be very valuable for them. 

In the other hand, if we want to avoid testing Jenkins < 2.164 on Java 11 due to the complexity, we should avoid talking about that in the rest of documentation already published. Because if we state it works on 2.155 and someone comes to test ATH in such version, the instructions should be clear enough.

I'm open to reduce the documentation as soon as we do the same in the other docs.

docs/JAVA11.md
MRamonLeon: If it's not needed on Jenkins, it's not on ATH

.gitignore
```diff
@@ -23,3 +23,4 @@ jenkins*home/
 jenkins-plugins/
 .jenkins_test/
 !src/test/resources/jacoco/test/target
+.history
```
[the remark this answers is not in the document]
MRamonLeon: No no, it's due to Visual Studio Code

raul-arabaolaza:
I strongly disagree to remove doc related to "old" versions of core, there are people building things on top of LTS and those "old" versions and more than probably they would like to start testing with java 11 as soon as possible. Removing doc and saying "you should be using this version of the core" is going to make things harder for them... unless the whole point here is that java 11 testing is meaningless unless you are on core `2.164`

alecharp:
@raul-arabaolaza I strongly disagree. 

We don't plan to remove how to use ATH with any LTS (there is no LTS with Java 11 yet). But we don't recommand to use any Jenkins version prior to `2.164` because it is way easier to use than the `[2.155,2.163]` versions. So, yes for the moment it means we recommend to use only 1 version of Jenkins core to run the tests on Java 11. 

But it's for the good of the developers experience.

MRamonLeon:
As commented, because the statement is that _the recommended version to test Java11 is 2.164_, I keep the documentation simple, and I link to the Jenkins documentation to get further information on how to run Jenkins on Java 11

MRamonLeon:
Removed the _jaxb params_ and `--enable-future-java` in the Jenkinsfile

MRamonLeon:
I think the comments are addressed @alecharp @batmat , but the CI is throwing:

_workflow-aggregator plugin not installed_
https://ci.jenkins.io/job/Core/job/acceptance-test-harness/view/change-requests/job/PR-478/4/testReport/plugins/AnalysisCollectorPluginTest/java_11_split0___should_compute_annotations_on_workflow/

I need to review that, any idea?

MRamonLeon:
@batmat Have you refreshed? I do think I already changed it.

olivergondza:
Thank you Ramon, this makes things cleaner.

MRamonLeon:
Thank you @olivergondza. Can we have a docker image with these changes? It would be great.

PR #482, opened 2019-02-15
[JENKINS-55406] Move scripts to be invoked from container to the container itself

olivergondza:
This makes no practical difference for ATH itself as the container mounts ATH
sources so scripts can be accessed that way. However, external consumers of the
library/container have to duplicate `vnc.sh`, `set-java.sh`, `run.sh` in one way or
another to run the tests. Moving those scripts to the container so no part of ATH codebase has to be duplicated/fetched.

https://issues.jenkins-ci.org/browse/JENKINS-55406


olivergondza:
The CI build has passed the critical part and and I have tested the image against kerberos-sso plugin locally.

MRamonLeon:
Does It make sense to include the selection of the java version inside the run.sh? It's like another option, like the browser. Then, the call should be:

run.sh firefox java11 latest ...

It could simplify even more the execution. I don't know exactly what the impact of this change could be, but as the script (set-java.sh) was introduced recently, I don't think It should be too big.

WDYT? @olivergondza @raul-arabaolaza ?

olivergondza:
@MRamonLeon, we would have to make it optional not to break current clients of `run.sh` and that feels quite challenging since third argument and onward is pass-through. Also, `run.sh` is used out of the container while `set-java.sh` is quite coupled with the container we distribute (in fact, the only reason I kept the symlink for `set-java.sh` is not to break people's automation - i belongs to the container).

olivergondza:
@raul-arabaolaza, That is what I aim to achieve by pointing symlinks from old locations to the new one so this will still work in way it used for people checking out ATH sources as well as for those who does not. I works for my tests but I would love to hear in case it does not work for you, guys.

raul-arabaolaza:
@olivergondza Oh, my bad, sorry I didn't realize the symlinks... is perfect for me :)

olivergondza:
Merging to get the :latest docker image for testing. Will not release until we are confident in this.

PR #485, opened 2019-02-19
[JENKINS-56198] Change the assert to pass on Java11 and test for Java 11 to fix the bug

MRamonLeon:
See **[[JENKINS-56198] Source details inference fails on Java 11 (Fix test, not the bug)](https://issues.jenkins-ci.org/browse/JENKINS-56198)**

This PR fixes the test to let Java 11 execution to pass. The test is still valid, we only change the text we are looking for in the console. The root cause of the bug is not addressed here, it needs to be analysed and fixed (or not) in the ticket: https://issues.jenkins-ci.org/browse/JENKINS-54942

As this behavior can be changed due to the internals of the JVM and may be hard to fix, we consider a good approach to change the test to have the Java 8 and Java 11 as close as possible.

To make easy fixing the real bug avoiding masking the failure, we keep a test failing on Java 11 (the original one with the original search on the console) but renamed and ignored. We've [added a comment](https://issues.jenkins-ci.org/browse/JENKINS-54942?focusedCommentId=360987&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-360987) on the ticket in charge of addressing the bug to know the test to use for testing the bad behavior. 


src/test/java/plugins/JobDslPluginTest.java
```diff
@@ -470,10 +472,47 @@ public void should_fail_on_missing_plugin() {
...
+
+    /**
+     * Verifies whether the build will be marked as failed or unstable if a plugin must
+     * be installed to support all features used in the DSL script. To check the right behavior we used the old assert
+     * which is failing on Java 11. This test can be used to fix the root cause.
+     */
+    @Ignore("Currently failing on Java 11, use this test to fix https://issues.jenkins-ci.org/browse/JENKINS-54942")
+    @Issue("https://issues.jenkins-ci.org/browse/JENKINS-54942")
+    @Test
+    public void should_fail_on_missing_plugin_but_doesnt_on_java_11() {
```
MRamonLeon: The old test renamed and ignored to help when addressing the bug

src/test/java/plugins/JobDslPluginTest.java
```diff
@@ -470,10 +472,47 @@ public void should_fail_on_missing_plugin() {
                          "}");
         seedJob.save();
         Build build = seedJob.scheduleBuild().shouldBeUnstable();
+        assertThat(build.getConsole(), containsString("plugin 'chucknorris' needs to be installed"));
```
MRamonLeon: just change the assert to pass on Java 11

src/test/java/plugins/JobDslPluginTest.java
```diff
@@ -470,10 +472,47 @@ public void should_fail_on_missing_plugin() {
...
+
+        seedJob.configure(() -> jobDsl.setFailOnMissingPlugin(true));
+        build = seedJob.scheduleBuild().shouldFail();
+        assertThat(build.getConsole(), containsString("plugin 'chucknorris' needs to be installed"));
```
MRamonLeon: just change the assert to pass on Java 11

olivergondza:
Do we want to ignore all problematic tests like that? RVM plugin is known broken on Java 11 as well...

Me preference would be to track problems in https://docs.google.com/document/d/1oluVrNVpQhXCIwW9CYVm09Y1vPc3H77d3q92LrzcpDw/edit#heading=h.xtman4uqqxz0

FTR, the remaining tests are wither fixed or on a way to get fixed:

- https://github.com/jenkinsci/acceptance-test-harness/pull/484
- https://issues.jenkins-ci.org/browse/JENKINS-51852?focusedCommentId=360992&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-360992
- https://github.com/jenkinsci/parallel-test-executor-plugin-sample/pull/2

MRamonLeon:
@olivergondza **_RvmPluginTest.smokes_** is failing on Java 8 as well. Due to that it doesn't showed on my comparison.

MRamonLeon:
I used this approach because:

1. I thought it was the only difference between both executions
1. having both executions with the same results could help us to find out more easily Java 11 related bugs. 
1. the fix look like it's not straightforward

I agree to track the problems there, I wrote the results previously to this document. This PR doesn't excuse us to do that, absolutely. If there is consensus about it's not needed, I'm fine closing it.

olivergondza:
Good point with the rvm, I have filed [JENKINS-56211](https://issues.jenkins-ci.org/browse/JENKINS-56211) to get it fixed. We would like to get the contribution guidelines created for ATH that would define a process here to set a due date for fixing such problems that would permits us to get rid of abandoned tests in efficient and defined way.

src/test/java/plugins/JobDslPluginTest.java
olivergondza: Is it me or the diff is incredibly confusing here? It appears you have fixed the test by _adding_ assertions while you have _modified_ lines instead.

[1] https://github.com/jenkinsci/acceptance-test-harness/blob/7e8c9b20d25e420c37940313f8243e21c86be1bb/src/test/java/plugins/JobDslPluginTest.java#L472-L477

src/test/java/plugins/JobDslPluginTest.java
olivergondza: Frankly, I do not feel strongly. You are right the part that causes assertion to fail is sort of cosmetic and the main thing is correct here (the missing plugin is repported correctly). I am ok narrowing the scope of the test to move forward here. On the other hand, we caught a regression thanks to that and it makes sense to keep it to work as a regression test for plugin authors.

src/test/java/plugins/JobDslPluginTest.java
MRamonLeon: Yes, I saw the same. It's due to the change on the _assertThat_ that the diff becomes crazy. Due to that, I added some comments to the PR :man_shrugging:

MRamonLeon:
Test success:
screenshot from 2019-02-20 09-57-32

src/test/java/plugins/JobDslPluginTest.java
batmat: Curious: wouldn't adding some `assumeTrue(isOnJava8OrLess())`, instead of adding a new ignored test, achieve some similar outcome?

I think it's generally better to try and avoid adding more ignored tests, WDYT?

MRamonLeon: The test is still valid with this assert and we still want the failing test on Java 11 for using it to fix the loss of information in the log (if it's really possible to fix).That's the reason I didn't use an approach like you suggest.

src/test/java/plugins/JobDslPluginTest.java
MRamonLeon: Are you suggesting to keep this test working, remove the @Ignore? Or just close the PR and keep the test failing? What I wouldn't like to loose is the fact that the test is doing the job, although without the information regarding the line. Perhaps just a comment on the original test warning about that to help when we address the bug itself. WDYT?

src/test/java/plugins/JobDslPluginTest.java
```diff
@@ -470,10 +472,47 @@ public void should_fail_on_missing_plugin() {
...
+    }
+
+    /**
+     * Verifies whether the build will be marked as failed or unstable if a plugin must
+     * be installed to support all features used in the DSL script. To check the right behavior we used the old assert
+     * which is failing on Java 11. This test can be used to fix the root cause.
+     */
+    @Ignore("Currently failing on Java 11, use this test to fix https://issues.jenkins-ci.org/browse/JENKINS-54942")
+    @Issue("https://issues.jenkins-ci.org/browse/JENKINS-54942")
```
oleg-nenashev:
```suggestion
    @Issue("JENKINS-54942")
```

batmat:
> Do we want to ignore all problematic tests like that? RVM plugin is known broken on Java 11 as well...

@olivergondza I for one would definitely like us to either fix quickly, or _ignore_ failing tests. I believe that we should aim constantly for a green build so that people not deeply knowledgeable about ATH status don't risk spending too much time understanding that actually a test they added or modified is fine. 

Now, as it's generally unrealistic to fix tests in a reasonable timeline because most maintainers are people doing fixes on their free time, I think a correct way to handle this would be to define something like a `@KnownToFail` JUnit category to tag these failing tests.

Then, we would define something like a "main" ATH stage, that would be always kept green, and critical failures be fixed immediately. And we'd add a latter non-blocking stage to run the `@KnownToFail` category and still get a status report for these.

(NOTE: why not just use `@Ignore` then? Indeed, that would be my intent, but using a JUnit category is somehow AFAIU the only way to be able to still run these tests. After a quick look recently, it seems very difficult to force running ignored tests, when running a special JUnit category for this instead would be trivial).

If you think this makes sense, I think we can try and see how we could work together on making it happen?

olivergondza:
@batmat, I do agree with you we can do better in pointing out _why_ tests are failing in case we know the reason (in this case we know Job DSL test is failing as the plugin was not adjusted for Java 11 yet). However, I assert those tests must fail to point out the bug itself. Without that, the testsuite would not report RVM plugin have problems with Java 11 and reported green status instead - that is not what we want.

I like the idea of annotating known-failing tests but a) we need a policy to deal with such problems in timely manner so they do not stay that way forever and b) I would like them to point to the cause (JIRA, PR, etc.). That would need a bit more though as JUnit categories are not annotations and we need something from both of the worlds.

Though, I appreciate the offer to help. I have moved our plan one step further: https://groups.google.com/d/msg/jenkinsci-dev/W7kA5822CHg/kCp-SpK6AgAJ. Please, have a look.

MRamonLeon:
I know this PR has involved in a helpfull discussion about how to improve the ATHs, something we already knew and it's being discussed on _jenkinksci-dev_. Regarding this PR, IIUC you opt for closing it, isn't it @olivergondza?

olivergondza:
Not necessarily, how about merging it without `should_fail_on_missing_plugin_but_doesnt_on_java_11`? It will effectively narrow down the scope of the test to the fact missing plugin is reported correctly but we will leave the detail of the source line not being correct for the plugin authors to test.

MRamonLeon:
SGTM, done, now the diff is straightforward.

@olivergondza @batmat does it look good now to merge? we'll work on the java11 issue on the  ticket of the plugin and the ATH flow improvements in the jenkins-dev group

olivergondza:
Thanks Ramon!

daspilker:
Hi, I fixed the problem in Job DSL. Do I need to release the fix before it can be tested in ATH?

https://github.com/jenkinsci/job-dsl-plugin/pull/1170

batmat:
@daspilker you could configure your Jenkinsfile so it uses the `runATH` pipeline library.
It is a _bit_ tricky, as an example you can look at https://github.com/jenkinsci/jenkins/blob/master/Jenkinsfile#L100.

You can also test it locally if you feel like it. https://github.com/jenkinsci/acceptance-test-harness/blob/master/docs/SUT-VERSIONS.md could be helpful.

@olivergondza is probably the best to advise the next course of action though given his deep knowledge of ATH.

olivergondza:
@daspilker, thanks!

You can tests things locally by `/acceptance-test-harness $ env LOCAL_JARS=path/to/your/job-dsl.hpi ./run.sh firefox latest -Dtest=JobDslPluginTest` while having Java 11 on PATH.

When it comes to getting the upstream test suite to pass, yes, you need to release first. Not to mention it is desirable to have it public for users on Java 11.

[1] https://github.com/jenkinsci/acceptance-test-harness/blob/master/docs/SUT-VERSIONS.md#use-custom-plugin-file

PR #491, opened 2019-03-11
Remove test for Rvm plugin.

lvotypko:
Remove Rvm plugin. It does not have a lot of instances and the scenario does not meet the usual Jenkins usage - it builds on master and require right for installing packages there (which is not good idea to do).


oleg-nenashev:
Cc @jenkinsci/java11-support

PR #492, opened 2019-03-17
Allow user to provide location of driver

henryju:
Current code will try to locate the driver in the OS PATH and always ends up setting the driver specific location property, regardless the property was already set.
This change preserve user value, allowing to have the driver *not* in the PATH.

Was discussed on the group: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/jenkinsci-dev/BNuQMy3lsFY/yMntNg0rBwAJ


olivergondza:
Thanks!

PR #493, opened 2019-03-17
Support installation of multiple tools versions

henryju:
Allow to install more than one tool of the same type, by expanding the collapsed tool area in recent Jenkins versions.

I have successfully tested the fix on our own test suite (SonarQube plugin) but I'm not very pleased by the proposed fix (try..catch) so if you know a better way, feel free to change.

I also tried to update a test to demonstrate what I'm trying to fix, but I didn't run it for real on my box.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/ToolInstallation.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/ToolInstallation.java
@@ -77,6 +77,12 @@ public abstract class ToolInstallation extends PageAreaImpl {
         final ConfigurablePageObject page = ensureConfigPage(jenkins);
 
         final String name = type.getAnnotation(ToolInstallationPageObject.class).name();
+        final Control expandButton = page.control(by.button(name + " installations..."));
+        try {
+            expandButton.click();
+        } catch (Exception e) {
+            // Ignore, this is likely because this is the first installation of this tool
+        }
         final Control button = page.control(by.button("Add " + name));
 
         String pathPrefix = button.resolve().getAttribute("path").replaceAll("repeatable-add", "tool");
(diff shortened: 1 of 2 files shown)


src/test/java/plugins/AntPluginTest.java
```diff
@@ -21,6 +21,9 @@
     public static final String INSTALL_VERSION_1_8 = "1.8.4";
     public static final String INSTALL_NAME_1_8 = "ant_" + INSTALL_VERSION_1_8;
 
+    public static final String INSTALL_VERSION_1_10 = "1.10.5";
+    public static final String INSTALL_NAME_1_10 = "ant_" + INSTALL_VERSION_1_10;
```
olivergondza: Please inline these since they are used only once. There is no reason to have them public.

src/main/java/org/jenkinsci/test/acceptance/po/ToolInstallation.java
```diff
@@ -77,6 +77,12 @@ private static boolean hasUpdatesFor(final Jenkins jenkins, Class<? extends Tool
         final ConfigurablePageObject page = ensureConfigPage(jenkins);
 
         final String name = type.getAnnotation(ToolInstallationPageObject.class).name();
+        final Control expandButton = page.control(by.button(name + " installations..."));
+        try {
+            expandButton.click();
+        } catch (Exception e) {
```
olivergondza: Be more specific which exception type is safe to ignore here.

src/test/java/plugins/AntPluginTest.java
henryju: Changed them to private. I like to keep "magic" strings as constants.

src/main/java/org/jenkinsci/test/acceptance/po/ToolInstallation.java
henryju: Done

olivergondza:
Merged, thanks!

PR #495, opened 2019-03-19
Restore support of JUT before 2.168 where only plugin sha1 was checked

henryju:
Removing the sha1 element like before https://github.com/jenkinsci/acceptance-test-harness/commit/240e438e2a3ab719a7301ab112de470fd503572e#diff-e5a5651006d379f437e8f02dfa5866de
allows to bring back support of older Jenkins versions.


olivergondza:
Pardon my question, have you verified it works on Jenkins releases close to the 2.168 (as it is not covered by CI)?

henryju:
I have tested on 2.89.4 and 2.164.1

PR #497, opened 2019-03-20
[JENKINS-25046] Confirm that logging out trims stale cookies for this path

jsoref:
This is a test for jenkinsci/jenkins#3939

See [JENKINS-25046](https://issues.jenkins-ci.org/browse/JENKINS-25046).

Note: due to the construction of the test harness it does not appear to be particularly easy
to test that jenkins does not eat cookies for other paths

### DIFF
--- a/src/test/java/core/JenkinsDatabaseSecurityRealmTest.java
+++ b/src/test/java/core/JenkinsDatabaseSecurityRealmTest.java
@@ -33,9 +33,16 @@ import org.jenkinsci.test.acceptance.po.User;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
+import org.openqa.selenium.Cookie;
 
-import static org.hamcrest.Matchers.any;
-import static org.hamcrest.Matchers.equalTo;
+import java.util.Calendar;
+import java.util.Collections;
+import java.util.Date;
+import java.util.Random;
+
+import static org.hamcrest.Matchers.*;
+import static org.openqa.selenium.WebDriver.Options;
+
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.fail;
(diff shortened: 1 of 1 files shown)


src/test/java/core/JenkinsDatabaseSecurityRealmTest.java
```diff
@@ -59,6 +66,54 @@ public void setUp() {
...
+                .stream()
+                .forEach(i -> addSessionCookie(manager, "/", tomorrow));
+        addSessionCookie(manager, "/will-not-be-sent", tomorrow);
+
+        jenkins.logout();
+
+        StringBuilder builder = new StringBuilder();
+        int sessionCookies = 0;
+
+        final boolean debugCookies = false;
```
Wadeck: I think this debug "feature" should be activated by default (and so, removing the variable) as it allow us to better understand a failure in the test if that happens.

jsoref: I wasn't sure how to properly dump output without failing the test.

src/test/java/core/JenkinsDatabaseSecurityRealmTest.java
```diff
@@ -59,6 +66,54 @@ public void setUp() {
...
+        jenkins.login().doLogin(user.id(), PWD);
+
+        Calendar calendar = Calendar.getInstance();
+        calendar.add(Calendar.DAY_OF_YEAR, 1);
+        Date tomorrow = calendar.getTime();
+        byte[] array = new byte[7];
+        Options manager = driver.manage();
+        Collections.nCopies(8, 1)
+                .stream()
+                .forEach(i -> addSessionCookie(manager, "/", tomorrow));
```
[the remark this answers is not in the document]
jsoref: This was Cookie cutter code if you'll excuse the pun.

jsoref:
No, I didn't try to do it in core. I don't know enough about the test harness and the obvious place didn't show an obvious way to do it, so I went with a place where I found a way.

olivergondza:
Thanks for the contribution, but I do agree with Wadeck, to keep here only the tests that cannot be written in core. Are there any objections to closing this?

PR #498, opened 2019-04-08
[JENKINS-56928] Scroll to select to avoid overlap

MRamonLeon:
**See [[JENKINS-56928] Fix NodeLabelParameterPluginTest # run_on_online_slave_and_master_with_node_restriction. Fail with element not clickable because other obscures it](https://issues.jenkins-ci.org/browse/JENKINS-56928)**

Added a scroll to the select before clicking it to avoid the failure:

`Element <select name="allowedSlaves"> is not clickable at point (728.5,57.69999694824219) because another element <div class="tab config-section-activator config_build_triggers"> obscures it`


olivergondza:
Thanks for pointing this out. It must have got broken after my fix for warnings plugin in https://github.com/jenkinsci/acceptance-test-harness/commit/39cbea43b73d32a0912d613a41e08ba9b54aad1d.

I have exposed more straightforward API to do explicit scrolling and incorporated that into `Control#select()` for other calls to benefit from this: https://github.com/jenkinsci/acceptance-test-harness/pull/501

PR #499, opened 2019-04-08
java.time.Clock instead of obsolete SystemClock from selenium-support

vtitov:
Recent versions of selenium-support do not have earlier deprecated class org.openqa.selenium.support.ui.SystemClock 

Related selenium commit:
https://github.com/SeleniumHQ/selenium/commit/4c9379516cba2475fcb611c1204837f71b922916


MarkEWaite:
Should this pull request be closed so that we don't spend 8 hours of execution on ci.jenkins.io to evaluate a pull request that is not progressing?

PR #501, opened 2019-04-09
Explicitly scroll to select elements 

olivergondza:
Generalized version of https://github.com/jenkinsci/acceptance-test-harness/pull/498

@MRamonLeon @raul-arabaolaza

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Control.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Control.java
@@ -3,6 +3,7 @@ package org.jenkinsci.test.acceptance.po;
 import javax.annotation.Nullable;
 
 import org.apache.commons.lang3.StringUtils;
+import org.jenkinsci.test.acceptance.selenium.Scroller;
 import org.openqa.selenium.By;
 import org.openqa.selenium.JavascriptExecutor;
 import org.openqa.selenium.Keys;
@@ -234,6 +235,8 @@ public class Control extends CapybaraPortingLayerImpl {
      */
     public void select(String option) {
         WebElement e = resolve();
+        // Make sure the select is scrolled into view before interacting with its options that has got special handling by scroller.
+        //new Scroller().scrollIntoView(e, driver);
         findElement(e, by.option(option)).click();
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/Control.java
```diff
@@ -3,6 +3,7 @@
 import javax.annotation.Nullable;
 
 import org.apache.commons.lang3.StringUtils;
+import org.jenkinsci.test.acceptance.selenium.Scroller;
```
MRamonLeon: It is not needed if line 238 is commented out

MRamonLeon:
IIUC you don't do that on the select method, so I need to do that in my fix. But now I have the `scrollIntoView` method to do that easily. Is this your proposal?

src/main/java/org/jenkinsci/test/acceptance/selenium/Scroller.java
```diff
@@ -83,7 +87,11 @@ public void beforeChangeValueOf(WebElement element, WebDriver driver, CharSequen
         scrollIntoView(element, driver);
     }
 
-    private void scrollIntoView(WebElement e, WebDriver driver) {
+    /**
+     * The framework is expected to take care of the correct scrolling. When you are tempted to scroll from PageObjects
+     * or tests, there is likely a framework problem to be fixed.
```
MRamonLeon: With this fix, by adding the scroll inside the `select` method, we are scrolling generally with all the selects. According to this new comment, it should not be needed. If the framework manage the scroll, shouldn't we leave the fix only for the specific test failing? :thinking:

olivergondza: By *framework* I meant the ATH framework. Put differently, we would like the helper classes such as `Control` work without any hack in tests or page objects. Can you think of better formulation for the comment?

MRamonLeon: No, that's right. I got it now, thanks

MRamonLeon:
@olivergondza Could you close and reopen or whatever to see if it works, now the image is running again. Thank you.

olivergondza:
All tests for NodeLabelParameterPluginTest passed.

MRamonLeon:
The java 11 branch seems to be stuck, I've seen `Corrupted STDOUT by directly writing to native stream in forked JVM 1` in the console.

I've seen this message while fixing the restart of Jenkins in the PR #502. Maybe it could help to make this one runs to the end.

PR #503, opened 2019-04-11
Switch ATH container from Debian to Centos that distributes all the Javas

olivergondza:
https://issues.jenkins-ci.org/browse/JENKINS-56950

### DIFF
--- a/src/main/resources/ath-container/Dockerfile
+++ b/src/main/resources/ath-container/Dockerfile
@@ -12,7 +12,6 @@ RUN yum -y install \
         ImageMagick \
         iptables \
         firefox \
-        maven \
         unzip \
         tigervnc-server \
         which \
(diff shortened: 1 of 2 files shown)


src/main/resources/ath-container/Dockerfile
```diff
@@ -2,28 +2,23 @@
...
-    apt-get install -y \
+RUN yum -y install \
         curl \
+        docker \
         git \
-        imagemagick \
+        ImageMagick \
         iptables \
-        firefox-esr \
+        firefox \
```
olivergondza: Firefox in Centos is ESR.

src/main/resources/ath-container/set-java.sh
```diff
@@ -1,118 +1,23 @@
...
-xjc
-xjc.1.gz"
-
-# Set to the right java version (if the program doesn't exist in this java version, it's not changed)
-for program in $programs
-do
-    # Change the links without showing the choices nor the programs not found
-    echo $selection | update-alternatives --config $program > /dev/null 2> /dev/null
-done
+# For some reason, all tools from JDK are split to 2 groups named java and javac
```
olivergondza: It was the trickiest part to figure out, not sure this is specific to Centos: Individual binaries are attached to `java` or `javac` using the alternative's `--slave` relationship so setting these 2 commands are both sufficient and necessary.

olivergondza:
@MRamonLeon, Can you please have a look? The changes are substantial but things are running again.

olivergondza:
I have not tested attaching VNC client but the server appears to work good enough to run tests in. Let me see

olivergondza:
It does work for me. Connected as `vncviewer localhost:32818` and the firefox run from shell can be seen and interacted with. Re firefox closing immediately, mine complains about missing dbus but start just fine:

```
[ath-user@bf791df45d9f sources]$ firefox &
[1] 90
[ath-user@bf791df45d9f sources]$ Failed to open connection to "session" message bus: /usr/bin/dbus-launch terminated abnormally without any error message
Running without a11y support!
```

olivergondza:
Great, thanks! I am merging this to unblock recent PRs.

PR #504, opened 2019-05-02
Update page object for ssh-credentials 1.16

jvz:
This updates the page object so that it clicks the button hiding the textarea before trying to refer to the otherwise nonexisting input field.

Compare to https://github.com/jenkinsci/jenkins/blob/master/test/src/test/java/lib/form/SecretTextareaTest.java

@reviewbybees @jeffret-b @Wadeck

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_credentials/SshPrivateKeyCredential.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_credentials/SshPrivateKeyCredential.java
@@ -30,6 +30,7 @@ public class SshPrivateKeyCredential extends BaseStandardCredentials {
 
     public Direct selectEnterDirectly() {
         WebElement e = choose("Enter directly");
+        find(by.input("./@class='secret-update-btn'")).click();
         return new Direct(getPage(), e.getAttribute("path"));
     }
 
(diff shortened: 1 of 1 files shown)


jvz:
Test failure is infra-related.

olivergondza:
Thanks, does this work with the older versions of the plugin?

jvz:
The only difference is the clicking of the button. I'm not very familiar with this testing framework, so that's unknown if that would cause an error or not.

src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_credentials/SshPrivateKeyCredential.java
```diff
@@ -30,6 +30,7 @@ public SshPrivateKeyCredential enterDirectly(String privateKey) {
 
     public Direct selectEnterDirectly() {
         WebElement e = choose("Enter directly");
+        find(by.input("./@class='secret-update-btn'")).click();
```
Wadeck:
```suggestion
        find(by.className("secret-update-btn")).click();
```

Simpler alternative
Seems better also in terms of code complexity ([fieldXPath](https://github.com/jenkinsci/acceptance-test-harness/blob/156754ea959583cd43b0a48681f5ac9f37b572ef/src/main/java/org/jenkinsci/test/acceptance/ByFactory.java#L166-L169))

Wadeck:
I cannot push myself to this PR so here is a better version:

```
        WebElement button = getElement(By.className("secret-update-btn"));
        if (button != null) {
            // for ssh-credentials >= 1.16
            button.click();
        }
```

You need to import `import org.openqa.selenium.By;`. 

This proposal supports previous plugin version (which is required from my PoV) and use the simpler `className` selector.

Wadeck:
@olivergondza FYI, the new UI proposes a button that will display the text area. In the previous version the text area was directly visible. So yes, checking if the button exists is a requirement there.

jvz:
- [ ] Update code to work with or without the secret textarea button
- [ ] Use more direct element lookup for button

jvz:
Updated for backward compatibility @olivergondza

jvz:
More build infra issues. Rebuilding again.

jvz:
It appears as though there's no way for me to make this build green since master also fails.

PR #512, opened 2019-06-11
Use 1.9 as default $FORM_ELEMENT_PATH_VERSION

jglick:
Hoping to solve a subtle ATH regression in weeklies as described in https://github.com/jenkinsci/form-element-path-plugin/pull/8#discussion_r292442826.


jglick:
> older versions of Jenkins

There is no coherent system of indicating which versions of Jenkins core or plugins may be used with which versions of the ATH. In general I think the expectation is that ATH trunk is going to test the newest of everything. If you are testing an old version of software, use an old version of ATH.

olivergondza:
Uh, that is a pain. This is for the first time we needed to bump core dependency without having an easy way to keep things working for old cores (normally, when UI changes, we keep both ways in not to break downstream testsuites).

@jglick, how much trouble would it be to get the plugin to depend on older cores?

jglick:
> normally, when UI changes, we keep both ways

Or at least pretend to—there is no mechanical verification of that (so far as I know).

> how much trouble would it be to get the plugin to depend on older cores?

Trouble. See discussion in the upstream PR.

This kind of thing is why I do not think it is practical to advertise an arbitrary version of ATH as being applicable to arbitrary versions of software under test. [Pinning revisions](https://github.com/jenkinsci/jenkins/blob/71cd2b968087867a974b2c59824429792724a0c0/essentials.yml#L4) leads to more reproducible builds.

PR #515, opened 2019-07-11
Add test and POs for JCasC 

olivergondza:
Adding utility classes and a trivial test as this is needed from kerberos-sso plugin, too.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Control.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Control.java
@@ -298,7 +298,12 @@ public class Control extends CapybaraPortingLayerImpl {
             waitFor().until(() -> !spinner.isDisplayed());
             validationArea = control.findElement(by.xpath("./../../../following-sibling::div[2]"));
         } else {
-            validationArea = control.findElement(by.xpath("./../../following-sibling::tr/td[2]"));
+            // Wait for validation area to stop being <div></div>
+            validationArea = waitFor().until(() -> {
+                WebElement va = control.findElement(by.xpath("./../../following-sibling::tr/td[2]"));
+                String cls = va.findElement(by.xpath("./div")).getAttribute("class");
+                return (cls == null || cls.isEmpty()) ? null : va;
+            });
         }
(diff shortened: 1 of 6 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/Control.java
```diff
@@ -298,7 +298,12 @@ public FormValidation getFormValidation() {
             waitFor().until(() -> !spinner.isDisplayed());
             validationArea = control.findElement(by.xpath("./../../../following-sibling::div[2]"));
         } else {
-            validationArea = control.findElement(by.xpath("./../../following-sibling::tr/td[2]"));
+            // Wait for validation area to stop being <div></div>
+            validationArea = waitFor().until(() -> {
```
jtnord: @olivergondza this change seems to have caused `core.FormValidationTest#validate` to have failed since it was merged with
```
Unable to locate element: ./div
*** Element info: {Using=xpath, value=./div}
	at org.openqa.selenium.support.ui.FluentWait.until(FluentWait.java:260)
	at org.jenkinsci.test.acceptance.junit.Wait.until(Wait.java:90)
	at org.jenkinsci.test.acceptance.po.Control.getFormValidation(Control.java:332)
	at core.FormValidationTest.validate(FormValidationTest.java:42)
```

This is because the div is not shown if there is no validation error (which is being tested [here](https://github.com/jenkinsci/acceptance-test-harness/blob/828d2a5f1ec91ddd91598bd1642e75af7a2b704e/src/test/java/core/FormValidationTest.java#L43))

I think this looks like an attempt to reduce the flakyness by waiting for the validation to have occured before returning but I think that needs to be optional, or built into `FormValidation`  WDYT?

olivergondza: Good catch, my bad. I agree the waiting logic needs to be made more resilient to handle this. I also agree this belongs to `FormValidation` class. Tracking that as [JENKINS-59605](https://issues.jenkins-ci.org/browse/JENKINS-59605).

PR #517, opened 2019-07-16
Fix test failures related to createSshKeys

jvz:
Now that ssh-credentials uses f:secretTextarea, the private key value
cannot be viewed after saving it. This removes the prior assumption that
this value will always be viewable.

This fixes the CredentialsTest failures.

@reviewbybees @bmunozm @raul-arabaolaza @amuniz @jeffret-b


jvz:
@raul-arabaolaza oooh that's so naughty, decrypting the secret. Good suggestion, though.

raul-arabaolaza:
Yeah, I would not use thus aproach to perform the actions the test wants to validate. This is an acceptance test suite, but valudating the outcome of the UI interactions is another story...

jvz:
@raul-arabaolaza I updated the assertion to use the logic you suggested.

PR #518, opened 2019-07-20
Add a Release Drafter configuration

oleg-nenashev:
Once merged, I can enable Release Drafter for this repository.
Other tools like PCT already use it, so IMHO it would be nice to enable it, even if we do not categorize pull requests here.


olivergondza:
Interesting addition, let me add that

PR #519, opened 2019-07-30
[JENKINS-58071] Add support for new github-branch-source single repo selector

kshultzCB:
# Description
This pull request adds a way for automated tests to use the new single-repository URL when creating Multibranch projects from GitHub repositories.

The new UI looks like this:

image

These changes allow for someone to directly supply their HTTPS GitHub URL, in much the same way they would when running `git clone https://github.com/jenkinsci/...` from a terminal. This came in the form of [PR 236 to GitHub Branch Source](https://github.com/jenkinsci/github-branch-source-plugin/pull/236), which was merged and released as [GitHub Branch Source 2.5.5](https://github.com/jenkinsci/github-branch-source-plugin/tree/github-branch-source-2.5.5).

I have intentionally not removed the `@Ignore` annotation from the test itself. The ill-performing repository selector which was fixed in GHBS PR 236 is a big improvement, but was not the only reason this test can fail. We still have the problem of anonymous scanning of GitHub, and resultant wait times for branch indexing, which cause this test to time out as written. That hasn't been changed here.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/workflow_multibranch/GithubBranchSource.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/workflow_multibranch/GithubBranchSource.java
@@ -5,6 +5,7 @@ import org.jenkinsci.test.acceptance.po.Describable;
 import org.jenkinsci.test.acceptance.po.PageAreaImpl;
 import org.jenkinsci.test.acceptance.po.WorkflowMultiBranchJob;
 import org.jenkinsci.test.acceptance.plugins.workflow_shared_library.WorkflowSharedLibrary;
+import org.jvnet.hudson.test.Issue;
 import org.openqa.selenium.support.ui.Select;
 
 import java.util.concurrent.Callable;
@@ -19,6 +20,7 @@ public class GithubBranchSource extends BranchSource {
     public final Control owner = control("repoOwner");
     public final Control repository = control("repository");
     public final Control credential = control("credentialsId" /* >= 2.2.0 */, "scanCredentialsId");
+    public final Control repositoryUrl = control("repositoryUrl"); /* As of GHBS 2.5.5 */
 
     public GithubBranchSource(WorkflowMultiBranchJob job, String path) {
         super(job, path);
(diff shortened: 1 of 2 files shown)


olivergondza:
As the added method is not being used anywhere in the test, is it being used by a downstream testsuite? I am reluctant to add API that is not used...

kshultzCB:
I think the last attempt failed because a build agent was out of disk space. I'm going to do the close/reopen dance to trigger another test.

kshultzCB:
@olivergondza - it's not being used yet in this ATH, because the test which would use it is still set to `@Ignore`. The reason for that being, it can't run quickly enough as anonymous. The branch indexing just takes too long, because we intentionally throttle our API requests to GitHub in order to avoid getting rate limited. The test then times out, even with [the existing timeout of 20 seconds](https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/test/java/plugins/WorkflowMultibranchTest.java#L54).

This will be usable by the CloudBees verified ATH, though, once it is testing against GitHub Branch Source 2.5.5 and higher. That would need to be done in a different PR to that particular repository. This UI control needed to be in place before that can happen though.

vivek:
@kshultzCB Looks like there are test failures. PTAL.

kshultzCB:
@vivek - I don't think those test failures are related to this change. As @olivergondza already pointed out, this is just an API change for future testing with GitHub Branch Source >=2.5.5, so that we can use the new UI elements for single-repository.

The first sets of failures on run #1 look to have been infra problems, because of something running out of disk space:

image

In the second attempt, I see a hodgepodge of failures, none of which I can attribute directly to this change, which so far as I know isn't being used by anything yet. Seems to be fairly common - for example, the second attempt with this subject PR shows [82 failures](https://ci.jenkins.io/job/Core/job/acceptance-test-harness/view/change-requests/job/PR-519/2/). But similarly, this build of a totally different PR shows a similar number of failures - [83 of them](https://ci.jenkins.io/job/Core/job/acceptance-test-harness/view/change-requests/job/PR-499/13/testReport/).

vivek:
@kshultzCB yeah something else is going on. Your PR looks good 👍

vilacides:
@kshultzCB closing and reopening to retrigger the build and see if it works now!

src/main/java/org/jenkinsci/test/acceptance/plugins/workflow_multibranch/GithubBranchSource.java
```diff
@@ -43,6 +45,12 @@ public GithubBranchSource credential(final String credName) {
         return this;
     }
 
+    @Issue("JENKINS-58717")
```
olivergondza: This is annotation for tests methods.

olivergondza:
I am closing this in favor of https://github.com/jenkinsci/acceptance-test-harness/pull/541. Thanks for pointing out the problem exists!

PR #521, opened 2019-08-17
login does not wait for the user to be logged in

jtnord:
If you have a large delay on your Authentication then login will return
but the login process will not have completed.

The upshot is you must never use WebElement.click() for anything that
will be gauranteed to change the current page (URL)
https://blog.codeship.com/get-selenium-to-wait-for-page-load/

Whilst this is a form I tried changing click() to submit() but this
caused a worse issue and did not even work.  So for now add sime methods
that will wait until the page has changed (ie the current element will
be stale.  This will occur even if the same page is returned)

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Control.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Control.java
@@ -123,7 +123,7 @@ public class Control extends CapybaraPortingLayerImpl {
     public void clickAndWaitToBecomeStale(Duration timeout) {
         WebElement webElement = resolve();
         // webElement.submit() despite advertising it does exactly this just blows up :(
-        webElement.submit();
+        webElement.click();
         waitFor(webElement).withTimeout(timeout).until(Control::isStale);
     }
 
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Control.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Control.java
@@ -1,5 +1,6 @@
 package org.jenkinsci.test.acceptance.po;
 
+import java.time.Duration;
 import javax.annotation.Nullable;
 
 import org.apache.commons.lang3.StringUtils;
@@ -94,9 +95,38 @@ public class Control extends CapybaraPortingLayerImpl {
         check(resolve(), state);
     }
 
+    /**
+     * sends a click on the underlying element.
+     * You should not use this on any page where the click will cause another page to be loaded as it will not
+     * gaurantee that the new page has been loaded.
+     * @see #clickAndWaitToBecomeStale()
+     * @see #clickAndWaitToBecomeStale(Duration)
+     */
     public void click() {
         resolve().click();
     }
+
+
+    /**
+     * like click but will block for up to 30 seconds until the underlying web element has become stale.
+     * see https://blog.codeship.com/get-selenium-to-wait-for-page-load/
+     */
+    public void clickAndWaitToBecomeStale() {
+        clickAndWaitToBecomeStale(Duration.ofSeconds(30));
+    }
+
+    /**
+     * like click but will block until the underlying web element has become stale.
+     * see https://blog.codeship.com/get-selenium-to-wait-for-page-load/
+     * @param timeout the amount of time to wait
+     */
+    public void clickAndWaitToBecomeStale(Duration timeout) {
+        WebElement webElement = resolve();
+        // webElement.submit() despite advertising it does exactly this just blows up :(
+        webElement.submit();
+        waitFor(webElement).withTimeout(timeout).until(Control::isStale);
+    }
+
 
     /**
      * The existing {@link org.jenkinsci.test.acceptance.po.Control#set(String)}
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Login.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Login.java
@@ -1,6 +1,9 @@
 package org.jenkinsci.test.acceptance.po;
 
+import java.time.Duration;
+
 import org.junit.Assert;
+import org.openqa.selenium.WebElement;
 
 import static org.jenkinsci.test.acceptance.Matchers.hasInvalidLoginInformation;
 import static org.jenkinsci.test.acceptance.Matchers.loggedInAs;
@@ -28,7 +31,8 @@ public class Login extends PageObject {
     public Login doLogin(String user, String password){
         cUser.set(user);
         cPassword.set(password);
-        cLogin.click();
+        // for some reason submit it just bogus...
+        cLogin.clickAndWaitToBecomeStale();
         return this;
     }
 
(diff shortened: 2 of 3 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/Login.java
```diff
@@ -28,7 +31,8 @@ public Login(Jenkins jenkins) {
     public Login doLogin(String user, String password){
         cUser.set(user);
         cPassword.set(password);
-        cLogin.click();
```
jtnord: there are likely many other page object doing the same thing - but this was obvious causing internal tests to reliably fail falsely when testing unhappy paths.

It could probably be worth while sweeping through the code and looking for any "Submit" buttons (not apply) and changing them over, however I've not seen others reliably fail so keeping this change sufficiently focused.

It could also be worthwhile looking at why submit() is not working correctly, but I have spent a few days tracking this down and Selenium is not my best friend right now and this works.

src/main/java/org/jenkinsci/test/acceptance/po/Control.java
```diff
@@ -94,10 +95,39 @@ public void check(boolean state) {
...
     public void click() {
         resolve().click();
     }
 
+
+    /**
+     * like click but will block for up to 30 seconds until the underlying web element has become stale.
+     * see https://blog.codeship.com/get-selenium-to-wait-for-page-load/
+     */
+    public void clickAndWaitToBecomeStale() {
```
olivergondza: I do not quite like this new API (majority of ATH should switch to). We will need to give this deeper look to identify a cleaner path forward.

jtnord: there will be some things where a click does not reload the page.  doing this in click would be bad as then we are not mirroring the selenium API in the page object as well as we would need a clickButDoNotWaitForWhatever for these.
Happy to hear your thoughts

mikecirioli: @olivergondza @raul-arabaolaza  What can be done to move this forward?  I don't see that adding this API has any significant downside, or that there is a better way to implement this.
cc: @jglick

src/main/java/org/jenkinsci/test/acceptance/po/Login.java
```diff
@@ -40,7 +44,10 @@ public Login doLogin(String user, String password){
     public Login doLoginDespiteNoPaths(String user, String password){
         driver.findElement(by.name("j_username")).sendKeys(user);
         driver.findElement(by.name("j_password")).sendKeys(password);
-        driver.findElement(by.name("Submit")).click();
+        WebElement we = driver.findElement(by.name("Submit"));
+        // for some reason submit() is bogus
+        we.click();
```
mikecirioli: Is there a reason this is not using the `clickAndWaitToBecomeStale()`?

src/main/java/org/jenkinsci/test/acceptance/po/Control.java
```diff
@@ -94,10 +95,39 @@ public void check(boolean state) {
...
+
+    /**
+     * like click but will block until the underlying web element has become stale.
+     * see https://blog.codeship.com/get-selenium-to-wait-for-page-load/
+     * @param timeout the amount of time to wait
+     */
+    public void clickAndWaitToBecomeStale(Duration timeout) {
+        WebElement webElement = resolve();
+        // webElement.submit() despite advertising it does exactly this just blows up :(
+        webElement.submit();
```
mikecirioli: I don't get what you mean here - you say it blows up but yet it is still used?

jtnord: erm it shouldn't be used it was blowing up locally.  looks like I forgot to push some changes :-(

src/main/java/org/jenkinsci/test/acceptance/po/Login.java
jtnord: because this is using the `WebElement` directly and not a control.

src/main/java/org/jenkinsci/test/acceptance/po/Control.java
jtnord: `clickAndWaitForNewPage()` may be clearer...
maybe make this `@Restricted(NoExternalUse.class)` would be good enough, to move forward and allow for later refactoring.
currently the methods map the Selenium ones so changing click to not be the same would surprise some users (even if it is suprising that click does not do what many people may expect...

src/main/java/org/jenkinsci/test/acceptance/po/Control.java
mikecirioli: Testing locally, `submit()` appears to work for me.  I am still reading through the selenium implementation to see if it might be susceptible to the same issues, but in any case i think having `clickAndWaitToBecomeStale(...)` (or possibly an overloaded version of `click()`) still makes sense to add as an API because the default behavior in non-intuitive and indeterminately used in our page objects.

src/main/java/org/jenkinsci/test/acceptance/po/Control.java
batmat: > there will be some things where a click does not reload the page

I love the idea of doing this in `click()` or so @olivergondza, but James' point is very true too: I do not think every clicks reload the pages, so we would also create issue by changing click()'s behavior to wait for the page to become stale by default too isn't it?

Vlatombe: Shouldn't this new method be named `submit()` ? As it would pretty much match the (expected) behaviour of webElement.submit(), which seem to have some restrictions (https://www.quora.com/What-is-the-difference-between-click-and-submit-in-webdriver).

mikecirioli: I think naming it `submit()` might cause the same sort of confusion as changing the behavior of `click()`.  

Why not create the new method as `click(int timeout)` - this would let existing code continue to work (or flake) as it does today, and we can selectively start creating PR's to fix the instances where we would want to actually use the new method?

See #523

raul-arabaolaza:
So according to the blog

```Note that this solution only works for “non-JavaScript” clicks, i.e., clicks that will cause the browser to load a brand new page, and thus load a brand new HTML body element.```

Which means using this as generic `clickAndWaitToPageBeLoaded` is a bit dangerous IMO, although I believe the new implementation is going to be useful in many occasions. Putting a disclaimer about javascript based links should be enough IMO.

| `clickAndWaitForNewPage()` may be clearer...

Well, that would actually be no true in some cases... the element may disappear (being collapsed, removed from the page, changed via Javascript) without a new page being rendered, so I believe the current name is better.

I do not see why working tests have to be migrated to this new API, IMO this is a special case with `login()` not working in an optimal way (maybe login submit triggers some js that has to finish after the button is clicked for the real login to be actually triggered) and this new API is to be used only in that kind of occasions. No test should expect a new element to be there immediately after a click.

So I agree with this PR

mikecirioli:
@olivergondza  - what do you think?  And how can we move this forward?

olivergondza:
Thanks for all the inputs, folks!

I share Raul's point of view that changing too many places that are working fine in impractical. I will integrate this as something login specific to scratch the itch and test the new approach at the same time. Once we will feel the need to disambiguate "on-page" and "navigational" clicks on multiple places, we can address that later learning from this. (My guideline would be to stay away from explicit timeouts and using distinct names based on expectations).

olivergondza:
Actually, what particular tests is this impacting and how does the failure look like? The current failures for `JenkinsDatabaseSecurityRealmTest` are unrelated.

olivergondza:
Merged amended with https://github.com/jenkinsci/acceptance-test-harness/commit/5606d3bb3a645311711b74833d5d3a4fdc4d713d. Thanks!

PR #522, opened 2019-09-04
Update the maven version used in the ath docker image

mikecirioli:
Maven version `3.6.0` is no longer available from https://www.apache.org/dist/maven/maven-3/, this PR updates the `Dockerfile` to use `v3.6.1`


olivergondza:
Thanks!

PR #523, opened 2019-09-05
Add an alternative `click(Duration timeout)` method for pageobject controls

mikecirioli:
Add an alternative `click(Duration timeout)` method for pageobject controls that lets you wait up to a specified amount of time for a new page to load before continuing execution.  See #521 for additional details and discussion


mikecirioli:
@jtnord @olivergondza @batmat @Vlatombe comments?

olivergondza:
After reading that referenced article, are we confident that as soon as the old element becomes stale, the new page is completely loaded? I am not convinced.

PR #530, opened 2019-09-24
[JENKINS-59446] Writing non-flaky tests

amuniz:
Some general guidelines to write more stable tests.
This is what I got from the top of my head. Please feel free to suggest more!


docs/CONTRIBUTING.md
```diff
@@ -35,3 +35,47 @@ Test is determined fragile/broken when it is failing despite no defects are pres
...
+
+When writing acceptance tests using Selenium it is common to click on an element and then run some other action on the resulting page.
+Most of the time the page is updated really quick (specially if there is not a network request involved), but on slow infrastructure this can take some time so that the next test action after clicking could be checking on something which is not the expected.
+
+To avoid this, use the click and wait pattern:
+
+```
+WebElement clickableElement = ...;
+clickableElement.click();
+waitFor(by.id("my-element")).withTimeout(10);
```
jtnord: you probably want to wait for something rather then the element (which may be there but invisible)?

docs/CONTRIBUTING.md
```diff
@@ -35,3 +35,47 @@ Test is determined fragile/broken when it is failing despite no defects are pres
...
+... now run actions safely on my-element ...
```
jtnord:
```suggestion
// now run actions safely on my-element ...
```

docs/CONTRIBUTING.md
```diff
@@ -35,3 +35,47 @@ Test is determined fragile/broken when it is failing despite no defects are pres
...
+A few examples:
+
+1. Do not use full xpath expressions. DOM can change anytime.
+2. Avoid using parts of the DOM which are not under your control (a change in core or any other plugin can make your test to fail).
+3. Use paths generated by `form-element-path` plugin as much as possible (installed by default by the harness code). Look for the `path` attribute in DOM elements.
+
+### Use existing PageObjects
+
+If your test needs to interact with Jenkins Core (or other plugins UI), check if there is already a PageObject for it instead of writing the code by yourself.
+Existing PageObjects are usually well tested so they are possibly more stable than your newly added code.
```
jtnord: prefer to write a new page object for re-use than to hard code elements etc in your tests?

docs/CONTRIBUTING.md
raul-arabaolaza: Implicit timeout in `find` is one second and not using `ElasticTime`. Even using ElasticTime that is good when you are running always in the same infra (ie ci.jenkins.io) and is not very dynamic as you can put a default that makes sense most of the times, but if you run the ath as part of other builds (like core or some internal builds we have, or even locally) that default timeout starts to be less usefull and you start to get flaky results... And is harder to understand to newcomers as the timeout is "hidden"

I tend to prefer this `click and wait` approach even if not allways needed for various reasons:

* Is much more easier to people reading the test understand that we expect some element to sometimes take time to appear, this makes debugging easier and IMO test intent easier
* Is more robust when running in resource constrined environments as you can put a proper timeout for the concrete operation without affecting the global speed of all tests being running by overriding the implicit global timeout

So, this is a pattern I tend to use when I need to interact with an element that I know is not already present on the page I am already in. This does not mean I recommend to put a `waitFor` allways. If the element is suposed to already be on the page the default timeout is more than enough to deal with general loading times (although it probably should be switched to use `ElasticTime` )

docs/CONTRIBUTING.md
```diff
@@ -35,3 +35,47 @@ Test is determined fragile/broken when it is failing despite no defects are pres
...
+
+Sleeping blindly is a bad practice, it just makes tests execution time longer. 
+Use the [`Wait` class](https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/main/java/org/jenkinsci/test/acceptance/junit/Wait.java) as much as possible. `CapybaraPortingLayerImpl#waitFor` is an easy way to create instances of `Wait`.
+
+### Use robust selectors
+
+When selecting elements from the DOM, use robust selectors. 
+In general, try to avoid selectors that are known to change or are outside your control (they are not in your code). 
+
+A few examples:
```
uhafner: It would make sense to add a few links to good examples from our project.

raul-arabaolaza: And maybe also mention the [form-element-path-plugin](https://wiki.jenkins.io/display/JENKINS/Form+Element+Path+Plugin)

docs/CONTRIBUTING.md
amuniz: > Is this needed even when they are used?

Yeah, when using `find` the waitFor is there, but I'd like to explicitly point out that the wait is generally needed. If it's implicit in your current call, good, but test devs should know (and check) that those wait calls are needed.

Anyways, I'll make a note for the find methods which already use this wait pattern.

amuniz: > which may be there but invisible

Right, fixing.

amuniz: @jtnord `CapybaraPortingLayerImpl#waitFor(By, int)` is already checking for visibility.

docs/CONTRIBUTING.md
```diff
@@ -35,3 +35,47 @@ Test is determined fragile/broken when it is failing despite no defects are pres
...
+
+1. Do not use full xpath expressions. DOM can change anytime.
+2. Avoid using parts of the DOM which are not under your control (a change in core or any other plugin can make your test to fail).
```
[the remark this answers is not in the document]
amuniz: I disagree on this one.
Using ID is a bad practice, specially in a plugable system. IDs must be unique in the DOM, two elements with the same ID can cause a lot of problems.

IDs should be avoided.

docs/CONTRIBUTING.md
amuniz: > It would make sense to add a few links to good examples from our project.

I'm trying to avoid linking sources. Code evolves and links get obsolete quickly, specially if the link is to a specific line of code. We could use commit links, which are granted to stay the same, but then we would be linking to old sources quite quickly.

> And maybe also mention the form-element-path-plugin

It's mentioned. See bullet 3.

docs/CONTRIBUTING.md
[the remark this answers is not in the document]
uhafner: Well, defining an ID is something that someone chooses carefully, a collision will rarely occur. Otherwise we have problems to find the element in the DOM in the code under test as well.  Using complex selectors is also quite fragile (see 1.) - in the test code *and* in the code under test. So I understand this is a risk, but it is a price I am willing to pay in order to simplify things.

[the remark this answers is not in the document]
jtnord: So this implies we need a section on what makes a good id (prefixed with the plugin-id? for example to avoid any collisions).
Id's don;t work well on f:repeatable for example.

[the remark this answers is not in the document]
olivergondza: We are talking about two different things here:

- Using hardcoded ids on elements that can repeat on a page (ex.: shell build step), which is not a good idea, obviously.
- Ids accidentally colliding with something unrelated on the page.

As Antonio said, ids would not help in the first case and that is where `path` attribute comes to rescue. Though are we really suffering from the latter?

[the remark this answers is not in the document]
jtnord: >  Though are we really suffering from the latter

I do not know as I am generally not using IDs.  this was in response to @uhafner's comment.

it sounds like we are saying "do not use IDs for selecting elements" but Ulli was saying the opposite?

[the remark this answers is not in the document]
amuniz: Added a note about this.

PR #535, opened 2019-09-30
Strings are not urls

jtnord:
builds upon #534 and as such the only change is 
b98713e


jtnord:
(just changing base to get a simplified view of the changes)

PR #536, opened 2019-09-30
Save should use submit

jtnord:
builds upon #534 and #535 but does not need any of them (I was just too lazy to rebase as I have other WIP in the same tree)

6faa6c878248846c1227955175166a2b098a5262 is the isolated changes for this PR.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/ConfigurablePageObject.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/ConfigurablePageObject.java
@@ -29,7 +29,11 @@ import java.util.concurrent.Callable;
 import com.google.inject.Injector;
 
 import groovy.lang.Closure;
+
+import org.jenkinsci.test.acceptance.utils.ElasticTime;
 import org.openqa.selenium.By;
+import org.openqa.selenium.WebElement;
+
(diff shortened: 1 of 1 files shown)


jtnord:
Want to try this locally as well as the CI (so marking as WIP) because last time I tried to use `submit()` it all went Pete Tong.

jtnord:
this should fix an observed issue where the configuration had not been saved whilst an attempt to do another operation was performed (that relied on the configuration being saved first)

amuniz:
Can you rebase to get this change alone? (upstream PRs are merged)

jtnord:
no need to revise, just GitHub is not smart enough..n

jtnord:
@amuniz tada!

src/main/java/org/jenkinsci/test/acceptance/po/ConfigurablePageObject.java
```diff
@@ -126,8 +130,10 @@ public void ensureConfigPage() {
     public abstract URL getConfigUrl();
 
     public void save() {
-        clickButton("Save");
+        WebElement e = find(by.button("Save"));
+        e.submit();
```
amuniz: Seems weird to call `submit` on a generic `WebElement`, but if it works...

PR #537, opened 2019-10-04
Use clickAndWait where needed to reduce flakiness

EstherAF:
We're facing a lot of flakiness in our test suite, and after extensively diagnosing some of them, a cause we face a lot is that PO doesn't wait (or not enough time) after doing a click in a element that causes a page refresh, a full page loading, or requesting a partial HTML.
 
I've gone through all PO looking for examples of this, and I've change only those cases that I'm sure  causes a full page loading, so not waiting is a potential source of flakiness.

In those cases, use `Control#clickAndWaitToBecomeStale`


olivergondza:
I feel that this adds a lot of nasty code to page objects. That is nothing unheared of in ATH, but as long as we cannot make it clean, I prefer to individually add those only if we have an empiric evidence it is needed. IOW, are all of these fixing a failing / flaky test?

EstherAF:
We're facing a lot of flakiness in our test suite, and after extensively diagnosing some of them, a cause we face a lot is that PO doesn't wait (or not enough time) after doing a click in a element that causes a page refresh, a full page loading, or requesting a partial HTML.

We've considered what you propose, just fix what we know is causing failures, but it takes a lot more time investigate a single failing test, that proactively add some explicit waits in the cases I described before.

I've gone through all PO looking for examples of this, and I've change only those cases that I'm sure that are network dependent, so are potential source of flakiness.

I agree that some code is now more difficult to read, I can try a different approach to avoid this, but still add those waits. What do you think?

olivergondza:
The way I understand this, selenium is not doing a very good job in ensuring next page has loaded, so instead of somehow configuring it to do so, we have decided to re-implement the waiting as we have empirically learned selenium cannot do this. Correct?

If it is so, I would like to come up with another verb we would use next to `click` that would differ in its semantics: one would wait for page reload and the other one would not. Then, we would change the client coll depending on what is the result of the click (page reload, or just JS action on page). The benefit would be, all nasty details are encapsulated and not exposed to client code.

Perhaps we can keep the `click` to mean no-reload-expected, and use `submit`, `followLink` to mean navigate-somewhere. Thoughts?

@EstherAF, Can you please explicitly confirm this tweaks are reliably fixing those issues in your infra?

src/main/java/org/jenkinsci/test/acceptance/po/JenkinsDatabaseSecurityRealm.java
```diff
@@ -81,7 +81,7 @@ public Signup email(String mail) {
 
         public User signup(String name) {
             control(by.input("username")).set(name);
-            control(by.name("Submit")).click();
+            control(by.name("Submit")).clickAndWaitToBecomeStale();
```
EstherAF: Wait after submitting the form, similar case to the login (#521)

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
```diff
@@ -444,8 +445,10 @@ public ScmPolling pollScm() {
      */
     public void delete() {
         open();
-        clickLink("Delete Project");
+        WebElement deleteLink = find(by.link("Delete Project"));
+        deleteLink.click();
         confirmAlert(2);
+        waitFor(deleteLink).until(Control::isStale);
```
EstherAF: Same thing, causes a full reload page.
Wait until the "delete" button is stale. The webElement must be obtained before confirming the alert; if try to obtain it after, the element may not be already in the DOM.

EstherAF:
I've tried to clean up the code a little so it's more readable.

I agree that this naming is not the best :D




The discussion about the naming already started in the #521
My opinion is that, as long as this method is not exposed to the user (currently is at package level), is ok to show in its name this specific detail about 'waitingForSomething'. A name like `submit` or `followLink` can be better because we hide this implementation detail, but if users want to create their own PO this detail can be important, so I'm not sure...

For me, the best usage would be something like: `click().waitForPageReload()`, but we can't use this kind of syntax with current classes.

Another idea is `clickAndWaitToReload`.

EstherAF:
> @EstherAF, Can you please explicitly confirm this tweaks are reliably fixing those issues in your infra?

They seem to work fine @olivergondza. At least now we have all our tests running successfully with this changes. I can't claim that the solution is this PR, or I just got lucky and they are still flaky.

In any case, this changes are not breaking our suit test, and the public interfaces aren't changing.

The current build triggered by this PR is failing because some tests, but I've checked that they are already failing in master with the same cause

src/main/java/org/jenkinsci/test/acceptance/utils/pluginTests/SecurityDisabler.java
```diff
@@ -47,7 +50,11 @@ public void stopUsingSecurityAndSave() {
         Control use = control("/useSecurity");
         if(use.resolve().isSelected()) {
             use.click();
-            control("/Submit").click();
+            WebElement submit = control("/Submit").resolve();
+            // Would be better to use Control.clickAndWaitToBecomeStale,
+            // but it's not visible because it's still experimental
+            submit.click();
+            waitFor(submit).until(Control::isStale);
```
jtnord: can this abuse 'submit'?

src/main/java/org/jenkinsci/test/acceptance/po/View.java
```diff
@@ -58,7 +58,7 @@ public void delete() {
         configure();
         clickLink("Delete View");
         waitFor(by.button("Yes"));
-        clickButton("Yes");
+        control(by.button("Yes")).clickAndWaitToBecomeStale();
```
jtnord: can this use submit?

src/main/java/org/jenkinsci/test/acceptance/po/Slave.java
```diff
@@ -141,7 +144,7 @@ public void delete() {
             clickLink("Delete Slave");
         }
 
-        clickButton("Yes");
+        control(by.button("Yes")).clickAndWaitToBecomeStale();
```
jtnord: can this use submit?

src/main/java/org/jenkinsci/test/acceptance/po/Slave.java
```diff
@@ -126,10 +127,12 @@ public void markOnline(){
...
         if (isOnline()) {
             open();
+            // "Disconnect" reloads the page
             find(by.link("Disconnect")).click();
-            find(by.input("offlineMessage")).clear();
+            waitFor(by.input("offlineMessage")).clear();
             find(by.input("offlineMessage")).sendKeys(message);
-            clickButton("Yes");
+            // "Yes" reloads the page
+            control(by.button("Yes")).clickAndWaitToBecomeStale();
```
jtnord: can this use submit?

src/main/java/org/jenkinsci/test/acceptance/po/JenkinsDatabaseSecurityRealm.java
jtnord: can this use submit?

src/main/java/org/jenkinsci/test/acceptance/po/Folder.java
```diff
@@ -72,7 +72,8 @@ public ViewsMixIn getViews() {
     public void delete() {
         open();
         clickLink("Delete Folder");
-        waitFor(by.button("Yes")).click();
+        waitFor(by.button("Yes"));
+        control(by.button("Yes")).clickAndWaitToBecomeStale();
```
jtnord: can this use submit?

src/main/java/org/jenkinsci/test/acceptance/po/View.java
EstherAF: Wouldn't be exactly the same. In this form the 'Yes' button is not a submit button, which means that the click event is handled by some javascript function, which triggers submitting form.

Calling .submit() is the same as triggering a submit event on the form, so that javascript function won't be executed. If the javascript function is only submitting the form, it's the same, but if it's doing something else, we're causing a different behavior impossible to reproduce using only the UI

src/main/java/org/jenkinsci/test/acceptance/po/Slave.java
EstherAF: 'Yes' button is not a submit button

src/main/java/org/jenkinsci/test/acceptance/po/Slave.java
EstherAF: 'Yes' button is not of type submit

src/main/java/org/jenkinsci/test/acceptance/po/Folder.java
EstherAF: 'Yes' button is not a submit button

src/main/java/org/jenkinsci/test/acceptance/utils/pluginTests/SecurityDisabler.java
EstherAF: The same. This is not a submit type  button

src/main/java/org/jenkinsci/test/acceptance/po/JenkinsDatabaseSecurityRealm.java
EstherAF: This can be replaced, working on it

amuniz:
@EstherAF could you please not rebase and force push on PRs? It makes harder to follow changes.

amuniz:
Why did you add the "bug" label @oleg-nenashev? Is this PR introducing some bug?

EstherAF:
@oleg-nenashev can this be merged?

amuniz:
@EstherAF can you resolve conflicts please?

timja:
is this still needed?

timja:
@EstherAF is this still needed?

EstherAF:
No, I'm closing it.

PR #539, opened 2019-10-08
Increased timeout to avoid Selenium issues

imonteroperez:


imonteroperez:
> Thanks for the PR. @imonteroperez, can you please elaborate on what is the problem here and how is this improving the situation? Mind the [ElasticTime#seconds(long)](https://github.com/jenkinsci/acceptance-test-harness/blob/ad2c73cab611f0873a40ad555b7838bfd066826b/src/main/java/org/jenkinsci/test/acceptance/utils/ElasticTime.java#L58) converts seconds to milliseconds and telling the driver the value is in seconds basically let it to multiply the timeout by 1000. To 16 hours and 40 minutes based on my calculations.

Yes, we are experiencing this issue again: [JENKINS-50790](https://issues.jenkins-ci.org/browse/JENKINS-50790), and I didn't realize about seconds to milliseconds conversion that you mentioned, so I will test again with `TimeUnit.MILLISECONDS` increasing the timeout value, because we are getting a lot of Selenium issues with `tickPluginToInstall` method

imonteroperez:
@olivergondza After doing some testing, taking into account your feedback, I just propose to increase the timeout value. Updated PR title. Thanks!

PR #544, opened 2019-10-15
 Do not install form-element-path for tests annotated with @WithInstallWizard

olivergondza:
Early installation of form-element-path causes to install implied dependencies - 
in version that was detached. This can easily cause such plugin need to be updated
for other plugins to install correctly. While Plugin Manager is ready for this and
handle it well, Setup Wizard is reporting an issue and demand Jenkins to be restarted.

Note its page objects _can_ be made more robust to restart if needed, but since
it would happen every time, it would just obscure real issues this can detect. Not
to mention this is running Setup Wizard out of its expected mode of execution, as
it does not expect any plugins to be present effectively testing corner case rather
that main code paths.

Alternative [approach](https://github.com/jenkinsci/form-element-path-plugin/pull/8/files#r292442826) used in the past was bumping form-element-path every time
plugin gets detached from core so it never have any implied dependencies to cause
problems. But the downside is someone has to update that plugin every time there
is a plugin detached from core, and it prevents the plugin load for older cores
making older cores essentially untestable with ATH!

This is essentially taking a trade-off of never using form-element-path for wizard
tests over getting them broken every so often _and_ restricting ATH usage to quite
recent cores.

@amuniz @vilacides @jglick @jtnord, PTAL

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/controller/LocalController.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/controller/LocalController.java
@@ -128,12 +128,21 @@ public abstract class LocalController extends JenkinsController implements LogLi
         }
 
         LOGGER.info("Running with given plugins: " + Arrays.toString(pluginDir.list()));
+    }
 
-        try {
-            FileUtils.copyFile(formElementPathPlugin, new File(pluginDir, "path-element.hpi"));
-        } catch (IOException e) {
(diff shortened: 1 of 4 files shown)


olivergondza:
> why we don't just bump form-elementpath plugin to a recent version to bypass this?

Because such form-element-path would have to require core 2.184 (in order not to have that implied dpendency) effectively preventing ATH to work with any older core as majority of the tests would not work without form-element-path.

PR #545, opened 2019-10-23
Make sure we can run even the first PR build parallelized

olivergondza:
CC @amuniz


olivergondza:
Resolved by 061f8d8d16f69f9fea3a8ffd467f4c17049619e0

PR #554, opened 2020-01-04
Switch ATH Docker images to Maven 3.6.3

jglick:
3.6.1 is no longer available on the download site, so CI builds are failing. Amends #503 & #522.

Maybe better to use Maven Central as the download location? We know releases cannot be pulled from there.


oleg-nenashev:
I will merge it to recover CI in this repository. FYI @olivergondza

oleg-nenashev:
> Maybe better to use Maven Central as the download location? We know releases cannot be pulled from there.

In linger term it may make sense to just use multi-stage Docker build and pull maven from the image. Bonus point for switching to the official Maven-Centos image if it ever becomes available (CC @carlossg )

PR #555, opened 2020-01-23
WIP: Prototype nested ATH invocation from Jenkins plugin

olivergondza:
Trying to find a sane way to invoke ATH from `buildPlugin()`.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/utils/ContainerizedRunner.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/utils/ContainerizedRunner.java
@@ -1,0 +1,69 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) Red Hat, Inc.
+ *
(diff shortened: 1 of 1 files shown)


olivergondza:
Redoing this completely as the way from last week was as hideous as it gets: ATH tests in a single module was not invoked during the module build via surefire, but maven-exec-plugin invoked custom Main class distributed by ATH that launched ATH container, bind-mounted the same directory and executed the submodule this way leaving the file system populated with surefire results. While working, this was fragile and felt surprising.

Now this week, the approach is different: Moving to docker selenium containers maintained by selenium project that does not wrap entire ATH invocation (like ATH container does), but run only the browser, driver and selenium server in a container. This permit to execute ATH via surefire from maven out of ATH container (hence integrating nicely with multimodule maven build run via buildPlugin) as the selenium container is being launched by the testsuite.

Continues in https://github.com/jenkinsci/acceptance-test-harness/pull/556

PR #561, opened 2020-02-25
user local repository may different from default

olamy:
Signed-off-by: olivier lamy <olamy@apache.org>


src/main/java/org/jenkinsci/test/acceptance/utils/MavenLocalRepository.java
```diff
@@ -0,0 +1,29 @@
+package org.jenkinsci.test.acceptance.utils;
+
+import java.io.File;
+
+public class MavenLocalRepository
+{
```
olivergondza: Can you please adjust the formatting to match established conventions?

olivergondza:
Thanks!

PR #570, opened 2020-04-30
Better UC updated status check logic

raul-arabaolaza:
Previously the stale check was not working properly, if the button status check failed with something different than `StaleElementException` the error was swallowed by the outer catch and `true` was returned making the callers believe the UC finished updating when it was not.

Not apparent in ci.jenkins.io but in other testing infra with custom wars using more than one update center for example.


raul-arabaolaza:
CI failure is quite strange, I have tested this locally and it works perfectly... Maybe the reson is some change in the core? Checking

raul-arabaolaza:
Closing this until I have a working version

PR #571, opened 2020-05-05
JENKINS-56109 Adjust locator for table -> div layout changes

timja:
See https://github.com/jenkinsci/jenkins/pull/3895

There's another class on the checkbox now, and looking for exact match caused 2 smoke tests to fail


timja:
@olivergondza you able to take a look please?

PR #572, opened 2020-05-06
[JENKINS-62172] Fix invalid crumb issue on PluginManager.installPlugin().

aHenryJard:
Test that reproduce https://issues.jenkins-ci.org/browse/JENKINS-62172

```
java.io.IOException: Failed to upload plugin: HTTP/1.1 403 Forbidden
<html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=utf-8"/>
<title>Error 403 No valid crumb was included in the request</title>
</head>
<body><h2>HTTP ERROR 403 No valid crumb was included in the request</h2>
```


daniel-beck:
Obvious problem in https://github.com/jenkinsci/acceptance-test-harness/blob/1cc579eac831af27f444100c6bde73b342ec7c22/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java#L301-L305 that does not set a CSRF crumb.

olivergondza:
Thanks. Note the method is deprecated on purpose to signal this is not something to be done. IOW, tests should be running with the versions the test suite is configured to run, or bail out in case it makes no sense. What is your motivation to control the explicit version.

aHenryJard:
This is used on some use case to verify that having plugin known as "risky" display warning messages on the right place to the end user. If there is another way of uploading a custom plugin I can look at it also.

olivergondza:
Ah, so in case the risky is some specific old version, then it is ok to use the deprecated API and suppress and document the justification.

aHenryJard:
> It would be great if we can go on without checking it binaries like that.
Yes I agree.

@olivergondza where do you want me to document the justification in the `installPlugin` javadoc or in Jira ?

amuniz:
> where do you want me to document the justification in the installPlugin javadoc

PR #573, opened 2020-05-08
Fix overwrite message when copying directories on Windows

Akavith:
When copying ressources into the workspace of a jenkins test instance, the command "xcopy" is used on windows machines.

When copying ressources directyl the options"/E" and "/Y" are added to the command. When copying directories only the "/E" option is added.

This poses a problem when copying directories with ressources while having duplicate files already in the workspace.  That's because then a overwrite message pops up for the user which is normally suppressed with the "/Y" option. This results in failing the jenkins build.

Therefore the "/Y" option is also added to copying directories.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Job.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Job.java
@@ -279,7 +279,7 @@ public class Job extends TopLevelItem {
                     // TODO: Make it work for RemoteJenkinsController like in Unix (below)
                     throw new AssumptionViolatedException("Copying files in Windows is only supported if a LocalController is in use. Test will be skipped.");
                 }
-                addBatchStep("xcopy " + file.getAbsolutePath() + " %cd% /E");
+                addBatchStep("xcopy " + file.getAbsolutePath() + " %cd% /E /Y");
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/Job.java
```diff
@@ -279,7 +279,7 @@ public void copyFile(File file) {
                     // TODO: Make it work for RemoteJenkinsController like in Unix (below)
                     throw new AssumptionViolatedException("Copying files in Windows is only supported if a LocalController is in use. Test will be skipped.");
                 }
-                addBatchStep("xcopy " + file.getAbsolutePath() + " %cd% /E");
+                addBatchStep("xcopy " + file.getAbsolutePath() + " %cd% /E /Y");
```
uhafner, line 282: Would it make sense to extract ` /E /Y` as constant and reuse it as kind of `WINDOWS_XCOPY_OPTIONS` for the other occurrence in line as well? Or add a parameterized `xcopy` method that can be used in both places?

Akavith: a method has been added

PR #575, opened 2020-05-15
Plugins need to be "filtered" before selecting

amuniz:
* The Check now button is not a button anymore but a link.
* Search the plugin in the filter box before selecting (To accommodate https://github.com/jenkinsci/jenkins/commit/e97fe13fbdfd0f1a311862adafb36c10e621e536)


amuniz:
@olivergondza I think this should fix quite a bit of the current failures.

daniel-beck:
Sorry about that 😞

PR #577, opened 2020-05-27
[JENKINS-62481] Fix AuditTrailPluginTest

EstherAF:
The failure was happening because the PO `AuditTrailGlobalConfiguration` was using a wrong selector to find an HTML input (`GlobalPluginConfiguration`)

Replace invalid selector in `AuditTrailGlobalConfiguration`, with the current one.

The failure was reproduced with jenkins-core 2.204.5 and 2.222.3, and the fix has been validated against those lines too.


olivergondza:
Thanks!

PR #578, opened 2020-05-27
Be sure "Check Now" button action is finished before continue with a test

bmunozm:
Be sure "check for updates" is working consistently, clicking the button, and waiting until the page is refresh. This will avoid some flakiness in testing.

Tested in Java8 and Java11

cc/ @batmat @varyvol @raul-arabaolaza @imonteroperez

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
@@ -5,6 +5,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URISyntaxException;
+import java.time.temporal.ChronoUnit;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
```diff
@@ -95,23 +97,23 @@ public void checkForUpdates() {
         // The check now button is a form submit (POST) with a redirect to the same page only if the check is successful.
         // We use the button itself to detect when the page has changed, which happens after the refresh has been done
         // And we check for the presence of the button again
-        clickLink("Check now");
+        WebElement checkButton = find(by.link("Check now"));
+        checkButton.click();
         // The wait criteria is: we have left the current page and returned to the same one
-        waitFor(find(by.link("Check now"))).withTimeout(30, TimeUnit.SECONDS).until(webElement -> {
+        waitFor(checkButton).withTimeout(java.time.Duration.of(time.seconds(30), ChronoUnit.MILLIS)).until(webElement -> {
```
amuniz, line 103: Not sure what's the difference between this and the previous form. `clickLink` does exactly `find(by.link(..)` (which is what you are doing explicitly now), and then `waitFor(checkButton).withTimeout(java.time.Duration.of(time.seconds(30), ChronoUnit.MILLIS))` is doing exactly the same than `waitFor(find(by.link("Check now"))).withTimeout(30, TimeUnit.SECONDS)`.

Could you explain how this change reduces flakiness?

bmunozm: The difference is, in that case, we define the button before clicking it, so the button took into account is the one at the beginning. With this change, we are sure we are waiting for the correct one but not the "Check now" button that reappears after reloading the page

olivergondza: @bmunozm, hmm, good catch.

PR #580, opened 2020-06-04
Kill SanityChecker

olivergondza:
The reason this was introduced was to identify clearly erroneous pages and provide
better defect localization. But it failed to deliver as the number of failures it
generates is not justified by its advantages.


olivergondza:
Failed tests went from 65 to 20.

PR #583, opened 2020-06-16
[JENKINS-62691] Fix bug where scroller is stuck in infinite loop

fqueiruga:
See [JENKINS-62691](https://issues.jenkins-ci.org/browse/JENKINS-62691).

When scrolling to a form sticky button the scroller would get stuck in an infinite loop when the form buttons container touched the footer. It's better described on the issue.

#### Requesting reviewers

@EstherAF 
@bmunozm
@amuniz


amuniz:
This is fixing a corner case and does not seem to be (negatively) affecting other tests, will merge it today if no one tells me otherwise.

PR #585, opened 2020-06-17
Revert "Merge pull request #536 from jtnord/save-should-use-submit"

amuniz:
This reverts commit 95698c6eb8262f272da951b1e9d65bb02dd3481a, reversing
changes made to 26b15fd0763ceaef27ed1f07728e152e7df21ba0.

I'm observing erratic behaviour in ATHs, some `visit` calls are randomly failing to actually visit the page when it's run after a `save` call. I'm not sure what's really happening, but reverting this change I get consistent test runs.

/cc @olivergondza

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/ConfigurablePageObject.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/ConfigurablePageObject.java
@@ -29,11 +29,7 @@ import java.util.concurrent.Callable;
 import com.google.inject.Injector;
 
 import groovy.lang.Closure;
-
-import org.jenkinsci.test.acceptance.utils.ElasticTime;
 import org.openqa.selenium.By;
-import org.openqa.selenium.WebElement;
-
 import static org.hamcrest.CoreMatchers.*;
 import static org.hamcrest.MatcherAssert.*;
 import static org.jenkinsci.test.acceptance.Matchers.*;
@@ -130,10 +126,8 @@ public abstract class ConfigurablePageObject extends PageObject {
     public abstract URL getConfigUrl();
 
     public void save() {
-        WebElement e = find(by.button("Save"));
-        e.submit();
+        clickButton("Save");
         assertThat(driver, not(hasContent("This page expects a form submission")));
-        
     }
 
     public void apply() {


jtnord:
the problem is that not having this causes random failures...  So you have traded one random failure (something it sounds like is more easy to reproduce) with one that is a timing issue and is not so easy to reproduce.

amuniz:
> something it sounds like is more easy to reproduce

jtnord:
the alternative here is the `clickAndWaitToBecomeStale` idiom

See #537

jtnord:
Antonio shared the internal test results with me.   this does indeed look strange.  a call to `PageObject.open()` did not cause the URL to change in the browser and thus following calls to find an element just failed.  
This is completely bizarre and would be the bread and butter for Selenium so I am currently at a loss to explain it.  
a HAR from the browser and debug logs from selenium would certainly help, as there is some cases where we have things using `submit()` outside of this code.

olivergondza:
@amuniz, how does this fail specifically?

amuniz:
It shows in a variety of ways, usually with a "NoSuchElementException" when trying to find an element in the page (real cause is that a previous call to `open` or `visit` didn't work).
The code pattern of the failure is:
```
SomePageObject o = ....
o.save(); // the e.submit() call is inside this
AnotherPageObject o = ...
o.open(); // doesn't do anything
o.findBy... // error here
```
It does not always fail, it's randomly fails (some timing condition I guess).

amuniz:
[This test (from one of the latest runs)](https://ci.jenkins.io/job/Core/job/acceptance-test-harness/job/master/234/testReport/junit/plugins/AntisamyMarkupFormatterTest/java_8_split2___safeHtmlTest/) is an example. Looking at the code, it fails [here](https://github.com/jenkinsci/acceptance-test-harness/blob/14b831e4a5ff43f8eb03fe1029779785e510049f/src/test/java/plugins/AntisamyMarkupFormatterTest.java#L30) when trying to create a job, but if you look at the archived screenshot the browser didn't navigate to the job creation page, it stayed at the manage page after the [security save](https://github.com/jenkinsci/acceptance-test-harness/blob/14b831e4a5ff43f8eb03fe1029779785e510049f/src/test/java/plugins/AntisamyMarkupFormatterTest.java#L28).

amuniz:
Build finished with 44 tests failing, against the 211 failures in the last master build.
Merging.

PR #589, opened 2020-06-30
[JENKINS-62868] Delete one Git ancestry strategy test

MarkEWaite:
## [JENKINS-62868](https://issues.jenkins-ci.org/browse/JENKINS-62868) Git ancestry strategy well tested in plugin tests

Table to div UI changes probably broke this UI test.  Already have strong coverage for ancestry strategy in the git plugin tests.  There are two other build chooser strategy tests in this suite.  More valuable to spend our time on other areas.

### DIFF
--- a/src/test/java/plugins/GitPluginTest.java
+++ b/src/test/java/plugins/GitPluginTest.java
@@ -313,40 +313,6 @@ public class GitPluginTest extends AbstractJUnitTest {
     }
 
     @Test
-    public void ancestry_strategy_to_choose_build() {
-        final String TEST_BRANCH = "testBranch";
-
-        GitRepo repo = buildGitRepo();
-        repo.createBranch(TEST_BRANCH);
-        repo.changeAndCommitFoo("Commit1 on master");
-        repo.checkout(TEST_BRANCH);
-        repo.changeAndCommitFoo("commit1 on " + TEST_BRANCH);
-        String sha1 = repo.getLastSha1();
-        repo.changeAndCommitFoo("commit2 on " + TEST_BRANCH);
-
-        repo.transferToDockerContainer(host, port);
-
-        job.useScm(GitScm.class)
-                .url(repoUrl)
-                .branch("")
-                .credentials(USERNAME)
-                .chooseBuildStrategy("Ancestry", 1, sha1);
-
-        job.save();
-        Build b = job.startBuild();
-        b.shouldSucceed();
-        // TODO Multiple selected branches create multiple builds, these should also be verified
-
-        assertThat(
-                b.getConsole(),
-                Matchers.containsRegexp(
-                        "Checking out Revision .* \\(origin/"+TEST_BRANCH+"\\)",
-                        Pattern.MULTILINE
-                )
-        );
-    }
-
-    @Test
     public void inverse_strategy_to_choose_build() {
         final String BRANCH_NAME = "secondBranch";
 


MarkEWaite:
The CI job has completed the execution of GitPluginTest successfully on both Java 8 and Java 11.  Other tests are failing, but this deletion resolves the one failure in GitPluginTest.

MarkEWaite:
@fqueiruga could you review this pull request and offer your opinion if it is acceptable to delete the test?  I'm fine if someone else wants to invest the effort to repair the test, but for me, the test does not add enough value to spend the effort to update it for the latest UI changes.

MarkEWaite:
@fqueiruga the job ran to completion (over 8 hours on two highmem machines) with no failures in the GitPluginTest.  I recommend merging this deletion of one test.

MarkEWaite:
@MRamonLeon @bmunozm @raul-arabaolaza @varyvol @rsandell @alecharp @amuniz Could you help to get this merged to remove this test failure. Thanks!

fqueiruga:
Just to confirm I understand it correctly, this piece of UI is already covered on a unit test that runs on the PCT then?

Also, shouldn't it be worrying that it fails on the ATH but not on the unit test? If that's the case maybe we should examine the failure to see if we should improve the unit test.

uhafner:
> Just to confirm I understand it correctly, this piece of UI is already covered on a unit test that runs on the PCT then?

Unit tests never cover the UI. 
(Some integrations tests may cover UI elements as well but this is rarely used in practice and should be avoided in favor of ATH tests.) 

I think the problem here is that the configuration of the page object is broken. And since we already have way too much ATH tests I think it is save to clean them up.

@MarkEWaite Wouldn't it then also make sense to remove the page object setter if it is not working anymore? (And should we also check other areas in the ATH for occurrences of the same UI pattern that does not work anymore?)

src/test/java/plugins/GitPluginTest.java
```diff
@@ -312,40 +312,6 @@ public void sparse_checkout() {
...
-        String sha1 = repo.getLastSha1();
-        repo.changeAndCommitFoo("commit2 on " + TEST_BRANCH);
-
-        repo.transferToDockerContainer(host, port);
-
-        job.useScm(GitScm.class)
-                .url(repoUrl)
-                .branch("")
-                .credentials(USERNAME)
-                .chooseBuildStrategy("Ancestry", 1, sha1);
```
uhafner, line 333: I think if `chooseBuildStrategy ` does not work anymore we should remove it. Otherwise someone else may use it in the future.

MarkEWaite: I think that `chooseBuildStrategy` works in some cases.  If `chooseBuildStrategy` does not work anymore in all cases, then there would be two other failing tests in this source file, since there are two places where it is called.  Something does not work in this specific case.

I can't run the acceptance tests reliably on my development computer.  I believe that the failing test is of relatively low value in its ability to help us find bugs.  I think it is more valuable to delete that test and have it not distract us with a failing test than to spend the time to determine why it is failing.  I don't see any failure in this area when I test interactively.  The other two test methods that use `chooseBuildStrategy` are not failing in this source file.

MarkEWaite:
> @MarkEWaite Wouldn't it then also make sense to remove the page object setter if it is not working anymore? (And should we also check other areas in the ATH for occurrences of the same UI pattern that does not work anymore?)

I'm not skilled in user interface test automation.  I'm also not a big fan of testing the git plugin through the user interface.  Those two things combine to make me unwilling to remove the page object setter.

I submitted this pull request hoping that it would remove a barrier to those who get value from the acceptance test harness and who maintain the acceptance test harness.  Unfortunately, I'm not one of those people.  I spent 3 or 4 hours trying (and failing) to make the tests in this source file run reliably on my Debian machine.  I'm not willing to spend significantly more time on this test.

If others would like to continue evolving and improving this test, they are welcome to do so.  I'd rather spend the user interface time that I can allocate for the git plugin on the table to div migration.

amuniz:
Plugin maintainers are (should) be the ones deciding what to do with ATHs related to their plugins, @MarkEWaite is the main maintainer of `git-plugin`, so I think it's ok to remove this test if he thinks it is the right thing to do.

I'll be merging this soon unless someone strongly disagree.

uhafner:
See https://github.com/jenkinsci/acceptance-test-harness/pull/588/commits/36d49f5d3b51c91361156ca54e61745c4add8e30

PR #591, opened 2020-07-02
JENKINS-62870 SAML Plugin test for tables to divs form rework

schottsfired:
This PR is a dependency of https://github.com/jenkinsci/jenkins/pull/3895.


olivergondza:
Thanks, will this fixed version work for the jenkins core before the fix? We strive to make the ATH to work different version of Jenkins core.

schottsfired:
Hi @olivergondza, this PR makes the SAML Plugin tests pass on a Jenkins core version built from https://github.com/jenkinsci/jenkins/pull/3895. I don't think it will work on a core version that doesn't include 3895. Do you want me to modify this PR so that tests pass on any version?

olivergondza:
@schottsfired, yes, that is what I meant. That way older ATH can be used with newer core and wise versa.

schottsfired:
@olivergondza that makes total sense! My latest commit should handle both versions.

PR #592, opened 2020-07-14
Fix OwnershipPluginTest for LTS

raul-arabaolaza:
Runs with recent LTS or even `master` results in failures because right now `getText` on the node page does not return the `Primary` literal even if it is visible in the UI, I guess some security hardening or the expected behaviour, in any case I prefer to change the assert in the test that change the behaviour of `getText` as that could break other tests.

See failure https://ci.jenkins.io/job/Core/job/acceptance-test-harness/job/master/241/testReport/plugins/OwnershipPluginTest/java_8_split6___explicitly_set_ownership/

### DIFF
--- a/src/test/java/plugins/OwnershipPluginTest.java
+++ b/src/test/java/plugins/OwnershipPluginTest.java
@@ -111,7 +111,7 @@ public class OwnershipPluginTest extends AbstractJUnitTest {
     }
 
     private Matcher<ContainerPageObject> ownedBy(final User user) {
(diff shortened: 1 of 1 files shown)


raul-arabaolaza:
Seems to be working https://ci.jenkins.io/job/Core/job/acceptance-test-harness/job/PR-592/2/testReport/plugins/OwnershipPluginTest/ and it does not require more reviews on my side. So IMO it can be merged

PR #595, opened 2020-07-24
upgrade selenium to 3.141.59

jtnord:
Upgrades selenium to `3.141.59` 

I don't actually need this - but in the process of actually trying to work out what the issue I was observing was I tried using the latest - turns out the issue was with the docker container & env I was using to run firefox.

so no urgency, but would probably need doing at some point so filing so that the work does not have to be duplicated.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/utils/ElasticTime.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/utils/ElasticTime.java
@@ -59,7 +59,16 @@ public class ElasticTime {
     }
 
     public long milliseconds(long ms) {
-        double actualSeconds = concurrency * factor;
+        double actualSeconds = getSlowDownFactor();
         return Math.round(ms * actualSeconds);
+    }
+    
+    /**
+     * Get the factor by which we slow down time.
+     * Default is {@code 1.0} (no difference). Use {@code >1} in case of slower environment, {@<1} in case of faster one.
+     * @return the factor by which we slow down time.
+     */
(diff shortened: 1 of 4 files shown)


olivergondza:
Thanks, @jtnord! Can you also update https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/main/java/org/jenkinsci/test/acceptance/FallbackConfig.java#L132-L134 so the containers match the version?

jtnord:
> Can you also update...

all this time I am manually launching a container and setting `BROWSER=remote-webdriver-firefox` and `JENKINS_LOCAL_HOSTNAME` , never even knew that was an option   :-)

src/main/java/org/jenkinsci/test/acceptance/utils/ElasticTime.java
```diff
@@ -59,7 +59,16 @@ public long seconds(long secs) {
     }
 
     public long milliseconds(long ms) {
-        double actualSeconds = concurrency * factor;
+        double actualSeconds = getSlowDownFactor();
         return Math.round(ms * actualSeconds);
     }
+    
+    /**
+     * Get the factor by which we slow down time.
```
amuniz, line 67: _If my calculations are correct, when this baby hits eighty-eight miles per hour… you’re gonna see some serious shit._

**Doc, “Back To The Future”**

olivergondza:
@jtnord, it is relatively recent addition...

olivergondza:
Thanks, James!

PR #597, opened 2020-08-01
JENKINS-62871 Fix ssh slaves with tables to divs

timja:
With tables to divs the page is longer, `moveToElement` doesn't seem to work with our credentials dialog popup.

This fixes it

cc @fqueiruga @olivergondza

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_credentials/SshCredentialDialog.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_credentials/SshCredentialDialog.java
@@ -6,13 +6,11 @@ import org.jenkinsci.test.acceptance.po.PageAreaImpl;
 import org.jenkinsci.test.acceptance.po.PageObject;
 import org.jenkinsci.test.acceptance.selenium.Scroller;
 import org.openqa.selenium.By;
-import org.openqa.selenium.JavascriptExecutor;
 import org.openqa.selenium.NoSuchElementException;
 import org.openqa.selenium.StaleElementReferenceException;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.interactions.Actions;
 
-import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 
 /**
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_credentials/SshCredentialDialog.java
```diff
@@ -61,6 +62,12 @@ private void findAndPerformClick(final By addSubmitButtonLocator) {
         final WebElement addSubmitButton = find(addSubmitButtonLocator);
 
         final Actions builder = new Actions(driver);
+
+        // move to element doesn't seem to be working properly
+        ((JavascriptExecutor) driver).executeScript(
```
olivergondza, line 67: IIRC, `moveToElement` was introduced to address similar problem in the past. Would reverting back to plain old button clicking (that invokes `Scroller`) work?

If not, note there is `#executeScript(String, Object...)` you can invoke here.

timja: > IIRC, moveToElement was introduced to address similar problem in the past. Would reverting back to plain old button clicking (that invokes Scroller) work?

Well it's not working, I think it's to do with the inner scrollbar on the credentials dialog...


> If not, note there is #executeScript(String, Object...) you can invoke here.

Is that on the parent? Otherwise that's what I'm calling here

fqueiruga: AFAIK scroller is an ATH util to scroll on a page until an element is found. It has had some weird bugs in the past. It is used for example here https://github.com/jenkinsci/acceptance-test-harness/blob/10d7dae64ae445a584827c5d96e516fd09cafeb5/src/main/java/org/jenkinsci/test/acceptance/po/Control.java#L269.

timja: I've changed it to use scroller, verified on tables to divs and 2.250 that both versions pass

olivergondza: Yeah, `Actions` seems to bypass those drive listeners  so the `Scroller` is not invoked automatically.

PR #608, opened 2020-09-23
Attempt to fix HarRecorderTest

Vlatombe:
Checking the tests execution, seems like there was a problem I didn't catch up before when executing `HarRecorderTest`


schottsfired (approved):

timja (approved):
the test passed

PR #609, opened 2020-09-24
Trust all servers to avoid SSL errors in logs

schottsfired:
Without this property on the proxy, one will see errors in their logs when the server presents an insecure cert:

image

```
2020-09-23 15:26:07,161 [LittleProxy-0-ProxyToServerWorker-0] ERROR  (ProxyToServerConnection.java:465) org.littleshoot.proxy.impl.ProxyToServerConnection - (DISCONNECTED) [id: 0xe2dbfcdd, L:0.0.0.0/0.0.0.0:51215 ! R:/35.199.14.121:443]: Caught an exception on ProxyToServerConnection
io.netty.handler.codec.DecoderException: javax.net.ssl.SSLHandshakeException: General SSLEngine problem
```


timja (approved):
seems ok given the context

Vlatombe (approved):

PR #612, opened 2020-10-26
[JENKINS-64038] environment variables are Strings

jtnord:
The initial build is not splitting correctly.

this attempts to fix that based on a guess that the if is failing as it
is comparing a String to an int.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,7 +8,7 @@ for (int i = 0; i < (BUILD_NUMBER as int); i++) {
 def branches = [:]
 
 def splits
-if (env.BUILD_NUMBER == 1) {
+if (env.BUILD_NUMBER == '1') {
     node() { // When there are no previous build, we need to estimate splits from files which require workspace
         checkout scm
         splits = splitTests estimateTestsFromFiles: true, parallelism: count(10)


jtnord:
image

this was the cause of the first batch not splitting.

This still has the issue that the excludes are not correctly taken into account.

Jenkinsfile
```diff
@@ -8,7 +8,7 @@ for (int i = 0; i < (BUILD_NUMBER as int); i++) {
 def branches = [:]
 
 def splits
-if (env.BUILD_NUMBER == 1) {
+if (env.BUILD_NUMBER == '1') {
```
jglick, line 11: or
```suggestion
if (BUILD_NUMBER == '1') {
```

jglick (approved):

jtnord:
marked as do not merge until we can also find out why the splits are not being excluded in surefire.
if merged before that all first builds of PRs will squat on the nodes running all tests 20 times (10 times for each of Java8/java11)

jtnord:
removed the do-not-merge the splitting seems to be working ok "most of the time".  attempts to work out why it was not always working are inconclusive.
so lets just merge this and keep an eye on things.

PR #617, opened 2020-10-28
[JENKINS-64038] we need to split the tests when there are no reference builds

jtnord:
https://issues.jenkins-ci.org/browse/JENKINS-64038

if previous builds are aborted or failed and the `splitTests` step has
no reference we still need to split the tests - so fallback in any case
there is not a previously completed non failed build.

this is an obvious improvement over #612

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,7 +14,7 @@ if (currentBuild.previousNotFailedBuild == null) {
         splits = splitTests estimateTestsFromFiles: true, parallelism: count(10)
     }
 } else {
-    echo "reference build should be : ${currentBuild.previousNotFailedBuild}"
+    echo "reference build should be : ${currentBuild.previousNotFailedBuild.projectName}:${currentBuild.previousNotFailedBuild.number}, with state ${currentBuild.previousNotFailedBuild.result}"
     splits = splitTests count(10)
 }
 for (int i = 0; i < splits.size(); i++) {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,7 +12,7 @@ def needSplittingFromWorkspace = true
 for (build = currentBuild.previousCompletedBuild; build != null; build = build.previousCompletedBuild) {
     if (build.resultIsBetterOrEqualTo("UNSTABLE")) {
         // we have a reference build
-        echo "not splitting from workkspace, reference build should be : ${currentBuild.previousNotFailedBuild.projectName}:${currentBuild.previousNotFailedBuild.number}, with state ${currentBuild.previousNotFailedBuild.result}"
+        echo "not splitting from workkspace, reference build should be : ${build.previousNotFailedBuild.projectName}:${build.previousNotFailedBuild.number}, with state ${build.previousNotFailedBuild.result}"
         needSplittingFromWorkspace = false
         break
     }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,15 +6,24 @@ for (int i = 0; i < (BUILD_NUMBER as int); i++) {
 }
 
 def branches = [:]
-
 def splits
-if (currentBuild.previousNotFailedBuild == null) {
+def needSplittingFromWorkspace = true
+
+for (build = currentBuild.previousCompletedBuild; build != null; build = build.previousCompletedBuild) {
+    if (build.resultIsBetterOrEqualTo("UNSTABLE")) {
+        // we have a reference build
+        echo "not splitting from workkspace, reference build should be : ${currentBuild.previousNotFailedBuild.projectName}:${currentBuild.previousNotFailedBuild.number}, with state ${currentBuild.previousNotFailedBuild.result}"
+        needSplittingFromWorkspace = false
+        break
+    }
+}
+
+if (needSplittingFromWorkspace) {
     node() { // When there are no previous build, we need to estimate splits from files which require workspace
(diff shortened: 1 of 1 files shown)


timja (approved):

Jenkinsfile
```diff
@@ -8,7 +8,7 @@ for (int i = 0; i < (BUILD_NUMBER as int); i++) {
 def branches = [:]
 
 def splits
-if (env.BUILD_NUMBER == '1') {
+if (currentBuild.previousNotFailedBuild != null) {
```
jtnord, line 11: OOPS!
```suggestion
if (currentBuild.previousNotFailedBuild == null) {
```

Jenkinsfile
```diff
@@ -8,7 +8,7 @@ for (int i = 0; i < (BUILD_NUMBER as int); i++) {
 def branches = [:]
 
 def splits
-if (env.BUILD_NUMBER == '1') {
+if (currentBuild.previousNotFailedBuild == null) {
```
jtnord, line 11: well this is disappointing...

image

jtnord: so `previousNotFailedBuild` is returning builds in progress :(

jtnord: https://github.com/jenkinsci/jenkins/blob/4ff3e8dbf7d3d0cd0619ffefc60d951b3e654c70/core/src/main/java/hudson/model/Run.java#L947-L952

need to iterate on the result of getPreviousCompletedBuild()

Jenkinsfile
```diff
@@ -6,9 +6,19 @@ for (int i = 0; i < (BUILD_NUMBER as int); i++) {
...
 def branches = [:]
-
 def splits
-if (env.BUILD_NUMBER == '1') {
+def needSplittingFromWorkspace = true
+
+for (build = currentBuild.previousCompletedBuild; build != null; build = build.previousCompletedBuild) {
+    if (build.resultIsBetterOrEqualTo("UNSTABLE")) {
+        // we have a reference build
+        echo "not splitting from workkspace, reference build should be : ${build.previousNotFailedBuild.projectName}:${build.previousNotFailedBuild.number}, with state ${build.previousNotFailedBuild.result}"
```
MarkEWaite, line 15: When I tried to run this fragment in a pipeline job that I created, it failed with an error that the build.previousNotFailedBuild was null.  Does the echo need some form of safeguards to handle cases where there is no previous not failed build?
```suggestion
        echo "not splitting from workkspace, reference build should be : ${build.previousNotFailedBuild.projectName}:${build.previousNotFailedBuild.number}, with state ${build.previousNotFailedBuild.result}"
```

MarkEWaite (commented):
Seemed to fail the build when there was no previousNotFailedBuild.

```
 java.lang.NullPointerException: Cannot get property 'projectName' on null object
12:35:00  	at org.codehaus.groovy.runtime.NullObject.getProperty(NullObject.java:60)
12:35:00  	at org.codehaus.groovy.runtime.InvokerHelper.getProperty(InvokerHelper.java:174)
12:35:00  	at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.getProperty(ScriptBytecodeAdapter.java:456)
12:35:00  	at org.kohsuke.groovy.sandbox.impl.Checker$7.call(Checker.java:355)
12:35:00  	at org.kohsuke.groovy.sandbox.impl.Checker.checkedGetProperty(Checker.java:357)
12:35:00  	at com.cloudbees.groovy.cps.sandbox.SandboxInvoker.getProperty(SandboxInvoker.java:29)
12:35:00  	at com.cloudbees.groovy.cps.impl.PropertyAccessBlock.rawGet(PropertyAccessBlock.java:20)
12:35:00  	at WorkflowScript.run(WorkflowScript:6)
12:35:00  	at ___cps.transform___(Native Method)
```

My job definition was:

```
def needSplittingFromWorkspace = true

for (build = currentBuild.previousCompletedBuild; build != null; build = build.previousCompletedBuild) {
    if (build.resultIsBetterOrEqualTo("UNSTABLE")) {
        // we have a reference build
        echo "not splitting from workkspace, reference build should be : ${build.previousNotFailedBuild.projectName}:${build.previousNotFailedBuild.number}, with state ${build.previousNotFailedBuild.result}"
        needSplittingFromWorkspace = false
        break
    }
}

echo "needSplittingFromWorkspace is ${needSplittingFromWorkspace}"
```

Jenkinsfile
jtnord: urgh... nope...

```suggestion
        echo "not splitting from workkspace, reference build should be : ${build.projectName}:${build.number}, with state ${build.result}"
```

timja (approved):

jtnord:
@MarkEWaite I have made a change to fix the NPE and tested in a pipeline elsewhere with the following (and a few builds)

```groovy
def needSplittingFromWorkspace = true

for (build = currentBuild.previousCompletedBuild; build != null; build = build.previousCompletedBuild) {
    if (build.resultIsBetterOrEqualTo("UNSTABLE")) {
        // we have a reference build
        echo "not splitting from workspace, reference build should be : ${build.projectName}:${build.number}, with state ${build.result}"
        needSplittingFromWorkspace = false
        break
    }
}

if (needSplittingFromWorkspace) {
    echo "needs splitting!"
} else {
    echo "does not need splitting!"
}
```

PR #618, opened 2020-11-04
Fix a flakiness observed when using master.runScript()

jtnord:
The script console is added by Javascript and not completely ready with
the page load.
Therefore we need to wait for the parts we want to manipulate to be
ready.

Fixes a flaky test observed when using master.runStript() with the following failure:

```
Unable to locate element: //*[@path='/script']

```


aHenryJard (approved):

timja (approved):
looks sensible

jtnord:
verified failures are not new :(

PR #619, opened 2020-11-06
deflake `Control.click` and `Control.selectDropdownMenu`

jtnord:
fixes various flakes.

`click()` was flaky when the WebElement had to be scrolled into view due to obsuring by the "Save" and "Apply" buttons (so we waitfor the click to happen).

the dropDown selector was flaky and would sometimes not be visible - so also changed to using waits (and update the code that was sleeping to use wait to make the tests faster where possible)

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Control.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Control.java
@@ -1,13 +1,15 @@
 package org.jenkinsci.test.acceptance.po;
 
 import java.time.Duration;
+import java.util.concurrent.TimeUnit;
+
 import javax.annotation.Nullable;
 
 import org.apache.commons.lang3.StringUtils;
 import org.jenkinsci.test.acceptance.selenium.Scroller;
 import org.openqa.selenium.By;
+import org.openqa.selenium.ElementClickInterceptedException;
 import org.openqa.selenium.JavascriptExecutor;
-import org.openqa.selenium.Keys;
 import org.openqa.selenium.NoSuchElementException;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.support.ui.Select;
@@ -103,7 +105,14 @@ public class Control extends CapybaraPortingLayerImpl {
      * @see #clickAndWaitToBecomeStale(Duration)
      */
     public void click() {
-        resolve().click();
+        WebElement we = resolve();
+        // button may be obscured by say the "Save Apply" screen so we wait as Selenium will do a scroll but the CSS 
+        // can take a while to update the layout \o/
+        waitFor(we).
+               withTimeout(time.seconds(1), TimeUnit.MILLISECONDS).
+               pollingEvery(time.milliseconds(100), TimeUnit.MILLISECONDS).
+               ignoring(ElementClickInterceptedException.class).
+               until(() -> {we.click(); return true;});
     }
 
 
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Control.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Control.java
@@ -196,6 +196,8 @@ public class Control extends CapybaraPortingLayerImpl {
     public void selectDropdownMenu(Class type) {
         click();
         WebElement we = findCaption(type,findDropDownMenuItem);
+        // the element may not yet be visible so wait for it to become shown after the click above
+        waitFor(we).pollingEvery(100L, TimeUnit.MILLISECONDS).withTimeout(time.milliseconds(1000), TimeUnit.MILLISECONDS).until(() -> we.isDisplayed());
         we.click();
         // wait until the menu is hidden
         waitFor(we).pollingEvery(100L, TimeUnit.MILLISECONDS).withTimeout(time.milliseconds(1000), TimeUnit.MILLISECONDS).until(() -> !we.isDisplayed());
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/Control.java
```diff
@@ -103,7 +105,14 @@ public void check(boolean state) {
      * @see #clickAndWaitToBecomeStale(Duration)
      */
     public void click() {
-        resolve().click();
+        WebElement we = resolve();
+        // button may be obscured by say the "Save Apply" screen so we wait as Selenium will do a scroll but the CSS 
+        // can take a while to update the layout \o/
+        waitFor(we).
```
jtnord, line 111: @aHenryJard this is the flake in the MCP tests.

I would prefer a method to actually check if something is clickable rather than trying repeatedly but I could not find anything.

src/main/java/org/jenkinsci/test/acceptance/po/Control.java
```diff
@@ -186,15 +195,18 @@ public void set(Object text) {
      */
     public void selectDropdownMenu(Class type) {
         click();
-        findCaption(type,findDropDownMenuItem).click();
-        elasticSleep(1000);
```
jtnord, line 190: sleeps must die :)

now handled in [`findDropDownMenuItem`](https://github.com/jenkinsci/acceptance-test-harness/pull/619/files#diff-924ccb08900e51e37d8f6768a863e87795544b16c56986caa3047a709fb7d0c5R233-R238)

src/main/java/org/jenkinsci/test/acceptance/po/Control.java
```diff
@@ -218,8 +230,13 @@ protected WebElement find(String caption) {
                             "    }" +
                             ");"
             );
+            // we can not use `Select` as these are YUI menus and we need to wait for it to be visible
+            WebElement context = waitFor(menuButton).
+                    pollingEvery(100, TimeUnit.MILLISECONDS).
+                    withTimeout(time.milliseconds(1000), TimeUnit.MILLISECONDS).
+                    ignoring(NoSuchElementException.class).
+                    until((me) -> findElement(me, by.xpath("ancestor::*[contains(@class,'yui-menu-button')]/..")));
```
jtnord, line 238: looking at this this seems incorrect.
the element already exists and this method is not checking visibility.
we just need to wait for the context to be visible. after finding it.

src/main/java/org/jenkinsci/test/acceptance/po/Control.java
```diff
@@ -103,7 +105,14 @@ public void check(boolean state) {
      * @see #clickAndWaitToBecomeStale(Duration)
      */
     public void click() {
-        resolve().click();
+        WebElement we = resolve();
+        // button may be obscured by say the "Save Apply" screen so we wait as Selenium will do a scroll but the CSS 
+        // can take a while to update the layout \o/
+        waitFor(we).
+               withTimeout(time.seconds(1), TimeUnit.MILLISECONDS).
```
aHenryJard, line 112: This seems a lot of code to say 1 second.
Maybe `Duration.ofSeconds(1)` if it's the same ? I am not sure on the duration just by reading.

src/main/java/org/jenkinsci/test/acceptance/po/Control.java
```diff
@@ -186,15 +195,22 @@ public void set(Object text) {
      */
     public void selectDropdownMenu(Class type) {
         click();
-        findCaption(type,findDropDownMenuItem).click();
-        elasticSleep(1000);
+        WebElement we = findCaption(type,findDropDownMenuItem);
+        // the element may not yet be visible so wait for it to become shown after the click above
+        waitFor(we).pollingEvery(100L, TimeUnit.MILLISECONDS).withTimeout(time.milliseconds(1000), TimeUnit.MILLISECONDS).until(() -> we.isDisplayed());
```
aHenryJard, line 200: I see now that `time` is an ElasticTime .... can ElasticTime return a Duration maybe ?

aHenryJard (commented):
Thank you !!
I just find all timout/waiting times hard to read, if it's possible to use a shorter way like in Duration something similar to time.asSecond(1) time.asMillisecond(1000) it would be great.

src/main/java/org/jenkinsci/test/acceptance/po/Control.java
jtnord: `Duration` causes a different `WaitFor`.  obvious isn't it :)

src/main/java/org/jenkinsci/test/acceptance/po/Control.java
jtnord: we can not use Duration in this `waitFor`  the duration is defined in the `FluentWait` super class and then we can not use the `until` methods defined in our `Wait`

src/main/java/org/jenkinsci/test/acceptance/po/Control.java
jtnord: this is wrong.  we already use the elastic time in the `waitFor` so do not need to apply it twice.

aHenryJard (approved):

amuniz (approved):

amuniz:
99 failures, 314 in the latest master build... so this is improving the situation! Go ahead!

src/main/java/org/jenkinsci/test/acceptance/po/Control.java
```diff
@@ -103,7 +105,14 @@ public void check(boolean state) {
      * @see #clickAndWaitToBecomeStale(Duration)
      */
     public void click() {
-        resolve().click();
+        WebElement we = resolve();
+        // button may be obscured by say the "Save Apply" screen so we wait as Selenium will do a scroll but the CSS
```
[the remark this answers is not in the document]
jtnord: ahh thanks @amuniz I was wondering where the scrolling was happening as it didn;t seem to be Selenium :)

PR #620, opened 2020-11-13
fix for pipeline-stage-view-2.18

olamy:
previous test was based on a bug fixed in new version https://github.com/jenkinsci/pipeline-stage-view-plugin/pull/108


dwnusbaum (approved):
Looks like there are a bunch of test failures in the CI build, but maybe that is expected? From a quick glance they don't seem related.

PR #621, opened 2020-11-17
Adding a test PluginManager.checkForUpdates() to reproduce 2.263 issue.

aHenryJard:
My issue is that the page reload when click on the "Check now" button on the Plugin management page is not detected by the driver (firefox 81.0.2 and Jenkins version 2.266).

- Add a test to run PluginManager.checkForUpdates()
- Use isStale instead of previous loop on webElement to wait for page to reload
- Changed comment message, as when the check now is in error there is just an additional div hapenned to the DOM, the page is the same that when on success.


jtnord:
@aHenryJard reported directly that the issue was fixed by using the `selenium/standalone-firefox-debug:3.141.59` container to run Firefox - so very likely a mismatch between the driver and the local Firefox version (or driver and selenium version)

so closing this

PR #622, opened 2020-11-27
Fix selector broken by case change in plugin

fqueiruga:
postbuildscript-2.11.0 changed  the string "Execute Scripts" to "Execute scripts" ([see commit](https://github.com/jenkinsci/postbuildscript-plugin/commit/8d0170cd875f7a6f755d89f3e0cbeb0a27ad4a74#diff-ebd76acd1f0295f4a5d4d9e7d79957b24d1df54421520cf75ad44f873e3d93c3)). This broke the following tests

- `PostBuildScriptPluginTest#run_post_build_step`
- `PostBuildScriptPluginTest#skip_for_jobs_that_succeeded`

This PR fixes the selector

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/post_build_script/PostBuildScript.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/post_build_script/PostBuildScript.java
@@ -26,7 +26,7 @@ package org.jenkinsci.test.acceptance.plugins.post_build_script;
 import org.jenkinsci.test.acceptance.po.*;
 import org.openqa.selenium.By;
 
-@Describable({"Execute Scripts", "Execute a set of scripts", "[PostBuildScript] - Execute a set of scripts"})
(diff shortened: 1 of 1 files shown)


uhafner (commented):
Wouldn't it be better to add both values (old and new) then the plugin tests will continue to work with the previous version as well?

fqueiruga:
You're right, done.

timja (approved):

timja:
> Wouldn't it be better to add both values (old and new) then the plugin tests will continue to work with the previous version as well?

not really sure why you would want to verify old versions of this plugin? unless it's core baseline is really new you would always be validating the latest version I would expect

uhafner:
I don't know if that makes sense for that particular plugin. But in general, if the new release of this plugin requires a newer core version, then a depending plugin will still pick up the old version of the plugin in its UI tests.

uhafner (approved):

PR #623, opened 2020-11-30
Reduce flakiness during signup: Make signup to wait until the page is loaded

EstherAF:
The signup is a source of flakiness because the control of the `signup` function is returned before making sure that the submission finished loading. This is addressing it, with the implicit timeout of the `clickAndWaitToBecomeStale`, which is 30s.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/JenkinsDatabaseSecurityRealm.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/JenkinsDatabaseSecurityRealm.java
@@ -23,6 +23,8 @@
  */
 package org.jenkinsci.test.acceptance.po;
 
+import java.time.Duration;
+
 import org.openqa.selenium.NoSuchElementException;
 import hudson.util.VersionNumber;
 
@@ -81,7 +83,7 @@ public class JenkinsDatabaseSecurityRealm extends SecurityRealm {
 
         public User signup(String name) {
             control(by.input("username")).set(name);
-            control(by.name("Submit")).click();
+            control(by.name("Submit")).clickAndWaitToBecomeStale(Duration.ofSeconds(120));
 
             return new User(getJenkins(), name);
         }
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/JenkinsDatabaseSecurityRealm.java
```diff
@@ -81,7 +83,7 @@ public Signup email(String mail) {
 
         public User signup(String name) {
             control(by.input("username")).set(name);
-            control(by.name("Submit")).click();
+            control(by.name("Submit")).clickAndWaitToBecomeStale(Duration.ofSeconds(120));
```
amuniz, line 86: 2 minutes? Seems too much, I think 1 or 2 seconds should be more than enough.

EstherAF: lol sorry, I didn't pushed the last commit :facepalm: 

now it should be using the default timeout of `clickAndWaitToBecomeStale`, which is 30s

src/main/java/org/jenkinsci/test/acceptance/po/JenkinsDatabaseSecurityRealm.java
```diff
@@ -23,6 +23,8 @@
  */
 package org.jenkinsci.test.acceptance.po;
 
+import java.time.Duration;
```
escoem, line 26: Is it used?

escoem (approved):

src/main/java/org/jenkinsci/test/acceptance/po/JenkinsDatabaseSecurityRealm.java
```diff
@@ -23,6 +23,8 @@
...
+
```
jtnord, line 27:
```suggestion
```

jtnord (approved):

PR #624, opened 2020-12-01
[JENKINS-62867] Fixes job filtering selection in views

alecharp:
[JENKINS-62867](https://issues.jenkins.io/browse/JENKINS-62867)

Since  2.239 and jenkinsci/jenkins#4466, the job filter has its own view.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/dashboard_view/controls/JobFiltersArea.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/dashboard_view/controls/JobFiltersArea.java
@@ -12,9 +12,13 @@ import org.jenkinsci.test.acceptance.po.PageObject;
  */
 public class JobFiltersArea extends PageAreaImpl {
     /**
+     * Button to add a new status filter
+     */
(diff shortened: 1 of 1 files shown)


fqueiruga (approved):

MRamonLeon (commented):
I think we have to keep the old behavior to keep backward compatibility with old versions of the plugin. Some `if` with or similar:

```
jenkins.getPlugin("dashboard_view").getVersion().isOlderThan(new VersionNumber("x.y.z"))
```

timja:
> I think we have to keep the old behavior to keep backward compatibility with old versions of the plugin. Some `if` with or similar:
> 
> ```
> jenkins.getPlugin("dashboard_view").getVersion().isOlderThan(new VersionNumber("x.y.z"))
> ```

this is core that changed not a plugin, but yes that could be done, not sure why you would want to though, this is now in 2 LTS versions, (2.263.1 and 2.249.x).

timja (approved):

alecharp:
Yes, this is a core change, not a plugin one. Getting the core version the test is running on from a `PageArea` class is not simple it seems.

timja:
Test is passing

PR #625, opened 2020-12-01
Try to fix stale test

Vlatombe:
After upgrading firefox to 83.0, I started seeing this stacktrace...

```
 Web element reference not seen before: 1ec17ca6-ad5d-469a-ab36-c961982fc12a
 For documentation on this error, please visit: https://www.seleniumhq.org/exceptions/no_such_element.html
 Build info: version: '3.141.59', revision: 'e82be7d358', time: '2018-11-14T08:17:03'
 System info: host: 'host', ip: '1.2.3.4', os.name: 'Linux', os.arch: 'amd64', os.version: '4.9.0-14-amd64', java.version: '1.8.0_275'
 Driver info: org.openqa.selenium.firefox.FirefoxDriver
 Capabilities {acceptInsecureCerts: true, browserName: firefox, browserVersion: 83.0, javascriptEnabled: true, moz:accessibilityChecks: false, moz:buildID: 20201112153044, moz:geckodriverVersion: 0.28.0, moz:headless: false, moz:processID: 2132, moz:profile: /tmp/rust_mozprofileuBQERw, moz:shutdownTimeout: 60000, moz:useNonSpecCompliantPointerOrigin: false, moz:webdriverClick: true, pageLoadStrategy: normal, platform: LINUX, platformName: LINUX, platformVersion: 4.9.0-14-amd64, proxy: Proxy(manual, http=host, rotatable: false, setWindowRect: true, strictFileInteractability: false, timeouts: {implicit: 0, pageLoad: 300000, script: 30000}, unhandledPromptBehavior: dismiss and notify}
 Session ID: f9c3668c-306b-4710-ae8c-749f30cf0768
 	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
 	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
 	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
 	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
 	at org.openqa.selenium.remote.http.W3CHttpResponseCodec.createException(W3CHttpResponseCodec.java:187)
 	at org.openqa.selenium.remote.http.W3CHttpResponseCodec.decode(W3CHttpResponseCodec.java:122)
 	at org.openqa.selenium.remote.http.W3CHttpResponseCodec.decode(W3CHttpResponseCodec.java:49)
 	at org.openqa.selenium.remote.HttpCommandExecutor.execute(HttpCommandExecutor.java:158)
 	at org.openqa.selenium.remote.service.DriverCommandExecutor.execute(DriverCommandExecutor.java:83)
 	at org.openqa.selenium.remote.RemoteWebDriver.execute(RemoteWebDriver.java:552)
 	at org.openqa.selenium.remote.RemoteWebElement.execute(RemoteWebElement.java:285)
 	at org.openqa.selenium.remote.RemoteWebElement.isEnabled(RemoteWebElement.java:156)
 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
 	at java.lang.reflect.Method.invoke(Method.java:498)
 	at org.openqa.selenium.support.events.EventFiringWebDriver$EventFiringWebElement.lambda$new$0(EventFiringWebDriver.java:404)
 	at com.sun.proxy.$Proxy37.isEnabled(Unknown Source)
 	at org.openqa.selenium.support.events.EventFiringWebDriver$EventFiringWebElement.isEnabled(EventFiringWebDriver.java:457)
 	at org.jenkinsci.test.acceptance.po.CapybaraPortingLayerImpl.isStale(CapybaraPortingLayerImpl.java:343)
 	at org.openqa.selenium.support.ui.FluentWait.until(FluentWait.java:249)
 	at org.jenkinsci.test.acceptance.po.Control.clickAndWaitToBecomeStale(Control.java:127)
 	at org.jenkinsci.test.acceptance.po.Control.clickAndWaitToBecomeStale(Control.java:115)
 	at org.jenkinsci.test.acceptance.po.Login.doLogin(Login.java:35)
```

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
@@ -344,6 +344,8 @@ public class CapybaraPortingLayerImpl implements CapybaraPortingLayer {
             return false;
         } catch (StaleElementReferenceException expected) {
             return true;
+        } catch (NoSuchElementException expected) {
+            return true;
         }
     }
 
--- a/src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
@@ -345,6 +345,7 @@ public class CapybaraPortingLayerImpl implements CapybaraPortingLayer {
         } catch (StaleElementReferenceException expected) {
             return true;
         } catch (NoSuchElementException expected) {
+            // work around https://github.com/SeleniumHQ/selenium/issues/8929
             return true;
         }
     }


src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
```diff
@@ -344,6 +344,8 @@ protected static boolean isStale(WebElement element) {
             return false;
         } catch (StaleElementReferenceException expected) {
             return true;
+        } catch (NoSuchElementException expected) {
```
jtnord, line 347: as I understand this, this exception should be impossible.  
in order to be here (given the stack) the `WebElement` has been interacted with (via `click`) so we know it exists at that point in time.

all we are doing is a freshess check which according to the Javadoc (which may be lying) is

> All method calls will do a freshness check to ensure that the element reference is still valid. This essentially determines whether or not the element is still attached to the DOM. If this test fails, then an StaleElementReferenceException is thrown, and all future calls to this instance will fail.

thus it should not even be trying to re-resolve the `WebElement` , but just saying yup it is no longer in the DOM, and throw a `StaleElementReferenceException`
`NoSuchElementExcetion` should only be throw when `findElement` is called (either in `WebDriver` or `WebElement`)

Vlatombe: I checked my docker images and it was working previously with Firefox 80, and got broken in Firefox 83.

Looking at the Firefox 82 [changelog](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/82), I can find this of interest

> Element.setPointerCapture() throws the NotFoundError exception if the pointer id is invalid, as per the specification (bug 1662124). Previously it wrongly threw an InvalidPointerId exception.

Which AFAIU could translate into `StaleElementReferenceException` vs. `NoSuchElementException` on selenium side.

Vlatombe:
Confirming that this change fixed the problem for me on Firefox 83.

timja (approved):

src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
jtnord: so this sounds like a bug in Selenium?  ie it should now translate this when not using `findBy`?

I'm hesitant to make this change as there will likely be a lot of other places that also do something similar.

jtnord: https://github.com/SeleniumHQ/selenium/issues/8929

jtnord:
looks like the cause will impact other things also that are just expecting a `StaleElementException`. Filed https://github.com/SeleniumHQ/selenium/issues/8929.
@Vlatombe can you run with the older firefox for now until we here back about the above bug report, or do you need to upgrade?

src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
jtnord, line 347:
```suggestion
        } catch (NoSuchElementException expected) {
        // work around https://github.com/SeleniumHQ/selenium/issues/8929
```

Vlatombe:
I'll give a try to latest firefox esr as I was getting hang with Firefox 80/81.

jtnord (approved):

Vlatombe:
Looks like it works for me with `78.5.0esr` so I don't need this change urgently.

timja:
> Looks like it works for me with `78.5.0esr` so I don't need this change urgently.

I tried to use firefox recently and gave up and used chrome.

jtnord:
@Vlatombe I think this may be fixed in firefox 85
https://github.com/mozilla/geckodriver/issues/1826

would you be able to test and then close this as uneeded if it now works for you?

jtnord:
> I tried to use firefox recently

FWIW I exclusivly use firefox to run the tests as does the ATH.  the firefox in a container is the easiest approach :)

jtnord:
also adding more direct link to https://bugzilla.mozilla.org/show_bug.cgi?id=1684827

closing this as should not be needed (esr is unaffected and version 85 is supposedily fixed and is now public)

PR #629, opened 2020-12-17
Support form-element-path installation through rest api

Vlatombe:
I'm running an ATH against an existing Jenkins instance running in Kubernetes. The usual workflow is to login, go through the setup wizard (once), then perform a functional test.

This PR brings a few changes to support this way of running an ATH:
* Support `form-element-path` installation through rest api
* `ExistingJenkinsController`: Initial support for authenticated Jenkins when using `TYPE=existing`
* `ExistingJenkinsController`: checking for form-element-path can be skipped. (this is called later)
* Support for access token generation

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/controller/JenkinsController.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/controller/JenkinsController.java
@@ -166,7 +166,7 @@ public abstract class JenkinsController implements IJenkinsController, AutoClean
     public abstract URL getUrl();
 
     @CheckForNull
-    public Credentials getCredentials() {
+    public Credentials getInitialCredentials() {
         return null;
     }
 
--- a/src/main/java/org/jenkinsci/test/acceptance/controller/JenkinsController.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/controller/JenkinsController.java
@@ -1,18 +1,16 @@
 package org.jenkinsci.test.acceptance.controller;
 
+import javax.annotation.CheckForNull;
 import javax.inject.Inject;
 import javax.inject.Named;
-import java.io.File;
-import java.io.FileOutputStream;
 import java.io.IOException;
-import java.io.OutputStream;
 import java.net.URL;
 import java.util.logging.Level;
 import java.util.logging.LogManager;
 import java.util.logging.Logger;
 
 import org.apache.commons.io.IOUtils;
-import org.codehaus.plexus.util.FileUtils;
+import org.apache.http.auth.Credentials;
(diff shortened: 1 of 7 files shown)


src/main/java/org/jenkinsci/test/acceptance/utils/AccessTokenGenerator.java
```diff
@@ -0,0 +1,177 @@
...
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Takes static credentials and generates access tokens. Handles caching so that
```
jglick, line 28: …?

src/main/java/org/jenkinsci/test/acceptance/utils/FormElementPath.java
```diff
@@ -0,0 +1,95 @@
...
+ * Ensures form-element-path plugin is running, and if not, installs it using the rest api to upload a plugin
+ */
+public class FormElementPath {
+    private static final Logger LOGGER = Logger.getLogger(FormElementPath.class.getName());
+
+    private Credentials credentials;
+
+    @Inject
+    @Named("form-element-path.hpi")
+    private File formElementPathPlugin;
```
jglick, line 39: Could actually be put into the ATH classpath via `dependency:copy`.

src/main/java/org/jenkinsci/test/acceptance/utils/AccessTokenGenerator.java
Vlatombe: Probably got interrupted by something ;)

src/main/java/org/jenkinsci/test/acceptance/utils/FormElementPath.java
Vlatombe: This was already in place, I figured this could just be reused.

src/main/java/org/jenkinsci/test/acceptance/controller/JenkinsController.java
```diff
@@ -164,6 +165,11 @@ public boolean isRunning(){
     @Override
     public abstract URL getUrl();
 
+    @CheckForNull
+    public Credentials getCredentials() {
```
jtnord, line 169: Should this be named 'initialCredentials'. They will become useless once a securotyRealm is configured won't they?

Vlatombe: Renamed.

Vlatombe:
I'll merge this tomorrow unless there are objections.

PR #631, opened 2021-01-19
Make check of logged-in user independent of UI

fcojfernandez:
Making the check of the logged user using Jenkins API instead the UI, so the check is more robust in case something is not well rendered.

@bmunozm @EstherAF 

@jtnord I think you're the new maintainer?

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/Matchers.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/Matchers.java
@@ -201,7 +201,8 @@ public class Matchers {
             @Override
             public boolean matchesSafely(final Jenkins jenkins) {
                 final User currentUser = jenkins.getCurrentUser();
-                return currentUser != null && currentUser.id().equals(user);
+                // if the user is not logged, currentUser can be not null with a null id
+                return currentUser != null && currentUser.id() != null && currentUser.id().equals(user);
             }
 
             @Override
@@ -216,7 +217,8 @@ public class Matchers {
             @Override
             public boolean matchesSafely(final Login login) {
                 final User currentUser = login.getJenkins().getCurrentUser();
-                return currentUser != null && currentUser.id().equals(user);
+                // if the user is not logged, currentUser can be not null with a null id
+                return currentUser != null && currentUser.id() != null && currentUser.id().equals(user);
             }
 
             @Override
--- a/src/main/java/org/jenkinsci/test/acceptance/Matchers.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/Matchers.java
@@ -200,12 +200,8 @@ public class Matchers {
         return new Matcher<Jenkins>("has logged in user %s", user) {
             @Override
             public boolean matchesSafely(final Jenkins jenkins) {
-                try {
-                    jenkins.find(by.href("/user/" + user));
-                    return true;
-                } catch (NoSuchElementException e) {
-                    return false;
-                }
+                final User currentUser = jenkins.getCurrentUser();
+                return currentUser != null && currentUser.id().equals(user);
             }
 
             @Override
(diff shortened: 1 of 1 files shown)


jtnord (approved):

amuniz (approved):

jglick (approved):
IIUC uses https://github.com/jenkinsci/acceptance-test-harness/blob/ca473c8f88711493adda9973066e0ec156533856/src/main/java/org/jenkinsci/test/acceptance/po/User.java#L45-L56

fcojfernandez:
When the user is not valid, the `User` object is not null but its `id` is. `User` seems to be null only if there has been no attempt to login. That made `LdapPluginTest` failed in a couple of tests.

Already Fixed

src/main/java/org/jenkinsci/test/acceptance/Matchers.java
```diff
@@ -201,7 +201,8 @@ public boolean matchesSafely(PageObject item) {
             @Override
             public boolean matchesSafely(final Jenkins jenkins) {
                 final User currentUser = jenkins.getCurrentUser();
-                return currentUser != null && currentUser.id().equals(user);
+                // if the user is not logged, currentUser can be not null with a null id
+                return currentUser != null && currentUser.id() != null && currentUser.id().equals(user);
```
jglick, line 205: or just
```suggestion
                return currentUser != null && user.equals(currentUser.id());
```

I suppose

jglick (approved):

src/main/java/org/jenkinsci/test/acceptance/Matchers.java
fcojfernandez: that was my first intention, but after a search I couldn't be 100% sure that `user` will be always non-null. I can commit the suggestion if you prefer

amuniz:
There are currently 264 test failures in master, this PR has 330. Is the diff because of this change?

fcojfernandez:
After analysing the failures (and I'm not experienced on the plugins whose tests are failing and I might be wrong):
- None of those tests are checking the logged user, that is what is changed in this PR
- Only `AuditTrailPluginTest` and `MissionControlTest` are failing in local. The rest of them are passing using the latest LTS in a local execution. Why are the tests failing then? I don't know and I don't have knowledge about those plugins to know if there's something in the test per se (my guess)

**The details:**

1. PlotPluginPropertiesTest: Not using either Login.doSuccessfullLogin or Matchers.hasLoggedInUser / Matchers.loggedInAs. Local execution with latest LTS is OK
1. JobDslPluginTest: Not using either Login.doSuccessfullLogin or Matchers.hasLoggedInUser / Matchers.loggedInAs. Local execution with latest LTS is OK
1. PrioritySorterPluginTest: Not using either Login.doSuccessfullLogin or Matchers.hasLoggedInUser / Matchers.loggedInAs. Local execution with latest LTS is OK
1. MissionControlTest: Not using either Login.doSuccessfullLogin or Matchers.hasLoggedInUser / Matchers.loggedInAs. Local execution with latest LTS is failing as well, but the tests are executed as anonymous user. NOT RELATED
1. FtpPublishPluginTest: Not using either Login.doSuccessfullLogin or Matchers.hasLoggedInUser / Matchers.loggedInAs. Local execution with latest LTS is OK
1. PlotPluginXmlTest: Not using either Login.doSuccessfullLogin or Matchers.hasLoggedInUser / Matchers.loggedInAs. Local execution with latest LTS is OK
1. AuditTrailPluginTest: Not using either Login.doSuccessfullLogin or Matchers.hasLoggedInUser / Matchers.loggedInAs. Local execution with latest LTS is is failing as well, but the tests are executed as anonymous user. NOT RELATED
1. core.BuildHistoryTest: Not using either Login.doSuccessfullLogin or Matchers.hasLoggedInUser / Matchers.loggedInAs. Local execution with latest LTS is OK

timja:
Plot plugin is known to be broken on weeklies. It should probably be removed from ATH, I don't think it meets the threshold for what should be here.

amuniz:
Great. Thanks @fcojfernandez! I'm merging now.

PR #638, opened 2021-02-02
geckodriver/marionette debugging

jtnord:
needed to get the geckodriver logs from a build in ci.jenknis.io (warning-ng / git-forensics) that are not archvied or provided in the job config.
this was just a temp solution not intended to be merged that I could use to dump those logs to the console.

Leaving open so it is findable if we need it again - but should not be merged


jtnord:
just need something crude to get the geckdriver logs from ci.jenkins.io as they are not part of the build (nor are many of the other interesting bits).

jtnord:
upstream issue has been fixed, can close this and find it later if ever needed again

PR #642, opened 2021-02-16
Attach support bundle after running the test

Vlatombe:
Collect a support bundle a after a failing test.

I found this very useful to investigate flaky tests as it collects way
more information than the standard output.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/junit/AbstractJUnitTest.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/junit/AbstractJUnitTest.java
@@ -2,11 +2,14 @@ package org.jenkinsci.test.acceptance.junit;
 
 import org.jenkinsci.test.acceptance.po.CapybaraPortingLayerImpl;
 import org.jenkinsci.test.acceptance.po.Jenkins;
+import org.jenkinsci.test.acceptance.utils.SupportBundleRequest;
+import org.junit.After;
 import org.junit.Rule;
 import org.openqa.selenium.WebDriver;
 
 import javax.inject.Inject;
 
+import java.io.File;
 import java.io.IOException;
 import java.net.ServerSocket;
 
@@ -28,6 +31,9 @@ public class AbstractJUnitTest extends CapybaraPortingLayerImpl {
      */
     @Inject
     public Jenkins jenkins;
+
+    @Inject
+    private FailureDiagnostics diagnostics;
 
     /**
      * This field receives a valid web driver object you can use to talk to Jenkins.
@@ -55,5 +61,11 @@ public class AbstractJUnitTest extends CapybaraPortingLayerImpl {
 
     protected void interrupt() {
         throw new RuntimeException("INTERACTIVE debugging");
+    }
+
+    @After
+    public void captureSupportBundle() {
+        File file = diagnostics.touch("support-bundle.zip");
+        jenkins.generateSupportBundle(SupportBundleRequest.builder().includeDefaultComponents().setOutputFile(file).build());
     }
 }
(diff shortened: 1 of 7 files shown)


src/main/java/org/jenkinsci/test/acceptance/junit/AbstractJUnitTest.java
```diff
@@ -56,4 +69,9 @@ public int findAvailablePort() {
     protected void interrupt() {
         throw new RuntimeException("INTERACTIVE debugging");
     }
+
+    @After
+    public void injectSpec() {
+        supportBundle.addSpec(jenkins, SupportBundleRequest.builder().includeDefaultComponents().setOutputFile(diagnostics.touch("support-bundle.zip")).build());
```
aHenryJard, line 75: What happens if several tests are failing, could you have a filename collision ?

Vlatombe: No, the attachment is per test case

jtnord: the attachment is - but the junit attachment archiver only kicks in at the end running all the tests doesn't it?

src/main/java/org/jenkinsci/test/acceptance/recorder/SupportBundle.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public SupportBundle() {}
+
+    public void addSpec(Jenkins jenkins, SupportBundleRequest request) {
+        specs.add(new SupportBundleSpec(jenkins, request));
+    }
+
+    @Override
+    protected void failed(Throwable e, Description description) {
+        for (SupportBundleSpec spec : specs) {
```
jtnord, line 35: should this not be opt in like `RECORDER` and the others?
Just because I have a plugin installed does not mean I want these bundles does it (especially when running locally) - as bundles are not always tiny.

src/main/java/org/jenkinsci/test/acceptance/po/Jenkins.java
```diff
@@ -217,4 +230,81 @@ public JobsMixIn getJobs() {
...
+
+        public void setCrumbRequestField(String crumbRequestField) {
+            this.crumbRequestField = crumbRequestField;
+        }
+    }
+
+    private HttpContext buildHttpClientContext() {
+        HttpClientContext context = HttpClientContext.create();
+        BasicCookieStore cookieStore = new BasicCookieStore();
+        for (org.openqa.selenium.Cookie cookie : driver.manage().getCookies()) {
```
amuniz, line 300: This adding all cookies used by `driver` up to this point. AFAICT It can contain cookies that are not from the target master here (ie. another master was accessed during the same test, which won't happen in regular Jenkins ATH but will happen in more complex ATH setups using this project as a base).

amuniz: I think there was something similar when checking for logged in users (before the recent switch to API instead of UI).

src/main/java/org/jenkinsci/test/acceptance/recorder/SupportBundle.java
Vlatombe: How about I enable it by default in CI, and disable it locally ?

Vlatombe: In any case, we're mostly talking about 100kb support bundles, since they are fresh instances.

Vlatombe: cab03793d88d7ab843dd588811407419dd15f63d

src/main/java/org/jenkinsci/test/acceptance/po/Jenkins.java
Vlatombe: Do you want me to change it? I see that as premature optimization.

jtnord:
all failures are currently expected :(

jtnord (approved):

jtnord:
@Vlatombe could you add something to the docs page about configuring this please then I will merge and cut a release.

Vlatombe:
@jtnord done

jtnord (approved):

PR #643, opened 2021-02-18
reduce flakiness of doSuccesfulLogin

Vlatombe:
I'm seeing regularly flakiness in my ATH such as

```
Expected: has logged in user tmpuser
     but: tmpuser is not logged in.
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.junit.Assert.assertThat(Assert.java:964)
	at org.junit.Assert.assertThat(Assert.java:930)
	at org.jenkinsci.test.acceptance.po.Login.doSuccessfulLogin(Login.java:67)
	at org.jenkinsci.test.acceptance.po.Login.doSuccessfulLogin(Login.java:72)
```

I think in some cases the current check is not enough and the navigation happens before the login is really completed.

While working on this area, I notices `doLoginDespiteNoPaths` was useless since `doLogin` has been changed to not use path since https://github.com/jenkinsci/acceptance-test-harness/commit/ebdd746c9a9b57877e4035ee5674f0344b32818e. So I deprecated it and removed calls I could find in this ATH.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Login.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Login.java
@@ -1,14 +1,17 @@
 package org.jenkinsci.test.acceptance.po;
 
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.time.Duration;
-import java.util.concurrent.TimeUnit;
 
+import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.ui.ExpectedConditions;
+import org.openqa.selenium.support.ui.WebDriverWait;
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.jenkinsci.test.acceptance.Matchers.hasInvalidLoginInformation;
 import static org.jenkinsci.test.acceptance.Matchers.loggedInAs;
-import static org.junit.Assert.assertEquals;
 
 /**
  * Page object for login page.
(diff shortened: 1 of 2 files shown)


src/test/java/plugins/ActiveDirectoryTest.java
```diff
@@ -88,11 +88,11 @@ public void wannabe_cannot_login_to_Jenkins_after_AD_security_configured() {
         String userWannabe = ENV.getUser() + "-wannabe";
         GlobalSecurityConfig security = saveSecurityConfig(userWannabe);
         jenkins.logout();
-        jenkins.login().doLoginDespiteNoPaths(userWannabe,
+        jenkins.login().doLogin(userWannabe,
```
jtnord, line 91: seems like this should be `doSuccessfulLogin` but existing behaviour so :shrug: ?

Vlatombe: Yes probably

src/main/java/org/jenkinsci/test/acceptance/po/Login.java
```diff
@@ -64,7 +58,11 @@ public Login doLogin(User user) {
 
     public Login doSuccessfulLogin(String user, String password) {
         this.doLogin(user, password);
-        Assert.assertThat(this, loggedInAs(user));
+        waitFor().withTimeout(30, TimeUnit.SECONDS).until(() -> {
+            assertThat(this, not(hasInvalidLoginInformation())); // login hasn't failed
+            return ExpectedConditions.visibilityOfElementLocated(by.id("jenkins")); // redirect has happened
```
jtnord, line 63: I am wondering if this can happen on other places where we also use `clickAndWaitToBecomeStale` or until(CapybaraPortingLayerImpl::isStale);

This could likely be the cause (and the fix) of  the race condition on creating a new `Folder` that has been observed a a few times. https://issues.jenkins.io/browse/JENKINS-64358

jtnord (approved):

src/main/java/org/jenkinsci/test/acceptance/po/Login.java
Vlatombe: Yes it looks like the same class of issue.

amuniz (approved):

PR #650, opened 2021-03-05
try to fix git plugin tests

jtnord:
The git plugin gets an old version of trilead as it uses an older core
with a detached version so it gets installed from the war and not the
update center.
attempt to work around this by requesting trilead be installed in the
"with-plugins" so we get something from the update center

- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!


jtnord:
see #651

PR #652, opened 2021-03-10
Removing SAMLPluginTest

jglick:
As of https://github.com/jenkinsci/saml-plugin/pull/93 in 2.0.1, the SAML plugin test here are redundant—basically the same coverage, just slower and harder to run, and apparently ignored since https://github.com/jenkinsci/saml-plugin/pull/90 introduced a regression which did not block the plugin release and was not communicated to the plugin developer after the fact either.


jglick:
@kuisathaverat any opinion?

kuisathaverat (approved):
Go a head, the same test is running on the Plugin

jtnord:
The test in the plugin is for a different audience. The ATH is used to ensure no regressions when releasing a security or LTS version.  Plugins tests at not re-run at this point.

Unclear why the test failure was not communicated, possibly due to the current sea of red, possibly also because no one is looking at the master job output anymore rather just running the ATH locally?

jglick:
> no regressions when releasing a security or LTS version. Plugins tests at not re-run at this point.

Currently true, at least in OSS; CloudBees would run plugin tests against new LTS versions via PCT, at least in the case of this plugin. Maybe the LTS checklist can be extended to include PCT? Someone would need to set up a job for it, and decide which plugins get covered.

> possibly due to the current sea of red, possibly also because no one is looking at the master job output anymore

Yeah I suspect no one is paying much attention to the numerous known failures, much less flakes.

timja:
> Currently true, at least in OSS; CloudBees would run plugin tests against new LTS versions via PCT, at least in the case of this plugin. Maybe the LTS checklist can be extended to include PCT? Someone would need to set up a job for it, and decide which plugins get covered.

Bom is done for new lines at least currently

timja:
> Yeah I suspect no one is paying much attention to the numerous known failures, much less flakes. (Other than SmokeTests, run by the jenkinsci/jenkins build.)

No there's too many failures

amuniz:
We have an internal test for CloudBees SSO using SAML which relies on the page objects and docker fixture being removed here. Could we keep them?

No objection on removing the test itself.

jglick:
> Could we keep them?

Why not just copy those classes to your internal tests if you are using them?

amuniz:
Ok. Fine.

amuniz (approved):

jglick:
> an internal test for CloudBees SSO using SAML which relies on the page objects and docker fixture being removed here

Resolved, FTR

jtnord (approved):

PR #628, opened 2020-12-13
Remove deprecated multiple-scms plugin test

timja:
https://plugins.jenkins.io/multiple-scms/

PR #656, opened 2021-04-15
Deleting MultipleScms page object

jglick:
The [`multiple-scms` plugin](https://plugins.jenkins.io/multiple-scms/) is deprecated, and this page object does not seem to be used in any actual test anyway as of #628.


timja (approved):

jtnord:
so we use it and have no e2e tests for it? https://github.com/cloudbees/cloudbees-ci-platform-plugin/blob/df1a45a9e2b2838f53e24a5810e5f6377b80dbb8/scm-urls/src/main/java/com/cloudbees/jenkins/plugins/data/scmurl/impl/MultiSCMScmInfoResolver.java?

jtnord (approved):

PR #657, opened 2021-04-15
Adapt `GlobalSecurityConfig` PO to modified layout of `GlobalSecurityConfiguration`

jglick:
Adapts tests to https://github.com/jenkinsci/jenkins/pull/5417.

https://github.com/jenkinsci/acceptance-test-harness/blob/a9027d40be247fbc51f4ea4af7e4bd0bd7f7e6e8/src/main/java/org/jenkinsci/test/acceptance/po/DumbSlave.java#L25 and a few other uses of `control("/")` should likely be updated to use a similar idiom.


src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
```diff
@@ -453,6 +454,14 @@ public void elasticSleep(long ms) {
         }
     }
 
+    protected <T> T findCaption(Class<?> type, Function<String, T> finder) {
```
jglick, line 457: Just to make it easier to use lambdas.

timja (approved):

timja:
Looks like this causes lots of failures.

Also would normally point the weekly version of this at an incremental / ci.jenkins.io build / a snapshot here: https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/main/resources/ath-container/run.sh#L48

jglick:
> Looks like this causes lots of failures.

Really? I thought there were already a ton of failures from the `credentials` upgrade.

timja:
> > Looks like this causes lots of failures.
> 
> Really? I thought there were already a ton of failures from the `credentials` upgrade.

Ah I thought Vincent's PR was merged, looks like not quite sorted: https://github.com/jenkinsci/acceptance-test-harness/pull/753

timja:
(green now 🎉 )

timja:
> Oh I was going to wait for the core PR to be merged first, but fine. I guess I can update the core PR to use the merge commit.

Generally I merge ATH first so core doesn't use unreleased code, released as https://github.com/jenkinsci/acceptance-test-harness/releases/tag/acceptance-test-harness-1.112

jglick:
Yeah I noticed that, thanks. (Does not really matter in this case since core is just checking out sources anyway. The releases are only really useful for derivative test suites using ATH as the framework via binary dep.)

PR #659, opened 2021-04-29
Reduce flakiness on plugins check

amuniz:
I've observed this failing with `StaleElementReferenceException` from
time to time. I tracked it down to the plugin manager UI dynamic
behaviour (DOM elements show up and down) so the initially select
element becomes stale after a few milliseconds.

There is no way to know if the element will become stale or not, so the
only way I found to make this less flaky is by waiting 1 second and
retrying, then give up if it fails again.

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
@@ -285,14 +285,19 @@ public class PluginManager extends ContainerPageObject {
         WebElement filterBox = find(By.id("filter-box"));
         filterBox.clear();
         filterBox.sendKeys(name);
-        WebElement pluginEntry = waitFor(by.xpath("//input[starts-with(@name,'plugin.%s.')]", name), 10);
-        // sometimes the web element gets stale because of the dynamic JS behaviour of the plugin manager UI
-        if (Control.isStale(pluginEntry)) {
-            // wait 1 second, re-try and give up if it fails again
-            sleep(1000);
-            pluginEntry = waitFor(by.xpath("//input[starts-with(@name,'plugin.%s.')]", name), 10);
-        }
-        check(pluginEntry);
+
+        // the target plugin web element becomes stale due to the dynamic behaviour of the plugin
+        // manager UI which ends up with StaleElementReferenceException.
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
```diff
@@ -285,7 +285,14 @@ private void tickPluginToInstall(PluginSpec spec) {
         WebElement filterBox = find(By.id("filter-box"));
         filterBox.clear();
         filterBox.sendKeys(name);
-        check(waitFor(by.xpath("//input[starts-with(@name,'plugin.%s.')]", name), 10));
+        WebElement pluginEntry = waitFor(by.xpath("//input[starts-with(@name,'plugin.%s.')]", name), 10);
+        // sometimes the web element gets stale because of the dynamic JS behaviour of the plugin manager UI
+        if (Control.isStale(pluginEntry)) {
```
jtnord, line 290: is this actually going to fix the issue described?  
IIUC the staleness is a race condition between the filtering and the resulting display / update and you may not have hit it with this.  (if the element shows up you do the check and *then* the javascript updates the dom)?

src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
```diff
@@ -285,7 +285,14 @@ private void tickPluginToInstall(PluginSpec spec) {
...
+            // wait 1 second, re-try and give up if it fails again
+            sleep(1000);
```
jtnord, line 292: Either the comment is wrong (if it uses elastic time it may wait more or less than 1 second) or this is not using `elasticTime`and it should be :)
How often do we hit this - is waiting 1 second going to make the tests even slower?

Could this be better written using a `waitForCont()` ignoring a `NoSuchElementException` and `StaleElementReferenceException`

amuniz: Good point, let me re-write it.

amuniz:
Should be better now.

timja (approved):

PR #672, opened 2021-06-16
fix: specifies UID and GID from the user running pipeline

dduportal:
Fixes https://issues.jenkins.io/browse/INFRA-3006 until the
docker.inside is removed.

Root cause was https://github.com/jenkins-infra/packer-images/pull/57
which changed the default UID if the agent's `jenkins` user.

Signed-off-by: Damien Duportal <damien.duportal@gmail.com>

### DIFF
--- a/src/main/resources/ath-container/Dockerfile
+++ b/src/main/resources/ath-container/Dockerfile
@@ -33,9 +33,9 @@ ENV PATH="$PATH:/opt/maven/bin"
 ENV SHARED_DOCKER_SERVICE true
 
 # Allow injecting uid and git to match directory ownership
-ARG uid=1000
+ARG uid=1001
 ENV uid $uid
-ARG gid=1000
+ARG gid=1001
 ENV gid $gid
 
 EXPOSE 5942
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -35,7 +35,7 @@ for (int i = 0; i < splits.size(); i++) {
             stage(name) {
                 node('docker && highmem') {
                     checkout scm
-                    def image = docker.build('jenkins/ath', "src/main/resources/ath-container")
+                    def image = docker.build('jenkins/ath', '--build-arg uid="$(id -u)" --build-arg gid="$(id -g)" ./src/main/resources/ath-container/')
                     image.inside('-v /var/run/docker.sock:/var/run/docker.sock --shm-size 2g') {
                         def exclusions = splits.get(index).join("\n")
                         writeFile file: 'excludes.txt', text: exclusions


src/main/resources/ath-container/Dockerfile
```diff
@@ -33,9 +33,9 @@ ENV PATH="$PATH:/opt/maven/bin"
 ENV SHARED_DOCKER_SERVICE true
 
 # Allow injecting uid and git to match directory ownership
-ARG uid=1000
+ARG uid=1001
 ENV uid $uid
-ARG gid=1000
+ARG gid=1001
```
jglick, line 38: Why do these need to be changed if you are overriding them anyway?

jglick (commented):
Whatever works in the short term I guess.

src/main/resources/ath-container/Dockerfile
dduportal: It's related to the condition in the shared library: https://github.com/jenkins-infra/pipeline-library/blob/master/vars/runATH.groovy#L98 .

Based on the logs I see on the failing job (`11:46:50  No building ATH docker container image. Using jenkins/ath:acceptance-test-harness-1.80 as specified` in https://ci.jenkins.io/job/Core/job/jenkins/job/PR-5575/5/consoleFull), I understand that the Docker image is not always build depending on some metadatas provided to the build (either a file or a variable).

That means that the 2nd case should be covered.

src/main/resources/ath-container/Dockerfile
```diff
@@ -33,9 +33,9 @@ ENV PATH="$PATH:/opt/maven/bin"
 ENV SHARED_DOCKER_SERVICE true
 
 # Allow injecting uid and git to match directory ownership
-ARG uid=1000
+ARG uid=1001
```
timja, line 36: shouldn't this already get detected properly?
https://github.com/jenkins-infra/pipeline-library/blob/a6261f941d974857a57929c5074955e741e2ba38/vars/runATH.groovy#L100-L102

src/main/resources/ath-container/Dockerfile
dduportal: It's a short-term thing: use the new UID set up in the packer images. I see the 2 following upcoming changes:

- Easy to do, quick win: Adding a routine in jenkins-infra/packer_images that would run on master branch and open a PR here to update the UID if it is changed. Only infra changes, and it would avoid knowledge to be lost in the future.
- Harder but improved ATH: always building the image, using an efficient caching (Docker buildX + `--cache-from` flag) so the layers could be cached, but the "create user step“ could be run all the time.

src/main/resources/ath-container/Dockerfile
dduportal: @timja our comments crossed: see my response to Jesse: https://github.com/jenkinsci/acceptance-test-harness/pull/672#discussion_r652974486

timja (approved):
FTR the publishing of this image is not automated, do you have permission or want me to publish it later? (not on a computer that can build it atm IIRC (arm didn't work last time I tried) )

dduportal:
@timja I don't have the rights, I'll need the help of someone to do it.
We can do it tomorrow, no problem.

Side note: we could totally open a PR to use the pipeline-libraries for auto building Docker images though :)

src/main/resources/ath-container/Dockerfile
jglick: If you are doing anything with UIDs you are probably doing it wrong.

PR #678, opened 2021-08-04
New WithSystemProperties annotation to add System properties to a single test execution

fcojfernandez:
This PR adds a new `WithSystemProperties` annotation that indicates that the test and Jenkins instance will run with the specified system properties. It is a way to extend `JENKINS_JAVA_OPTS` and `JENKINS_OPTS` at test level (those variables are at full execution and shared by all the tests).

These properties are set as an array of pairs <i>property=vale</i>. Both formats `property=value` and `-Dproperty=value` are acepted


fcojfernandez:
@jtnord @amuniz FYI

src/main/java/org/jenkinsci/test/acceptance/controller/LocalController.java
```diff
@@ -285,15 +285,15 @@ public void diagnose(Throwable cause) {
...
+                FileUtils.deleteDirectory(jenkinsHome);
             }
         } catch (IOException e) {
             LOGGER.log(Level.WARNING, "Cleaning up temporary JENKINS_HOME failed, retrying in 5 sec.", e);
             //maybe process is shutting down, wait for a sec then try again
             try {
                 Thread.sleep(5000);
                 if (jenkinsHome.exists()) {
-                    FileUtils.forceDelete(jenkinsHome);
+                    FileUtils.deleteDirectory(jenkinsHome);
```
amuniz, line 296: Why was this needed?

amuniz (approved):

amuniz:
49 test failures, but they are the same that are currently failing on master.

src/main/java/org/jenkinsci/test/acceptance/controller/LocalController.java
fcojfernandez: Oh! I got a stacktrace there and I made a check. I thought I had reverted it, so doing it now.

src/main/java/org/jenkinsci/test/acceptance/controller/WinstoneController.java
```diff
@@ -26,13 +28,23 @@
...
         if (getenv == null) return Collections.emptyList();
         return Arrays.asList(getenv.split("\\s+"));
     }
 
+    public void addSystemProperty(String systemProperty) {
+        if (StringUtils.isNotBlank(systemProperty)) {
+            if (!systemProperty.startsWith("-D")) {
+                systemProperty = "-D" + systemProperty;
+            }
+            SYSTEM_PROPERTIES.add(systemProperty);
```
jtnord, line 44: I would not expect that someone really creates a `WithSystemProperty("-DFoobar=wibble").  or calls `addSystemPropety("-Dfoo=bar")` 
Otherwise these are not system properties, they are specific java arguments.  It may also make having a systemproperty starting with `-D` interesting (not even sure that is possible however...)

src/main/java/org/jenkinsci/test/acceptance/junit/WithSystemProperties.java
```diff
@@ -0,0 +1,98 @@
...
+        JenkinsController controller;
+
+        @Override
+        public Statement apply(final Statement base, final Description d) {
+            return new Statement() {
+
+                @Override
+                public void evaluate() throws Throwable {
+                    if (!(controller instanceof WinstoneController)) {
+                        throw new AssumptionViolatedException("Test skipped. WithSystemProperties should be used with a local controller.");
```
jtnord, line 53: message is incorrect.  this will be thrown even if the controller is a [`LocalController`](https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/main/java/org/jenkinsci/test/acceptance/controller/LocalController.java#L44)

src/main/java/org/jenkinsci/test/acceptance/junit/WithSystemProperties.java
```diff
@@ -0,0 +1,98 @@
...
+            List<String> systemProperties = new ArrayList<>();
+            for (String property : properties) {
+                if (StringUtils.isNotBlank(property)) {
+                    if (!property.startsWith("-D")) {
+                        property = "-D" + property;
+                    }
+                    systemProperties.add(property);
+                }
+            }
+            return systemProperties.toArray(new String[0]);
```
jtnord, line 95: this is duplicated in many places.  I don't beleive you should even accept this.  assume the property is always `key=value` and use it.  then just blindly always add the `-D` when consuming it.  (or make this actually take a `key` and a `value` array rather than a string.)

src/test/java/org/jenkinsci/test/acceptance/junit/WithSystemPropertiesTest.java
```diff
@@ -0,0 +1,36 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.arrayContainingInAnyOrder;
+import static org.hamcrest.Matchers.emptyArray;
+import static org.junit.Assert.assertNull;
+
+public class WithSystemPropertiesTest {
+
+    @Test
+    public void combine_WithPlugins_annotations() throws Exception {
+        assertThat(WithSystemProperties.RuleImpl.getProperties(FakeTestClass.class.getAnnotation(WithSystemProperties.class))
```
jtnord, line 14: this smells to me.  these are not properties they are property string arguments (ie they have -D).

src/main/java/org/jenkinsci/test/acceptance/controller/WinstoneController.java
fcojfernandez: So, IIUC, `SYSTEM_PROPERTIES` should be renamed to `TEST_JAVA_OPT` (and the method as well) and the add "-D" sentence should be removed, so it's the caller who must add the "-D" 

> It may also make having a system property starting with -D interesting (not even sure that is possible however...)

it is possible. I've tested locally and using a break point, I've accessed the Jenkins page to see the System properties and it was there.

src/main/java/org/jenkinsci/test/acceptance/controller/WinstoneController.java
```diff
@@ -47,6 +59,7 @@ public ProcessInputStream startProcess() throws IOException{
         String java = javaHome == null ? "java" : String.format("%s/bin/java",javaHome.getAbsolutePath());
         CommandBuilder cb = new CommandBuilder(java);
         cb.addAll(JENKINS_JAVA_OPTS);
+        cb.add(SYSTEM_PROPERTIES.toArray(new String[0]));
```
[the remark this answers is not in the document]
fcojfernandez: I'll check again, but my local testing showed that tose properties were add as "prop=val prop=val prop=val" (notice the quotes, so they were ignored.

src/main/java/org/jenkinsci/test/acceptance/controller/WinstoneController.java
jtnord: well that is one route.  :)
the other was to not do the check that it `startsWith` `-D` and add that only at the last minute unconditionally

src/main/java/org/jenkinsci/test/acceptance/controller/WinstoneController.java
```diff
@@ -26,13 +28,20 @@
 
     private static final List<String> JENKINS_JAVA_OPTS = envVarOpts("JENKINS_JAVA_OPTS");
     private static final List<String> JENKINS_OPTS = envVarOpts("JENKINS_OPTS");
+    private final List<String> JAVA_OPTS = new ArrayList<>();
```
jtnord, line 31:
```suggestion
    // options to the JVM that can be added on a testcase basis in code
    private final List<String> JAVA_OPTS = new ArrayList<>();
```

src/main/java/org/jenkinsci/test/acceptance/junit/WithSystemProperties.java
```diff
@@ -0,0 +1,88 @@
...
+ * <p>
+ * The system properties can be set only when the test are being run with a {@link WinstoneController}.
+ * This annotation is a way to extend JENKINS_JAVA_OPTS and JENKINS_OPTS at test level
+ */
+@Retention(RUNTIME)
+@Target({METHOD, TYPE})
+@Inherited
+@Documented
+@RuleAnnotation(value = WithSystemProperties.RuleImpl.class, priority = -10)
+public @interface WithSystemProperties {
```
jtnord, line 37: this is no longer system properties but JVMOptions?

jtnord (commented):
code is called `SystemProperties` yet is actually Java Options.

I do not mind if it is either `SystemProperties` (and then the -D is added only when consumed), or it is arbitrary `JVMoptions` (or a name like that) that need to be fully constructed, but as is it is a mix of boththis mixes

jtnord (approved):

fcojfernandez:
I've looked into the test failures in CI and the only test is not failing in master is `AntPluginTest#autoInstallMultipleAnt`. The stacktrace shows a timeout during the build that is unrelated to this PR. Besides, I cannot reproduce the failure in local using either the weekly or the LTS.
Anyway, I've committed your suggestion @jtnord in 
https://github.com/jenkinsci/acceptance-test-harness/pull/678/files#r686631173 . Let's see again

fcojfernandez:
Tests have finished in CI and the tests failing are the same as master. Can we merge and release this PR, @jtnord ? Thanks in advance!

PR #685, opened 2021-09-14
Adding ability to specify a JAVA_HOME path to the constructor of WinstoneController

kerogers-cloudbees:
This is to allow testing multiple Java versions simultaneously during an acceptance test.
In particular it will allow for testing of communications between Controllers that are operating in different Java environments.

Added a File javaHome to LocalController, Added new constructor for LocalController that sets the javaHome File. Updated getJavaHome() method to check if javaHome has been set and return it if it has. Added constructor to WinstoneController that will call the super constructor with a javaHome File. 

Added a unit test for the WinstoneController constructor that accepts a javaHome. Did not test the constructor without the javaHome because the returned JAVA_HOME path will vary based on where it is run.

### DIFF
--- a/src/test/java/org/jenkinsci/test/acceptance/controller/WinstoneControllerTest.java
+++ b/src/test/java/org/jenkinsci/test/acceptance/controller/WinstoneControllerTest.java
@@ -1,0 +1,46 @@
+package org.jenkinsci.test.acceptance.controller;
+
+
+import com.google.inject.AbstractModule;
+import com.google.inject.Guice;
+import com.google.inject.Injector;
+import com.google.inject.name.Names;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+import java.io.File;
+import java.io.IOException;
+
+public class WinstoneControllerTest {
+    private Injector i;
+    private File tempFile;
+
+    @Before
+    public void setUp() {
+        i = Guice.createInjector(new AbstractModule() {
+            @Override
+            protected void configure() {
+                try {
+                    tempFile = File.createTempFile("junit", "test");
+                } catch (IOException e) {
+                    e.printStackTrace();
+                }
+                bind(String.class).annotatedWith(Names.named("WORKSPACE")).toInstance("/tmp/");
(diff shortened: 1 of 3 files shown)


src/test/java/org/jenkinsci/test/acceptance/controller/WinstoneControllerTest.java
```diff
@@ -0,0 +1,46 @@
...
+    private Injector i;
+    private File tempFile;
+
+    @Before
+    public void setUp() {
+        i = Guice.createInjector(new AbstractModule() {
+            @Override
+            protected void configure() {
+                try {
+                    tempFile = File.createTempFile("junit", "test");
```
jtnord, line 27: are you going to delete this file at all?

src/test/java/org/jenkinsci/test/acceptance/controller/WinstoneControllerTest.java
```diff
@@ -0,0 +1,46 @@
...
+                bind(String.class).annotatedWith(Names.named("quite")).toInstance("");
+                bind(File.class).annotatedWith(Names.named("form-element-path.hpi")).toInstance(tempFile);
+                bind(File.class).annotatedWith(Names.named("jenkins.war")).toInstance(tempFile);
+            }
+        });
+    }
+
+    @Test
+    public void constructWithJavaHome() {
+        File javaHome = new File("/tmp/path/to/java_home");
```
jtnord, line 41: please `/tmp` files and stuff on my machine when I run tests.   `/tmp` is NOT the temp path, and I do not want you to do that.  I happen to use windows so this makes `c:\tmp\path\to\java_home`, but even then the temp directory may not be the standard shared one than some (but not Nixes use).

the windows agents should actually throw an Exception here as you are trying to use something that is not appropriate - but I doubt we use them here, and I also doubt they are setup correctly to prevent this mishcheif 

Also who is cleaning up here

jtnord (commented):
despite the file related issues I am wondering if this is the best place for this.

1> In particular it will allow for testing of communications between Controllers that are operating in different Java environments.

no such thing happens in OSS land at all, and so there is really only one or 2 place where this can be used.
The ATH run for those needs to be configured to Inject the new parameter(s) for the controller that it needs, so could that test harness no be adapted instead (subclass WinstoneController and override JenkinsHome)?

PR #697, opened 2021-10-11
Allow caller to pass a tcp port without relying on the random logic

Vlatombe:
In my case, I need to allocate multiple ports (for multiple Jenkins instance) at the same time, and with the current logic, sometimes I get collisions.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/controller/WinstoneController.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/controller/WinstoneController.java
@@ -47,8 +47,12 @@ public class WinstoneController extends LocalController {
 
     @Inject
     public WinstoneController(Injector i) {
+        this(i, IOUtil.randomTcpPort());
+    }
+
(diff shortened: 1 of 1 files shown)


timja (approved):

src/main/java/org/jenkinsci/test/acceptance/controller/WinstoneController.java
```diff
@@ -47,8 +47,12 @@ public void addJavaOpt(String javaOpt) {
 
     @Inject
     public WinstoneController(Injector i) {
+        this(i, IOUtil.randomTcpPort());
+    }
+
+    public WinstoneController(Injector i, int httpPort) {
```
timja, line 53: do you need `@Inject` on this as well?

Vlatombe: I don't need it because this is being used by subclasses.

PR #705, opened 2021-10-26
Adjustments for new view / new node changes in core part 2

timja:
Fix path detection for radio group


src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java
```diff
@@ -57,6 +57,7 @@
 
     public GitRepo() {
         dir = initDir();
+        git("config", "--global", "init.defaultBranch", "master");
```
jglick, line 60: This does not look safe: if a developer runs this on their host OS, it will overwrite `~/.gitconfig`!

PR #710, opened 2021-11-15
Add no-transfer-progress

timja:


src/main/resources/ath-container/run.sh
```diff
@@ -98,4 +98,4 @@ shift 2
 
 set -x
 
-BROWSER=$browser JENKINS_WAR=$war mvn --show-version test "$@"
+BROWSER=$browser JENKINS_WAR=$war mvn --show-version --no-transfer-progress test "$@"
```
jtnord, line 101: Personally I hate this option with a passion - so will not approve it - very often CI will seemingly "hang"  but its not CI it is artifactory or some mirror behaving very slow - or it is just in the middle of downloading something "large".

At the same time I will not block it either.

timja: @jglick / @MarkEWaite ^^

The main reason to use this argument IIRC is to prevent CI logs being spammed by it, reducing their size and showing you more relevant information.

I'm just adding this to fix the GradleTest in ATH when running through pipeline library.

Its already been removing from pipeline library anyway, so I'm not attached to this

timja:
failing tests should hopefully be fixed by https://github.com/jenkinsci/jenkins/pull/5917

src/main/resources/ath-container/run.sh
MarkEWaite: I prefer to use the `--no-transfer-progress` argument to reduce log spamming.  I accept the risk that @jtnord describes as less of a problem in my use than seeing log spam.

jglick (approved):
> I accept the risk that @jtnord describes as less of a problem in my use than seeing log spam.

Ditto. We have discussed a possible future improvement to Maven to print a warning if a particular transfer seems to be hanging. Have never gotten around to writing it.

PR #712, opened 2021-11-19
Enable form element path from core

timja:
Not sure the best way to disable installing form element path itself, as people install ATH in older versions of core and not sure if we can get the version number at this point, the standard way doesn't work with `Jenkins.getVersion()` as it's not installed in a page object

At the very least we can empty form-element-path and bump the version to the version of core it's released in, unless other ideas

draft till core is released

I've tested that this works on at least a few tests with the core patch and not installing form-element-path plugin

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/controller/LocalController.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/controller/LocalController.java
@@ -131,19 +131,7 @@ public abstract class LocalController extends JenkinsController implements LogLi
     }
 
     private void installFormElementPath(File pluginDir) {
-        if (runInstallWizard) {
-            out.println("Skipping installation of form-element-path.jpi");
-        } else {
-            out.println("Installing form-element-path.jpi");
-            try {
-                FileUtils.copyFile(formElementPathPlugin, new File(pluginDir, "form-element-path.jpi"));
-            } catch (IOException e) {
-                String msg = String.format(
-                        "Failed to copy form-element-path file %s to plugin dir %s.", formElementPathPlugin, pluginDir
-                );
-                throw new RuntimeException(msg, e);
-            }
-        }
+        out.println("Skipping installation of form-element-path.jpi");
     }
 
     @Override
--- a/src/main/java/org/jenkinsci/test/acceptance/controller/LocalController.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/controller/LocalController.java
@@ -131,7 +131,19 @@ public abstract class LocalController extends JenkinsController implements LogLi
     }
 
     private void installFormElementPath(File pluginDir) {
-        out.println("Skipping installation of form-element-path.jpi");
+        if (runInstallWizard) {
+            out.println("Skipping installation of form-element-path.jpi");
+        } else {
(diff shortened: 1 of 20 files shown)


src/main/resources/ath-container/run.sh
```diff
@@ -45,7 +45,7 @@ if [ ! -f $war ]; then
     case "$war" in
         "latest")
             war=jenkins-latest.war
-            url=$mirrors/war/latest/jenkins.war
+            url=https://ci.jenkins.io/job/Core/job/jenkins/job/PR-5926/8/artifact/org/jenkins-ci/main/jenkins-war/2.322-rc31730.8d0c77b9d0c2/jenkins-war-2.322-rc31730.8d0c77b9d0c2.war
```
timja, line 48:
```suggestion
            url=https://ci.jenkins.io/job/Core/job/jenkins/job/PR-5926/10/artifact/org/jenkins-ci/main/jenkins-war/2.322-rc31737.185803a3fe42/jenkins-war-2.322-rc31737.185803a3fe42.war
```

src/main/resources/ath-container/run.sh
```diff
@@ -45,7 +45,7 @@ if [ ! -f $war ]; then
     case "$war" in
         "latest")
             war=jenkins-latest.war
-            url=$mirrors/war/latest/jenkins.war
+            url=https://ci.jenkins.io/job/Core/job/jenkins/job/PR-5926/10/artifact/org/jenkins-ci/main/jenkins-war/2.322-rc31737.185803a3fe42/jenkins-war-2.322-rc31737.185803a3fe42.war
```
timja, line 48:
```suggestion
            url=https://ci.jenkins.io/job/Core/job/jenkins/job/PR-5926/12/artifact/org/jenkins-ci/main/jenkins-war/2.323-rc31744.6335018022ea/jenkins-war-2.323-rc31744.6335018022ea.war
```

src/main/java/org/jenkinsci/test/acceptance/controller/LocalController.java
```diff
@@ -131,19 +131,7 @@ public void postConstruct(Injector injector) {
...
-            try {
-                FileUtils.copyFile(formElementPathPlugin, new File(pluginDir, "form-element-path.jpi"));
-            } catch (IOException e) {
-                String msg = String.format(
-                        "Failed to copy form-element-path file %s to plugin dir %s.", formElementPathPlugin, pluginDir
-                );
-                throw new RuntimeException(msg, e);
-            }
-        }
+        out.println("Skipping installation of form-element-path.jpi");
```
jtnord, line 134: I think before making non draft we need to retain backwards compatibility with Jenkins versions without the inlining (for the LTS) otherwise we won;t be able to run the ATH with new / updated plugin `PageObjects` against LTS versions, or we need to branch and maintain the compatibility.

timja: Yes of course 😄 . This was just to make sure it worked with the core form-element-path because I reworked the code to drop legacy, remove prototype.js and drop updating the dom every 1s.

PR #713, opened 2021-11-21
Fix some tests

timja:
Mostly related to the new tables

One test for the plugin manager deploy from url as well

All tests should pass now, before was 50+


src/test/java/plugins/PrioritySorterPluginTest.java
```diff
@@ -36,8 +34,6 @@
 
     @Before
     public void setUp() throws Exception {
-        assumeTrue("This test requires a restartable Jenkins", jenkins.canRestart());
-        jenkins.restart(); // Priority sorter plugin needs this
```
timja, line 40: doesn't seem to be required these days, and doesn't work on my machine with it as restarting fails on mac m1.

I've ran all the tests in this class and they all pass without the restart

Vlatombe (approved):
LGTM

PR #717, opened 2021-12-08
Click label instead of input for checkbox

timja:
Required for https://github.com/jenkinsci/jenkins/pull/5923
Already applied for radio buttons in
https://github.com/jenkinsci/acceptance-test-harness/pull/704

Browser can't click on the new checkboxes if you ask them to click the input, the label must be clicked instead.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/ByFactory.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/ByFactory.java
@@ -138,7 +138,7 @@ public class ByFactory {
      * Finds checkbox.
      *
      * @param locator
-     *      Text, id, title.
+     *      Text
      */
     public By checkbox(String locator) {
         return xpath(String.format(".//label[contains(normalize-space(.), '%1$s')]", locator));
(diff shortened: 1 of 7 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/Job.java
```diff
@@ -397,7 +397,7 @@ public Build getLastBuild() {
     }
 
     public void disable() {
-        check("disable");
+        check("Disable this project");
```
Vlatombe, line 400: Hum... Does it change the semantic?

src/main/java/org/jenkinsci/test/acceptance/ByFactory.java
```diff
@@ -141,7 +141,7 @@ public By href(String locator){
      *      Text, id, title.
```
timja, line 141:
```suggestion
     *      Text
```

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
timja: it's the label, passes all tests here that use it (quite a few)

I've adjusted javadoc in https://github.com/jenkinsci/acceptance-test-harness/pull/717/commits/57808b234e29b7bdb8f842d611d5a4c6e8b7b05a

Vlatombe (approved):

PR #733, opened 2021-12-31
EOL JSR 305

basil:


timja:
Compilation error. 

> [2021-12-31T18:14:21.119Z] [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile (default-compile) on project acceptance-test-harness: Fatal error compiling: CompilerException: NullPointerException -> [Help 1]

You've probably seen the error before but it's a bug in javac that maven triggers. You can often run another compiler on it to see what it is, e.g. whatever your IDE uses if it doesn't delegate to maven

PR #734, opened 2022-01-06
Validate adding and removing projects to users' favorites

alecharp:
This changeset contains a test to validate that the Favorite plugin can correctly add and remove projects from the users' favorite.

### DIFF
--- a/src/test/java/plugins/FavoriteTest.java
+++ b/src/test/java/plugins/FavoriteTest.java
@@ -1,18 +1,15 @@
 package plugins;
 
+import org.jenkinsci.test.acceptance.Matchers;
 import org.jenkinsci.test.acceptance.junit.AbstractJUnitTest;
 import org.jenkinsci.test.acceptance.junit.WithPlugins;
 import org.jenkinsci.test.acceptance.plugins.mock_security_realm.MockSecurityRealm;
-import org.jenkinsci.test.acceptance.po.Control;
 import org.jenkinsci.test.acceptance.po.FreeStyleJob;
 import org.jenkinsci.test.acceptance.po.GlobalSecurityConfig;
 import org.jenkinsci.test.acceptance.po.User;
 import org.junit.Test;
-import org.openqa.selenium.WebElement;
 
-import java.time.Duration;
-
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
 
 @WithPlugins({"favorite", "mock-security-realm"})
 public class FavoriteTest extends AbstractJUnitTest {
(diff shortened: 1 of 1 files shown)


src/test/java/plugins/FavoriteTest.java
```diff
@@ -0,0 +1,31 @@
...
+        MockSecurityRealm realm = security.useRealm(MockSecurityRealm.class);
+        realm.configure(USER);
+        security.save();
+
+        jenkins.login().doLogin(USER);
+
+        jenkins.open();
+        waitFor(by.id("fav_my-project")).click();
+        jenkins.visit("/user/"+USER+"/configure");
+        waitFor(by.id("fav_my-project")).click();
```
jtnord, line 29: `click` is async and so may or may not complete any javascript if you then navigate away.

Also this test does not validate that it actually works, just that the control was available and clickable?

jtnord (changes requested):
it is unclear that the test is actually testing the functionality vs just checking a control exists.

src/test/java/plugins/FavoriteTest.java
```diff
@@ -0,0 +1,31 @@
...
+        final GlobalSecurityConfig security = new GlobalSecurityConfig(jenkins);
+        security.open();
+        MockSecurityRealm realm = security.useRealm(MockSecurityRealm.class);
+        realm.configure(USER);
+        security.save();
+
+        jenkins.login().doLogin(USER);
+
+        jenkins.open();
+        waitFor(by.id("fav_my-project")).click();
```
raul-arabaolaza, line 27: Something should happen after clicking here that you can check to be sure the relevant js has correctly ejecuted

alecharp: yes. At this moment, the `div#favorites` should be empty. From my tests, the div is not even found by ATH when it's really there (could see it debuging the ATH run).

alecharp:
At this moment, the test is not valid. I just pushed the code to not loose it.

alecharp:
thanks to @raul-arabaolaza, this is now up for review.

raul-arabaolaza (approved):

alecharp:
@timja do you think you could look at this test addition and see if it's good enough to be added into the repository? Thanks.

src/test/java/plugins/FavoriteTest.java
alecharp: This was resolved by the next line as we see something in the page with `icon-fav-active` CSS class.

timja (approved):
@jtnord any opinion?

src/test/java/plugins/FavoriteTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+        jenkins.login().doLogin(USER);
+
+        waitFor(by.id("fav_my-project")).click();
+        waitFor(by.css(".icon-fav-active"));
+
+        final User user = new User(jenkins, USER);
+        jenkins.visit(user.getConfigUrl().toString());
+        waitFor(by.id("favorites"));
+        waitFor(by.id("fav_my-project")).click();
```
jtnord, line 36: should we not have a check that the favoutite is actually checked so that the page here reflects the Jenkins job view.

jtnord (commented):
Seems ok - couple of NITs where `waitFor` is actually part of the test rather than waiting for a page load.

src/test/java/plugins/FavoriteTest.java
alecharp: The line 36 actually does that. If the span with `fav_my-project` is not present on the page, it means that the project is not in the user favorite.

src/test/java/plugins/FavoriteTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+        jenkins.login().doLogin(USER);
+
+        waitFor(by.id("fav_my-project")).click();
+        // ensure the project is now a favourite
+        waitFor(by.css(".icon-fav-active"));
+
+        final User user = new User(jenkins, USER);
+        jenkins.visit(user.getConfigUrl().toString());
+        waitFor(by.id("favorites"));
```
jtnord, line 36:
```suggestion
        // will fail if the project has not been favourited
        waitFor(by.id("favorites"));
```

jtnord (approved):

PR #748, opened 2022-03-02
Compatibility with tango item replacement

timja:
https://github.com/jenkinsci/jenkins/pull/6307

cc @janfaracik

### DIFF
--- a/src/main/resources/ath-container/run.sh
+++ b/src/main/resources/ath-container/run.sh
@@ -45,7 +45,8 @@ if [ ! -f $war ]; then
     case "$war" in
         "latest")
             war=jenkins-latest.war
-            url=$mirrors/war/latest/jenkins.war
+	    # TODO https://github.com/jenkinsci/jenkins/pull/6307
+            url=https://ci.jenkins.io/job/Core/job/jenkins/job/PR-6307/10/artifact/org/jenkins-ci/main/jenkins-war/2.338-rc32220.9a_b_eb_4c259f4/jenkins-war-2.338-rc32220.9a_b_eb_4c259f4.war
         ;;
         "latest-rc")
             war=jenkins-latest-rc.war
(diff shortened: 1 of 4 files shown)


src/main/resources/ath-container/run.sh
```diff
@@ -45,7 +45,8 @@ if [ ! -f $war ]; then
     case "$war" in
         "latest")
             war=jenkins-latest.war
-            url=$mirrors/war/latest/jenkins.war
+	    # TODO https://github.com/jenkinsci/jenkins/pull/6307
+            url=https://ci.jenkins.io/job/Core/job/jenkins/job/PR-6307/10/artifact/org/jenkins-ci/main/jenkins-war/2.338-rc32220.9a_b_eb_4c259f4/jenkins-war-2.338-rc32220.9a_b_eb_4c259f4.war
```
timja, line 49:
```suggestion
            url=$mirrors/war/latest/jenkins.war
```

src/test/java/plugins/StageViewTest.java
```diff
@@ -21,7 +21,7 @@
  * needs adaption for future blue ocean
  * @author Boris Dippolter
  */
-@WithPlugins("workflow-aggregator")
+@WithPlugins({"pipeline-stage-view", "workflow-durable-task-step", "workflow-basic-steps"})
```
jglick, line 24: Seems that this picks up `workflow-job` via `pipeline-stage-view` (likely a design flaw in the latter); what about `workflow-cps` though?

timja: I've had a bit of a look through and have no idea from a quick look of trawling dependencies on https://plugins.jenkins.io/workflow-cps/#dependencies

/

https://plugins.jenkins.io/pipeline-stage-view/#dependencies

jglick: OK…but it passes somehow nonetheless?

timja: Yup passed locally and on CI

PR #752, opened 2022-03-26
Fix SSH Agents on recent core

timja:
Pre is no longer nested as pre -> pre, just pre -> div now.

Plugin manager tests worked fine because it just loads `pre`.

So adjusted to the same


src/test/java/plugins/SshSlavesPluginTest.java
```diff
@@ -210,12 +210,16 @@ private void verifyUnexpectedValueForCredential(String message, CredentialsPage
     @Test public void customJavaPath() {
         setUp();
         SshSlaveLauncher launcher = configureDefaultSSHSlaveLauncher().pwdCredentials("test", "test");
-        
-        launcher.javaPath.set("/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java");
```
timja, line 214: this test failed on arm64 before, adjusted so it works on my machine

PR #753, opened 2022-03-29
Fix credentials and weekly changes

Vlatombe:
Fix failures caused by 
* https://github.com/jenkinsci/credentials-plugin/pull/287 (button is labelled `Create` instead of `OK`). Switched matcher to element name instead.
* and another one due to an added Jenkins logo in svg causing the matcher to fail. Matching the element title instead.

### DIFF
--- a/src/test/java/plugins/DashboardViewPluginTest.java
+++ b/src/test/java/plugins/DashboardViewPluginTest.java
@@ -1,32 +1,27 @@
 package plugins;
 
+import java.util.Arrays;
+import java.util.List;
+import javax.inject.Inject;
 import org.jenkinsci.test.acceptance.junit.WithPlugins;
 import org.jenkinsci.test.acceptance.plugins.dashboard_view.BuildStatisticsPortlet;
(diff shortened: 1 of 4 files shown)


timja (approved):

src/test/java/plugins/DashboardViewPluginTest.java
```diff
@@ -197,126 +192,6 @@ public void latestsBuildsPortlet_onlyLatest() {
...
-        Map<Integer, Integer> colorsOccurrences = getColorDistributionOfSignificantColors(chart.getImage());
-
-        assertThat(colorsOccurrences.size(), is(1));
-
-        Integer color = colorsOccurrences.keySet().iterator().next();
-        int blueComp = (color << 24) >>> 24;
-        int greenComp = (color << 16) >>> 24;
-        int redComp = (color << 8) >>> 24;
-
-        assertThat(blueComp, greaterThan(150));
```
timja, line 228: the colour changed in core, I don't think this is a great test for here to have.

Too fragile to change.

I've left the tests in place that don't look at colours in an image

PR #744, opened 2022-02-10
Test against the LTS as well as the weekly

jtnord:
As we need to run the ATH against LTS version when they are released we
should ensure that the ATH itself has no regressions when run against
it.

This allows us to actually identify issues in the LTS as opposed to the
ATH no longer being compatible with it.
This requires that the ATH be compatible with the weekly and the LTS
line, which should have been mostly done but was never tested in the PR
builder.

As we have recently removed the Java8 test this should not add any extra load onto ci.jenkins.io that was not present a few weeks ago but adding Mark for this as I know the ATH is relatively intensive for CI resources.

There is possible going to be some test failures here that need to be fixed too (so WIP until at least I have a report on failures)

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #754, opened 2022-04-02
Try stabilise builds

timja:
Build appears to be flakey since https://github.com/jenkinsci/acceptance-test-harness/pull/744

Lots of node disconnections


timja:
Hmm seems stable atm, will leave for now

PR #761, opened 2022-04-15
Exchange Jenkins http mirror in favor of https

NotMyFault:
Let's recommend the https mirror in favor over the plain old http one.


timja (approved):
RCs aren't currently published but meh maybe one day they will be.

I verified the latest url works

PR #762, opened 2022-04-16
Adapt to new project configuration page

janfaracik:
PR to fix the tests for new project configuration page: https://github.com/jenkinsci/jenkins/pull/6485

---

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Job.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Job.java
@@ -403,7 +403,7 @@ public class Job extends TopLevelItem {
             check("Disable this project");
         } catch (org.openqa.selenium.NoSuchElementException exception) {
             // Newer versions of Jenkins use a toggle switch with the active state as its label
-            check("Enabled");
+            check(find(by.id("toggle-switch-enable-disable-project")));
         }
     }
 
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Job.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Job.java
@@ -401,7 +401,7 @@ public class Job extends TopLevelItem {
     public void disable() {
         try {
             check("Disable this project");
-        } catch (NoSuchElementException exception) {
+        } catch (org.openqa.selenium.NoSuchElementException exception) {
             // Newer versions of Jenkins use a toggle switch with the active state as its label
             check("Enabled");
         }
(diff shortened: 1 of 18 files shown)


timja:
FTR I did try to fix these tests a couple of weeks back and ran away 😭.
scroller issues are not fun...

janfaracik:
> FTR I did try to fix these tests a couple of weeks back and ran away 😭.
> scroller issues are not fun...

Not really sure why we're getting all these new errors considering the sticky submit bar is pre-existing :/

timja:
> Not really sure why we're getting all these new errors considering the sticky submit bar is pre-existing :/

I think it's because the top bar moved, I think it was accounting for it but I fiddled with numbers for awhile and wasn't able to get it working.

> Not sure if this would be okay on Jenkins, but on past projects we've disabled sticky elements in tests to get around this (e.g. setting the position from sticky to relative)

I can't see an issue with that

janfaracik:
Pushed a commit that adds a rough (not working) script to disable sticky elements and smooth scrolling, it works except for the configure page, where I think it takes too long to activate due to how much JS is being loaded etc

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
```diff
@@ -399,7 +399,12 @@ public <T extends Parameter> T addParameter(Class<T> type) {
     }
 
     public void disable() {
-        check("Disable this project");
+        try {
+            check("Disable this project");
+        } catch (org.openqa.selenium.NoSuchElementException exception) {
+            // Newer versions of Jenkins use a toggle switch with the active state as its label
+            check(find(by.id("toggle-switch-enable-disable-project")));
```
timja, line 406: normally we default to the new version and fallback to old versions

janfaracik:
> This seems to hide the submit bar when loading a configuration page (in a proprietary CloudBees plugin) Capture d’écran 2022-05-10 à 15 41 03 causing a test error when running https://github.com/jenkinsci/acceptance-test-harness/pull/762/files#diff-10617378e7d7b2709164848eb1587225b855270efae2e7354f596f74f4211659R114
> 
> The bottom bar then reappears as soon as some scrolling is done. If I understand correctly, this disabling of sticky elements is intended to prevent form fields from appearing below the bar?

Hey! Is there any chance you could share the source for the bottom submit bar? It should 🤞 work fine as long as it's using the default `bottomButtonBar.jelly` without any custom JS/styling.

PR #777, opened 2022-05-27
Switch to using MSBuild as an non suggested wizard plugin

jtnord:
CloudBees has its own list for platform-plugins and `locale` is not in
that list.
Looking between the 2 MSBuild is in both and suggested in neither so
switch to that plugin.


cathychan (approved):

timja (approved):

jtnord:
multiple failures all due to a timout with `installPlugin`

I have seen that this code path waits for a refresh of the UC metadata which can often time out.  Not sure why we do this - Jenkins performs an update as soon as it starts ([code](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/jenkins/DailyCheck.java#L65-L67)) and we should force an UC refresh if we [install the mock update center](https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/main/java/org/jenkinsci/test/acceptance/update_center/MockUpdateCenter.java#L205)
Then we should just check that the last metadata refresh time is within the window (if not we can force a check).  I think this could improve the test execution time (not by a huge amount - but obtaining and parsing the UC is not "free").  

Having said that - this does bring in some condiftion different behaviour into the test framework which could make things harder to reproduce.  WDYT @Vlatombe @timja ?

Vlatombe:
There is a problem with UC infra happening right now fyi cc @dduportal

jtnord:
```
12:04:01  [WARNING] Corrupted STDOUT by directly writing to native stream in forked JVM 1. See FAQ web page and the dump file /home/jenkins/workspace/e_acceptance-test-harness_PR-777/target/surefire-reports/2022-05-30T10-50-07_962-jvmRun1.dumpstream
12:04:01  [INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 65.67 s - in plugins.AuditTrailPluginTest
12:04:01  [INFO] Running plugins.JiraPluginTest
12:08:37  [INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 274.594 s - in plugins.JiraPluginTest
12:08:37  [INFO] Running plugins.XvncPluginTest
12:10:48  Cannot contact ubuntu-20-highmem458611: java.lang.InterruptedException
```

infra issue?

timja:
> infra issue?

Fairly common on ATH 😢 
Often have to run them multiple times

timja:
> Cannot contact ubuntu-20-highmem458611: java.lang.InterruptedException

Yes that's the 'infra' issue.

> At first sight, it seems to me that the agent process died on the VM: does is correlate with what you see?

🤷

timja:
https://github.com/jenkinsci/azure-vm-agents-plugin/pull/359

dduportal:
> Dunno why it gets disconnected, at a guess remoting process has too many resources and ATH steals enough that the OS kills the process. But would need someone to collect logs and metrics after it happens and before the VM gets deleted which is likely not much time.
> 
> (something like opentelemetry may be useful?)
> 
> I've copied the code from the default retention strategy in Jenkins core that is what makes this work for regular SSH launchers: [jenkinsci/azure-vm-agents-plugin#359](https://github.com/jenkinsci/azure-vm-agents-plugin/pull/359)

To be sure that we catch metrics and data outside the context of Jenkins itself with opentelemetry, we could also install the datadog agent, and configure Jenkins so that it starts the VM providing the datadog APi key and start the agent in cloud init, so we could get more informations. WDYT?

timja:
yes that would be helpful. so we could see what's going on CPU / memory wise when it dies

PR #779, opened 2022-06-01
Add vars.cmd to ease setting up an environment on windows

jtnord:
I have this script locally in another test repo that also uses the ATH.  It's probably useful to anyone else wanting to run the ATH on windows (with WSL) so add it here for others

### DIFF
--- a/vars.cmd
+++ b/vars.cmd
@@ -1,0 +1,21 @@
+@REM script for setting all the variables in order to run the ATH locally on windows
+@REM use config.groovy in the same folder if desired uncomment below
+@REM set CONFIG=%~dp0%config.groovy
+set DISPLAY=:0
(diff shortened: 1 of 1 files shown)


vars.cmd
```diff
@@ -0,0 +1,21 @@
+@REM script for setting all the variables in order to run the ATH locally on windows
+@REM use config.groovy in the same folder if desired uncomment below
+@REM set CONFIG=%~dp0%config.groovy
```
jtnord, line 3: this script does not disable the `echo` output using `@echo off` so that consumers can see what is set.  hence things that should not be echoed start with `@`

timja (approved):
I guess? I don't really understand it but haven't used windows in a long time

PR #792, opened 2022-06-07
Add a timeout value parameter to PluginManager.checkForUpdates. 

kerogers-cloudbees:
New parameter to allow for adjusting of timeout duration during testing. Overload parameterless checkForUpdates to use the 30 second timeout as before. 

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
@@ -86,15 +86,10 @@ public class PluginManager extends ContainerPageObject {
         this.jenkins = jenkins;
     }
 
-    public void checkForUpdates() {
-        checkForUpdates(30L);
-    }
-
     /**
      * Force update the plugin update center metadata.
-     * @param timeOutSeconds seconds to wait before timeout.
      */
-    public void checkForUpdates(long timeOutSeconds) {
+    public void checkForUpdates() {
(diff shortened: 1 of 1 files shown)


timja:
why would you need this?

kerogers-cloudbees:
The 30 second timeout isn't long enough when the custom update center in the OC-ATH. The custom update center tests already override the ElasticTime page timeouts, but checkForUpdates() isn't using the timeout in the selenium driver.

timja (approved):
fine i guess but concerning that you need to increase this.

without understanding the details it may be better to fix the performance there as it will slow down a lot of your tests I would expect

jtnord:
Similar to the comment from @timja 

When our tests are failing for this - I do not beleive making it longer will help - our UpdateCenter is hosed and will eventually return HTTP 504.  (as noted by datadog monitoring)

Have you actually shown that this does fix the case you are intending to fix?

now - I have noted that we are calling the updatecenter refresh unescessarily in the majority of cases - as Jenkins now does a refresh when it starts up. https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/jenkins/DailyCheck.java   so calling refresh here may _seem_ to be a bit redundant?

Vlatombe:
To me it looks like a problem on the UC backend side (the failing test interacts with a custom UC). The time spent doesn't seem to come from the client side.

Vlatombe:
@kerogers-cloudbees This can  be closed now.

Issue #745, opened 2022-02-11
GitRepo does not work on windows


jtnord:
### Jenkins and plugins versions report

[GitRepo](https://github.com/jenkinsci/acceptance-test-harness/blob/a9027d40be247fbc51f4ea4af7e4bd0bd7f7e6e8/src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java#L98) fails on windows because 
1. `posixPermissions` are not supported and it throws a `java.lang.UnsupportedOperationException`
2. even if that silently worked or was swallowed, the SSH executable is using `/bin/sh`

### What Operating System are you using (both controller, and any agents involved in the problem)?

Windows

### Reproduction steps

the github template is not suitable for this project

### Expected Results

the github template is not suitable for this project

### Actual Results

the github template is not suitable for this project

PR #805, opened 2022-06-29
enable `GitRepo` to work on windows

jtnord:
fixes: #745

tested with `mvn test -Dtest=plugins.WorkflowPluginTest#hello_world_from_git` locally (windows)  ([source](https://github.com/jtnord/acceptance-test-harness/blob/0f5270fd68e2d278188cf6ebd911a13de418d514/src/test/java/plugins/WorkflowPluginTest.java#L99-L119))

There will be some environments where this does not work (no ssh.exe is available - and then if your git global config is set to use `plink` (an option when installing) then whilst gitRepo works, tests will eventual fail due to the git-client plugin not supporting this, and also https://issues.jenkins.io/browse/JENKINS-71122 is ssh.exe is on the path but not in a "well known" location (workaround is to set GIT_SSH to the path of the ssh executable).

these issues are either environmental or bugs in a plugin not the ATH.  this works in more cases that it did before so considering this "good engough"

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java
@@ -9,8 +9,9 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.net.URISyntaxException;
 import java.net.URL;
-import java.nio.charset.StandardCharsets;
+import java.nio.charset.Charset;
 import java.nio.file.DirectoryStream;
+import java.nio.file.FileSystem;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -19,8 +20,8 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Properties;
-
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.SystemUtils;
 import org.zeroturnaround.zip.ZipUtil;
 
 import com.jcraft.jsch.ChannelExec;
@@ -34,7 +35,6 @@ import org.jenkinsci.test.acceptance.docker.fixtures.GitContainer;
 
 import static java.lang.ProcessBuilder.Redirect.*;
 import static java.nio.file.attribute.PosixFilePermission.*;
-import static java.util.Collections.*;
 import static org.jenkinsci.test.acceptance.docker.fixtures.GitContainer.*;
 
 /**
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java
```diff
@@ -94,16 +94,22 @@ public void setIdentity(String userName, String userMail) {
...
 
-            ssh = File.createTempFile("jenkins", "ssh");
-            FileUtils.writeStringToFile(ssh,
-                    "#!/bin/sh\n" +
-                            "exec ssh -o StrictHostKeyChecking=no -i " + privateKey.getAbsolutePath() + " \"$@\"" , StandardCharsets.UTF_8);
-            Files.setPosixFilePermissions(ssh.toPath(), new HashSet<>(Arrays.asList(OWNER_READ, OWNER_EXECUTE)));
+            FileUtils.copyURLToFile(GitContainer.class.getResource("GitContainer/unsafe"), privateKey);
 
+            if (SystemUtils.IS_OS_WINDOWS) {
+                ssh = Files.createTempFile("jenkins", "ssh.cmd").toFile();
```
jtnord, line 102: windows does not have an `executable` bit per say - it is based on the extension so ensure we end with `.cmd`

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java
```diff
@@ -94,16 +94,22 @@ public void setIdentity(String userName, String userMail) {
...
+                // this works if ssh is on the path, for example
+                // https://learn.microsoft.com/en-us/windows-server/administration/openssh/openssh_install_firstuse
+                Files.writeString(ssh.toPath(), "ssh.exe -o StrictHostKeyChecking=no -i " + privateKey.getAbsolutePath() + " %*", Charset.defaultCharset());
```
jtnord, line 105: both the linux and windows implementations use the default charset as this is run by tooling that may be not using `UTF-8`

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java
```diff
@@ -198,7 +204,7 @@ public void commit(final String message) {
 
     public void touch(final String fileName) {
         try {
-            FileUtils.writeStringToFile(file(fileName), "", StandardCharsets.UTF_8);
+            FileUtils.writeStringToFile(file(fileName), "", Charset.defaultCharset());
```
jtnord, line 207: although both produce the same output (it is an empty string after all) - expecting this to be used by native tooling to use the native charset.

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java
```diff
@@ -291,18 +297,21 @@ public void transferToDockerContainer(String host, int port) {
             ChannelSftp channel = (ChannelSftp) session.openChannel("sftp");
             channel.connect();
             channel.cd("/home/git");
-            channel.put(Files.newInputStream(zippedRepo.toPath()), zippedFilename);
+            try (InputStream fis = Files.newInputStream(zippedRepo.toPath())) {
```
jtnord, line 300: prevent a file leak, which causes random failures to clean up after the test

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java
```diff
@@ -291,18 +297,21 @@ public void transferToDockerContainer(String host, int port) {
...
             InputStream in = channelExec.getInputStream();
             channelExec.setCommand("unzip " + zippedFilename + " -d " + REPO_NAME);
             channelExec.connect();
 
-            BufferedReader reader = new BufferedReader(new InputStreamReader(in));
-            String line;
-            int index = 0;
-            while ((line = reader.readLine()) != null) {
-                System.out.println(++index + " : " + line);
+            try (BufferedReader reader = new BufferedReader(new InputStreamReader(in))) {
```
jtnord, line 309: prevent a file leak, which causes random failures to clean up after the test

jtnord:
image

PR #812, opened 2022-07-06
Use supported URL for fetching latest WARs

timja:
see https://github.com/jenkins-infra/helpdesk/issues/3034


jtnord (commented):
what about the release candidates?  is there no pipeline that tests these using this script?

timja:
> what about the release candidates? is there no pipeline that tests these using this script?

release candidates don't run in this repo they haven't been published to those locations in > 2 years.

If ATH is run against the RCs it's in the jenkinsci/jenkins repo using the essentials.yml file

timja:
(INFRA team is looking at fixing get.jenkins.io for this, we may swap back but this is safer for now)

PR #860, opened 2022-07-11
make Build.isInProgress also cover post building stages. (deflake core.FreestyleJobTest#discardBuilds)

jmdesprez:

`FreestyleJobTest#discardBuilds` is flaky because of `org.jenkinsci.test.acceptance.po.Build#waitUntilFinished()` is not waiting for the build to be actually finished but completed. This is related to https://github.com/jenkinsci/jenkins/pull/6829 and [JENKINS-68981](https://issues.jenkins.io/browse/JENKINS-68981) as there is currently no (easy/reliable) way to wait for a build to be finished.

The problem is that `org.jenkinsci.test.acceptance.po.Build#waitUntilFinished()` exits before [logRotate](https://github.com/jenkinsci/jenkins/blob/aba1c17edee4877bca4253a2a6f4a322b2b4973d/core/src/main/java/hudson/model/Run.java#L1963) -> [LogRotator#perform](https://github.com/jenkinsci/jenkins/blob/aba1c17edee4877bca4253a2a6f4a322b2b4973d/core/src/main/java/hudson/tasks/LogRotator.java#L145) gets called, leading to a race condition

If `assertThat(b2, pageObjectDoesNotExist());` is called before `LogRotator#perform` then the test fails with this stacktrace
```
java.lang.AssertionError: Expected: Page object exists
     but: http://mvn:64648/job/bald_ski/2/ does exist
at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:8)
at core.FreestyleJobTest.discardBuilds(FreestyleJobTest.java:200)
```
Running the ATH with `INTERACTIVE=true`, shows that the build is removed as expected.

The workaround add a timeout to mitigate the race condition.


src/main/java/org/jenkinsci/test/acceptance/Matchers.java
```diff
@@ -177,7 +177,7 @@ public boolean matchesSafely(PageObject item) {
     }
 
     public static Matcher<PageObject> pageObjectDoesNotExist() {
-        return new Matcher<PageObject>("Page object exists") {
+        return new Matcher<PageObject>("Page object does not exists") {
```
timja, line 180:
```suggestion
        return new Matcher<PageObject>("Page object does not exist") {
```

jtnord (approved):

src/main/java/org/jenkinsci/test/acceptance/po/Build.java
```diff
@@ -131,6 +131,7 @@ public boolean isInProgress() {
         }
 
         JsonNode d = getJson();
+        // TODO use "inProgress" if https://github.com/jenkinsci/jenkins/pull/6829 ever reached the LTS (see also JENKINS-68981)
```
jglick, line 134: Well, no, you should use https://github.com/jenkinsci/jenkins/pull/6829 right away if the field is defined. If it is not, then you may have fall back to the old implementation, with a TODO note to clean that up eventually.

src/main/java/org/jenkinsci/test/acceptance/po/Build.java
```diff
@@ -131,8 +131,15 @@ public boolean isInProgress() {
         }
```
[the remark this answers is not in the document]
jtnord: correct @jmdesprez and I discussed that outside of this PR.
The issue is that `inProgress` will return `false` if the build has not yet started `return state.equals(State.BUILDING) || state.equals(State.POST_PRODUCTION);`  but existing users of this may be expecting the old behaviour.

That said your comment prompted me to do a quick github search of this repo and CloudBees' proprietery ones and there are only 5 callers that I found - and all of those seem to be protected by `waitForStarted` (would need validation) with the exception of the single CLoudBees proprietry one, so I think the change makes sense. (it does after all match the name of the function!)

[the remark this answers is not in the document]
jglick: Not sure why this was resolved—according to the above comment it is still a feasible suggestion? Something like

```java
    public boolean isInProgress() {
        JsonNode d = getJson();
        JsonNode inProgress = d.get("inProgress");
        if (inProgress != null) {
            // see https://github.com/jenkinsci/jenkins/pull/6829
            return inProgress.booleanValue();
        }

        // TODO delete when inProgress becomes part of the LTS
        if (result != null) {
            return false;
        }
        if (!hasStarted()) {
            return false;
        }
        // Please note that the test below denote a completed build, not a finished build
        return d.get("building").booleanValue() || d.get("result") == null;
    }
```
which would clarify that going forward we expect this page object method to have identical behavior to the Java method inside Jenkins, and that the complicated logic is just a workaround for older versions.

src/main/java/org/jenkinsci/test/acceptance/po/Build.java
```diff
@@ -123,14 +123,24 @@ public String diagnose(Throwable lastException, String message) {
...
-        if (result != null) {
+        JsonNode d;
+        try {
+            d = getJson();
+            JsonNode inProgress = d.get("inProgress");
+            if (inProgress != null) {
+                // see https://github.com/jenkinsci/jenkins/pull/6829
+                return inProgress.booleanValue();
+            }
+        } catch (Exception e) {
```
jglick, line 134: What exception is expected here?

jglick:
There was never any response to https://github.com/jenkinsci/acceptance-test-harness/pull/860#discussion_r943387134. OK to reject with some reason, but it is not clear the comment was even seen.

src/main/java/org/jenkinsci/test/acceptance/po/Build.java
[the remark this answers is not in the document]
jmdesprez: test

jmdesprez:
> There was never any response to [#860 (comment)](https://github.com/jenkinsci/acceptance-test-harness/pull/860#discussion_r943387134). OK to reject with some reason, but it is not clear the comment was even seen.

I'm sorry, my comments were in `Pending` state :cry:

jglick:
I did not read the response in much detail and I do not really care that much about this whole test suite to begin with (only got here via core PR), but you seem to have thought about it so good enough from my PoV.

src/main/java/org/jenkinsci/test/acceptance/po/Build.java
```diff
@@ -117,14 +118,24 @@ public String diagnose(Throwable lastException, String message) {
...
+        } catch (NoSuchElementException e) {
+            // Build has not started, so it is not in progress.
             return false;
         }
-        if (!hasStarted()) {
+
+        // TODO delete when inProgress becomes part of the LTS
+        if (result != null) {
             return false;
         }
```
jtnord, line 137: @jmdesprez this is now part of the LTS.  (available in 2.369+)
```suggestion
```

src/test/java/core/FreestyleJobTest.java
```diff
@@ -197,6 +197,10 @@ public void discardBuilds() {
         b2.keepForever(false);
 
         j.scheduleBuild().waitUntilFinished();
+        // TODO remove as soon as 'waitUntilFinished' effectively waiting for the build to be finished (including hudson.model.Job#logRotate)
+        waitFor(b2).withMessage("waiting for build #2 to be deleted")
+                .withTimeout(30, TimeUnit.SECONDS).until(pageObjectDoesNotExist());
+
```
jtnord, line 203: as the change is merged upstream this can now be deleted I beleive

```suggestion
```

jmdesprez:
@jtnord thanks for the suggestions
I have removed the no longer needed workarounds.

src/main/java/org/jenkinsci/test/acceptance/po/Build.java
```diff
@@ -117,15 +118,15 @@ public String diagnose(Throwable lastException, String message) {
     }
 
     public boolean isInProgress() {
```
jtnord, line 120:
```suggestion
    /** Checks if the build is in progress,  that is the build has started and not yet completed. */
    public boolean isInProgress() {
```

jtnord:
2 failures looks related

```
java.lang.IllegalArgumentException: No enum constant org.jenkinsci.test.acceptance.po.Build.Result.null
	at java.base/java.lang.Enum.valueOf(Enum.java:240)
	at org.jenkinsci.test.acceptance.po.Build$Result.valueOf(Build.java:28)
	at org.jenkinsci.test.acceptance.po.Build.getResult(Build.java:192)
	at org.jenkinsci.test.acceptance.po.Build$2.matchesSafely(Build.java:240)
	at org.jenkinsci.test.acceptance.po.Build$2.matchesSafely(Build.java:237)
	at org.hamcrest.TypeSafeMatcher.matches(TypeSafeMatcher.java:67)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:10)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
	at org.jenkinsci.test.acceptance.po.Build.shouldSucceed(Build.java:212)
```

jtnord:
> 2 failures looks related
> 
> ```
> java.lang.IllegalArgumentException: No enum constant org.jenkinsci.test.acceptance.po.Build.Result.null
> 	at java.base/java.lang.Enum.valueOf(Enum.java:240)
> 	at org.jenkinsci.test.acceptance.po.Build$Result.valueOf(Build.java:28)
> 	at org.jenkinsci.test.acceptance.po.Build.getResult(Build.java:192)
> 	at org.jenkinsci.test.acceptance.po.Build$2.matchesSafely(Build.java:240)
> 	at org.jenkinsci.test.acceptance.po.Build$2.matchesSafely(Build.java:237)
> 	at org.hamcrest.TypeSafeMatcher.matches(TypeSafeMatcher.java:67)
> 	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:10)
> 	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
> 	at org.jenkinsci.test.acceptance.po.Build.shouldSucceed(Build.java:212)
> ```

a3571c8156a872f9b72904592b68f0338878606b

PR #861, opened 2022-07-11
Remove prelaunching functionality

basil:
Does anyone use the prelaunching functionality? While useful as an optimization, the implementation is fairly complex and results in the following dependencies:

- `jnr-unixsocket`
- `jffi`
- Remoting

If nobody is using this functionality, deleting it could be a good way to simplify the dependency tree and ease maintenance.

If somebody is actively using this functionality, this PR can be closed.

If we are not sure, a compromise approach could be to remove the functionality by merging this PR and restore it later if somebody complains about its removal.


timja (approved):

uhafner:
👎

Puh, seems that I missed that topic. I always using this functionality. This is very helpful feature while developing test! In my university testing courses this helps to start new Jenkins instances while developing new tests.

timja:
revert it?

PR #865, opened 2022-07-22
ATH should not be Java11 only (yet)

jtnord:
The parent was updated to produce java11 bytecode in https://github.com/jenkinsci/acceptance-test-harness/pull/851#issuecomment-1192377049.
We should not yet be requiring Java11, indeed even when Jenkins is Java11 there is no rush to move to Java11 until we want to take advantage of a java11 library or use java11 features.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -40,12 +40,8 @@
     <mockito.version>1.10.19</mockito.version>
     <slf4j.version>1.7.36</slf4j.version>
     <!-- TODO remove these when we can finally switch to Java 11 requirements -->
-    <maven.compiler.source>8</maven.compiler.source>
-    <maven.compiler.target>8</maven.compiler.target>
-    <maven.compiler.testSource>8</maven.compiler.testSource>
-    <maven.compiler.testTarget>8</maven.compiler.testTarget>
-    <maven.compiler.release/>
-    <maven.compiler.testRelease/>
+    <maven.compiler.release>8</maven.compiler.release>
+    <maven.compiler.testRelease>8</maven.compiler.testRelease>
   </properties>
 
   <repositories>
@@ -574,14 +570,6 @@
                   <version>[3.8.1,)</version>
                   <message>3.8.1 required to no longer download dependencies via HTTP (use HTTPS instead).</message>
                 </requireMavenVersion>
-                <!-- TODO failing during incrementals deploy: MENFORCER-281
-                <requirePluginVersions>
-                  <banSnapshots>false</banSnapshots>
-                </requirePluginVersions>
-                -->
-                <requireJavaVersion>
-                  <version>[1.8.0,]</version>
-                </requireJavaVersion>
                 <enforceBytecodeVersion>
                   <maxJdkVersion>1.8</maxJdkVersion>
                 </enforceBytecodeVersion>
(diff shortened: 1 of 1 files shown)


olamy (approved):

timja (approved):

uhafner:
Why is the compile and runtime for the ATH not totally independent of the Jenkins runtime requirements?

timja:
> Why is the compile and runtime for the ATH not totally independent of the Jenkins runtime requirements?

It was a change in the parent Pom that bumped the version wasn’t intentional

NotMyFault (approved):

uhafner:
> > Why is the compile and runtime for the ATH not totally independent of the Jenkins runtime requirements?
> 
> It was a change in the parent Pom that bumped the version wasn’t intentional

Yes I have seen that. My question is why we are not even using Java 17 here in ATH? My UI test maven modules have a parent pom that is totally independent from Jenkins parents.

jtnord:
> Yes I have seen that. My question is why we are not even using Java 17 here in ATH? My UI test maven modules have a parent pom that is totally independent from Jenkins parent poms.

If we used java17 here it would force that on all users of it.  You would no longer be able to run your multi-module project with anything lower.   
By default for `LocalControllers` it would also launch the Jenkins version using that JVM so wihtout jumping through hoops you would not be able to test a version of Jenkins on master passes the ATHs for Java 11,17 or anywhere in between.  Jenkins LTS still supports java8 so it is important to be able to test that - and plugin tests too may want to be able to test against an older the Jenkins version using java8 (as well as a more recent Jenkins version using some other JVM).

NotMyFault:
Auto merging is blocked by https://github.com/jenkinsci/acceptance-test-harness/issues/866, FTR.

jtnord:
> > > Yes I have seen that. My question is why we are not even using Java 17 here in ATH? My UI test maven modules have a parent pom that is totally independent from Jenkins parent poms.
> > 
> > 
> > If we used java17 here it would force that on all users of it. You would no longer be able to run your multi-module project with anything lower. By default for `LocalControllers` it would also launch the Jenkins version using that JVM so without jumping through hoops you would not be able to test a version of Jenkins on master passes the ATHs for Java 11,17 or anywhere in between. Jenkins LTS still supports java8 so it is important to be able to test that - and plugin tests too may want to be able to test against an older the Jenkins version using java8 (as well as a more recent Jenkins version using some other JVM).
> 
> I see. I was not aware that the architecture of the ATH is so tightly coupled with the architecture of Jenkins. My previous understanding was that the ATH runs Jenkins instances in new separate processes with any JVM you want to use.

[it can do](https://github.com/jenkinsci/acceptance-test-harness/blob/4b1d54f3523a54aa7de455f0422a6a43d59691c3/src/main/java/org/jenkinsci/test/acceptance/controller/LocalController.java#L179) - but I have never seen it run that way.   Generally it is just easier to use the same JVM to run the ATH and the Jenkins instance,

jtnord:
failing test is flaky - given this is not a new flake introduced here if the next build passes I am going to "ship-it" as there are downstream libraries need a release that includes the updates for `matrix-auth`

jtnord:
should pass tests once https://github.com/jenkinsci/acceptance-test-harness/pull/867 is merged.

PR #868, opened 2022-07-25
Adapt changes to breadcrumbs accessibility

NotMyFault:
https://github.com/jenkinsci/jenkins/pull/6912


janfaracik:
Tests should be passing now, just the `ul` to `ol` failure was caused by the breadcrumbs PR. Think something else broke the dashboard view tests but those are fixed now too 👍 

Cheers for running them @NotMyFault :)

NotMyFault:
> Tests should be passing now, just the `ul` to `ol` failure was caused by the breadcrumbs PR. Think something else broke the dashboard view tests but those are fixed now too 👍


Thanks for addressing, `plugins.DashboardViewPluginTest.unstableJobsPortlet_showOnlyFailedJobs` didn't work out unfortunately :/

Jenkinsfile
```diff
@@ -27,7 +27,7 @@ if (needSplittingFromWorkspace) {
 for (int i = 0; i < splits.size(); i++) {
   int index = i
   for (int j in [11]) {
-    for (String v in ['lts', 'latest']) {
+    for (String v in ['latest']) {
```
NotMyFault, line 30:
```suggestion
    for (String v in ['lts', 'latest']) {
```

src/main/resources/ath-container/run.sh
```diff
@@ -44,7 +44,7 @@ if [ ! -f $war ]; then
     case "$war" in
         "latest")
             war=jenkins-latest.war
-            url=https://updates.jenkins.io/latest/jenkins.war
+            url=https://repo.jenkins-ci.org/incrementals/org/jenkins-ci/main/jenkins-war/2.361-rc32700.a_d891ddb_da_0c/jenkins-war-2.361-rc32700.a_d891ddb_da_0c.war
```
NotMyFault, line 47:
```suggestion
            url=https://updates.jenkins.io/latest/jenkins.war
```

NotMyFault (commented):
Needs to be changed before merging

Jenkinsfile
jtnord: if the change is going to be merged upstream the ATH needs to be compatible with the old way of doing things and the new way.  having a pre-tested change that works both with and without the PR will minimze the time to adapt if the upstream is merged (there would not be any possibility of regressions here).

src/main/java/org/jenkinsci/test/acceptance/plugins/dashboard_view/read/BreadCrumbs.java
```diff
@@ -18,7 +18,7 @@ public class BreadCrumbs extends PageAreaImpl {
     /**
      * The path to the breadcrumbs list.
      */
-    final By nameCrumb = By.xpath("//ul[@id=\"breadcrumbs\"]");
+    final By nameCrumb = By.xpath("//ol[@id=\"breadcrumbs\"]");
```
jtnord, line 21: As I understand the discussion this is for the new breadcrumb - but this will break without the PR and it should be compatable with Jenkins both with and without the upstream PR.

jtnord (changes requested):

src/main/java/org/jenkinsci/test/acceptance/plugins/dashboard_view/read/BreadCrumbs.java
timja:
```suggestion
    final By nameCrumb = By.xpath("//ol[@id=\"breadcrumbs\"] | //ul[@id=\"breadcrumbs\"]");
```

timja:
Tests pass locally for me so hopefully updating to a new build sorts it

PR #869, opened 2022-07-25
Adapt to removing items from sidebar

NotMyFault:
https://github.com/jenkinsci/jenkins/pull/6907

Core looks like a solid change, just checking incase something depends on it...

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -27,7 +27,7 @@ if (needSplittingFromWorkspace) {
 for (int i = 0; i < splits.size(); i++) {
   int index = i
   for (int j in [11]) {
-    for (String v in ['lts', 'latest']) {
+    for (String v in ['latest']) {
       int javaVersion = j
       String jenkinsUnderTest = v
       def name = "java-${javaVersion}-jenkins-${jenkinsUnderTest}-split${index}"
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -27,7 +27,7 @@ if (needSplittingFromWorkspace) {
 for (int i = 0; i < splits.size(); i++) {
   int index = i
   for (int j in [11]) {
-    for (String v in ['latest']) {
+    for (String v in ['lts', 'latest']) {
       int javaVersion = j
       String jenkinsUnderTest = v
       def name = "java-${javaVersion}-jenkins-${jenkinsUnderTest}-split${index}"
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Slave.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Slave.java
@@ -1,5 +1,6 @@
 package org.jenkinsci.test.acceptance.po;
 
+import hudson.util.VersionNumber;
 import java.util.regex.Pattern;
 
 import org.jenkinsci.test.acceptance.Matcher;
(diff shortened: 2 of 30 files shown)


timja:
2 to go and making sure it still works on LTS :)

NotMyFault:
> 2 to go and making sure it still works on LTS :)

Looks like we can strike out the 2 to go 👀

Jenkinsfile
```diff
@@ -27,7 +27,7 @@ if (needSplittingFromWorkspace) {
 for (int i = 0; i < splits.size(); i++) {
   int index = i
   for (int j in [11]) {
-    for (String v in ['lts', 'latest']) {
+    for (String v in ['latest']) {
```
timja, line 30:
```suggestion
    for (String v in ['lts', 'latest']) {
```

src/main/java/org/jenkinsci/test/acceptance/po/Slave.java
```diff
@@ -21,7 +22,7 @@ public class Slave extends Node {
     private final String name;
 
     public Slave(Jenkins j, String name) {
-        super(j, j.url("computer/%s/",name));
+        super(j, j.getVersion().isNewerThan(new VersionNumber("2.363")) ? j.url("manage/computer/%s/", name) : j.url("computer/%s/", name));
```
timja, line 25: can't see a smooth way to do this other than versions so will need this to be adapted if it misses the next weekly.

Most tests passed with the initial PR in 2.360 but one failed as there was some more work done on top of the feature in the core PR for this

src/test/java/core/SlaveTest.java
```diff
@@ -40,6 +41,7 @@
 /**
  * @author Orjan Percy <orjan.percy@sonymobile.com>
  */
+@WithPlugins("command-launcher")
```
timja, line 44: technical debt from https://github.com/jenkinsci/acceptance-test-harness/blob/e3801a6803342ac57234aa3757004ad978ca2935/src/main/java/org/jenkinsci/test/acceptance/po/DumbSlave.java#L38-L48

all `SlaveController` using `DumbSlave` at least in the default setup require this plugin.

It caused a couple of flakes where the save button couldn't be clicked when it was trying to execute the hack, not strictly related to this PR but added for extra stability.

timja:
@NotMyFault want to check my changes? (aware you can't leave a ✅  due to me hijacking your PR)

NotMyFault:
Looks good ✅

jtnord (approved):

src/main/java/org/jenkinsci/test/acceptance/po/Slave.java
daniel-beck: @timja For posterity, could you be more specific why this change?

timja: Tests were asserting the expected url but they were now getting a different one from links changing to include `manage`

correct_redirect_after_save – plugins.OwnershipPluginTest
```
Expected: <http://127.0.0.1:32861/computer/precise_cluster/>
     but: was <http://127.0.0.1:32861/manage/computer/precise_cluster/>
```

latestsBuildsPortlet_correctJobAndBuild – plugins.DashboardViewPluginTest
```
Expected: is <true>
     but: was <false>

at plugins.DashboardViewPluginTest.latestsBuildsPortlet_correctJobAndBuild(DashboardViewPluginTest.java:175)
```

daniel-beck: Right, but that should be 2.359. Why 2.363? (I was looking at https://github.com/jenkinsci/acceptance-test-harness/pull/869/commits/fa5f58289a133eab18dee981d97f1dfa5df3d39c when I commented)

timja: see https://github.com/jenkinsci/acceptance-test-harness/runs/7824521017

it requires something from the open PR. likely the links changing in the table to include manage

PR #882, opened 2022-08-08
Test master build

NotMyFault:


timja:
it's failing because of git plugin, the default has now changed to verifying without accept-new which will fail here.

Can be seen in https://github.com/jenkinsci/acceptance-test-harness/pull/869

NotMyFault:
> it's failing because of git plugin, the default has now changed to verifying without accept-new which will fail here.

Oic, I was gonna recheck whether tests are actually impacted, as suspected on my issue risen.

timja:
can prob close this I think :)

Issue #877, opened 2022-08-02
SECURITY-1468's solution impacts ATH builds


NotMyFault:
After investigating into the failure of https://github.com/jenkinsci/acceptance-test-harness/pull/876, ATH is impacted by the security fix in the git-client-plugin too, https://github.com/jenkinsci/git-client-plugin/pull/875, because ATH runs on CentOS 7 and doesn't know how to deal with "accept new" as strategy.

Later reading over [JENKINS-69149](https://issues.jenkins.io/browse/JENKINS-69149), am I correct assuming that https://github.com/jenkinsci/git-client-plugin/pull/882 restores the functionality of ATH, once merged and released?

dwnusbaum:
> am I correct assuming that https://github.com/jenkinsci/git-client-plugin/pull/882 restores the functionality of ATH, once merged and released?

Unless `Main.isUnitTest` is set when running the ATH (I don't think it is), then if the ATH is built on ephemeral agents with a blank `known_hosts` file, or if it connects to dynamically created Git servers where the SSH keys change every time the test runs, then no, it will still need to be adapted.

From a quick look, I think that either `GitPluginTest` and similar tests will need to either populate `known_hosts` on the agent with the host key information from the container running the Git server, or they will need to [configure the controller to use the "Manually Provided Keys" host key verifications strategy](https://plugins.jenkins.io/git-client/#plugin-content-ssh-host-key-verification) and provide the keys for the Git server there, or maybe configure the "No verification" host key verification strategy if we know the ATH only connects to local Git servers and we do not care about MitM attacks.

CC @Pldi23 in case my comment is not quite right

timja:
Or we can move to recent Ubuntu with accept-new

NotMyFault:
> Or we can move to recent Ubuntu with accept-new

I started with that in https://github.com/jenkinsci/acceptance-test-harness/pull/878 earlier, needs some fixup in terms of Java directories to get it to work 👀

dduportal:
Dropping a note: we discussed this issue in today's infra meeting because the weekly release 2.363 was also bitten by this behavior on ci.jenkins.io.

Expect a fix in ci.jenkins.io wednesday 10th of August to apply the config as code change.

timja:
It’s an ATH framework change though, we’re trying to upgrade to a more maintained OS but tests are failing on it :(

dduportal:
[...]

PR #878, opened 2022-08-02
Switch to ubuntu 22.04

NotMyFault:
Closes https://github.com/jenkinsci/acceptance-test-harness/pull/803

PR #886, opened 2022-08-11
Use "No verification" ssh host key verification strategy

Pldi23:
Proposed solution for https://github.com/jenkinsci/acceptance-test-harness/issues/877

Git-client 3.11.2 and later uses ‘Known hosts file’ as default strategy (instead of git-client 3.11.1 used ‘Accept first connection'), so tests will continue to fail, I suggest to use ’No verification’ strategy for now (but please confirm that ATH only connects to local Git servers and we do not care about MitM attacks), but once #878 will be merged better to switch to ‘Accept first connection’.

Fixes https://github.com/jenkinsci/acceptance-test-harness/issues/877


NotMyFault (approved):
🎉

PR #889, opened 2022-08-12
More javadoc fixes

NotMyFault:
Follow up to https://github.com/jenkinsci/acceptance-test-harness/pull/888.

Apparently, maven shows 100 errors only, but this should really resolve all of them.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/junit/RuleAnnotation.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/junit/RuleAnnotation.java
@@ -30,12 +30,12 @@ public @interface RuleAnnotation {
 
     /**
      * Optional ordering among rules.
-     *
-     * Annotation with <code>priority >= 0</code> are guaranteed to be run after
+     * <p>
+     * Annotation with {@code priority >= 0} are guaranteed to be run after
      * Jenkins is up. Negative priorities are run before startup on best effort
      * basis. (It might not happen before for ExistingJenkinsController,
      * PooledJenkinsController and possibly others).
-     *
+     * <p>
      * Annotations that skips execution are encouraged to run before Jenkins is
      * booted up to save time. Note, that these implementations can not inject
      * Jenkins for obvious reasons.
--- a/src/main/java/org/jenkinsci/test/acceptance/junit/Wait.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/junit/Wait.java
@@ -103,7 +103,7 @@ public class Wait<Subject> extends FluentWait<Subject> {
      * Create wait with configurable timer.
      * <p>
      * This is useful for timeout waiting wall-clock time to pass.
-     * @see Wait<Subject>
+     * @see #Wait(Subject, ElasticTime)
      */
     public Wait(Subject input) {
         super(input);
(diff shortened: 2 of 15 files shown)


timja (approved):

src/main/java/org/jenkinsci/test/acceptance/junit/RuleAnnotation.java
```diff
@@ -30,12 +30,12 @@
...
      * Optional ordering among rules.
-     *
-     * Annotation with <code>priority >= 0</code> are guaranteed to be run after
+     * <p>
+     * Annotation with {@code priority >= 0} are guaranteed to be run after
      * Jenkins is up. Negative priorities are run before startup on best effort
      * basis. (It might not happen before for ExistingJenkinsController,
      * PooledJenkinsController and possibly others).
-     *
+     * <p>
```
jtnord, line 38: technically I think this is incorrect and should be `</p><p>`   but if the tool is happy then I think that is what we are aiming for.

src/main/java/org/jenkinsci/test/acceptance/junit/Wait.java
```diff
@@ -101,9 +101,9 @@ public Wait(Subject input, ElasticTime time) {
 
     /**
      * Create wait with configurable timer.
-     *
+     * <p>
      * This is useful for timeout waiting wall-clock time to pass.
-     * @see {@link Wait<Subject>(Subject, ElasticTime)}
+     * @see Wait<Subject>
```
jtnord, line 106: this seems like an incorrect change - it is pointing to the class doc and not the constructor?

if I have got the syntax right it should be (but this was done without IDE assistance)
```suggestion
     * @see #Wait(Subject, ElasticTime)
```

src/main/java/org/jenkinsci/test/acceptance/po/Trigger.java
```diff
@@ -2,8 +2,7 @@
 
 /**
  * Base type for {@link PageAreaImpl} for trigger.
- *
- * @see {@link Job#addTrigger(Class)}.
+ * See {@link Job#addTrigger(Class)}.
```
jtnord, line 5: Why is this not using @see as that is the purpose of it?

```suggestion
 * @see Job#addTrigger(Class)
```

jtnord (commented):
thanks for these fixes.  a couple of the `see` changes appear wrong to me.

OT: we do not publish the javadoc of the ATH - are these fixes to enable us to publish it?

src/main/java/org/jenkinsci/test/acceptance/junit/RuleAnnotation.java
NotMyFault: With javadoc 11 onwarads and its output using HTML 5 by default, we can use the appropriate shorthand to indicate a new paragraph. You don't necessarily need enclosing paragraph tags.

jtnord:
> > OT: we do not publish the javadoc of the ATH - are these fixes to enable us to publish it?
> 
> Java 11 is stricter with JavaDoc, and it's failing to release
> 
> https://github.com/jenkinsci/acceptance-test-harness/runs/7804549034?check_suite_focus=true
> 
> https://github.com/jenkins-infra/github-reusable-workflows/blob/3ac7ddc91af68ef1db6ea4bf1c695f2c06a46dd4/.github/workflows/maven-cd.yml#L56-L60

I released 1.118 with java11 yesterday - so I guess this is a consequence of enabling CD?

jtnord (approved):

NotMyFault:
> OT: we do not publish the javadoc of the ATH - are these fixes to enable us to publish it?

For CD, yes. It looks like it builds a javadoc jar now, but I do not recall getting one locally, when building the ATH.
Typically, I just have a sources jar alongside the normal build 🤔

PR #890, opened 2022-08-12
Use proper `@see` javadoc notation

NotMyFault:
Even if James' proposed changes were valid according to IJ, they didn't suffice for javadoc ☹️ 

Screenshot 2022-08-12 at 17 02 28

I changed your proposal to a Class#method reference to be explicit.

(On a future note, these issues should probably fail the build, because addressing them post merge isn't the best imo)

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/junit/Wait.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/junit/Wait.java
@@ -103,7 +103,7 @@ public class Wait<Subject> extends FluentWait<Subject> {
      * Create wait with configurable timer.
      * <p>
      * This is useful for timeout waiting wall-clock time to pass.
-     * @see #Wait(Subject, ElasticTime)
+     * @see Wait#Wait(Subject, ElasticTime)
      */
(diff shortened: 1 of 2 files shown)


jtnord (approved):

jtnord:
> (On a future note, these issues should probably fail the build, because addressing them post merge isn't the best imo)

👍  we need an invocation of `package -DskipTests` somewhere I guess...

timja (approved):

NotMyFault:
The release fails and complaints about the format of `@see Wait#Wait(Subject, ElasticTime)` still.

I took a further look and suspect a possible issue with javadoc in Java 11.

If I build the ATH with Temurin 11.0.16, the build fails with the error we can see on the build log, but if I build the ATH with Temurin 17.0.4, the build passes and and the javadoc reference is warranted in the .html file.
Maybe there's something up in Java 11 with parsing the generic properly? I didn't find something on JBS about that.

I'd recommend we temporarily move from `@see` to `@see Wait#Wait(Object, ElasticTime)` or something to unblock the CD release.

PR #891, opened 2022-08-12
Don't skip doclint

NotMyFault:
The build as-is fails for me locally if I invoke `mvn clean verify -DskipTests`, and should fail on ci.j too, because we're lacking https://github.com/jenkinsci/acceptance-test-harness/pull/890.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,6 +6,7 @@
     <groupId>org.jenkins-ci</groupId>
     <artifactId>jenkins</artifactId>
     <version>1.85</version>
+    <relativePath />
   </parent>
 
   <artifactId>acceptance-test-harness</artifactId>
@@ -485,6 +486,8 @@
         <artifactId>maven-source-plugin</artifactId>
         <executions>
           <execution>
+            <phase>compile</phase>
+            <id>generate-sources</id>
             <goals>
               <goal>jar</goal>
             </goals>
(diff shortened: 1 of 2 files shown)


timja (approved):

pom.xml
```diff
@@ -485,6 +486,8 @@
         <artifactId>maven-source-plugin</artifactId>
         <executions>
           <execution>
+            <phase>compile</phase>
```
jtnord, line 489: Why compile, this will be run on everysplit ATH.

Similar comment to below and the javadoc

pom.xml
```diff
@@ -495,15 +498,13 @@
...
-          <additionalparam>-Xdoclint:none</additionalparam>
-        </configuration>
         <executions>
           <execution>
-            <id>attach-javadocs</id>
-            <configuration>
-              <additionalparam>-Xdoclint:none</additionalparam>
-            </configuration>
+            <phase>compile</phase>
+            <id>generate-javadocs</id>
```
jtnord, line 504: This now IIUC means we generate javadocs twice during the build pipeline,
Once with the parent invocation bound to package and now here?

This also means we create the javadoc on each ATH split.  In the grand scheme of the ATH times this is probably lost but would we just be better off with a `mvn verify -DskipTests` in a branch on the pipeline?

jtnord (commented):
Works, but would seem to be inefficient?

pom.xml
timja: see also https://github.com/jenkinsci/acceptance-test-harness/pull/893

timja: `mvn verify -DskipTests` doesn't run javadoc currently by the looks of it

timja: https://github.com/jenkinsci/acceptance-test-harness/pull/893 works now

timja:
Is this needed after https://github.com/jenkinsci/acceptance-test-harness/pull/893?

NotMyFault:
> Is this needed after #893?

Given, the new task runs separately to the maven workflow to verify results independantly.

PR #893, opened 2022-08-13
Add CI stage

timja:
https://github.com/jenkinsci/acceptance-test-harness/pull/891#discussion_r945121846

Failed correctly in https://ci.jenkins.io/blue/organizations/jenkins/Core%2Facceptance-test-harness/detail/PR-893/5/pipeline/31

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,6 +24,16 @@ if (needSplittingFromWorkspace) {
 } else {
   splits = splitTests count(10)
 }
+
+branches['CI'] = {
+  stage('CI') {
+    node('maven-11') {
+      checkout scm 
+      sh "mvn verify --no-transfer-progress -DskipTests -P jenkins-release"
+    }
+  }
+}
+
 for (int i = 0; i < splits.size(); i++) {
   int index = i
   for (int j in [11]) {


NotMyFault (approved):

jtnord (approved):

Jenkinsfile
```diff
@@ -24,6 +24,16 @@ if (needSplittingFromWorkspace) {
 } else {
   splits = splitTests count(10)
 }
+
+branches['CI'] = {
+  stage('CI') {
+    node('maven-11') {
+      checkout scm 
+      sh "mvn verify --no-transfer-progress -DskipTests -P jenkins-release"
```
jtnord, line 32: I know this does run the javadoc  - but should this prefer to use the incrementals profile give we are using CD (and hence use the incremental profile)?

https://github.com/jenkinsci/pom/blob/jenkins-1.85/pom.xml#L931-L971?

most likely something for a different ticket - I'm not entirely sure how incremental publishing works from custom pipelines.

jtnord (approved):

Jenkinsfile
timja: I think this is fine although that could work too

PR #898, opened 2022-08-17
Test 2.361.1 RC

NotMyFault:
core.InstallWizardTest.wizardInstallSuggestedTest fails: https://ci.jenkins.io/job/Core/job/acceptance-test-harness/view/change-requests/job/PR-898/lastCompletedBuild/testReport/core/InstallWizardTest/java_11_jenkins_latest_split4___wizardInstallSuggestedTest/

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -37,7 +37,7 @@ branches['CI'] = {
 for (int i = 0; i < splits.size(); i++) {
   int index = i
   for (int j in [11]) {
-    for (String v in ['lts', 'latest']) {
+    for (String v in ['latest']) {
       int javaVersion = j
       String jenkinsUnderTest = v
       def name = "java-${javaVersion}-jenkins-${jenkinsUnderTest}-split${index}"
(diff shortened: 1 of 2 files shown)


NotMyFault:
I sense a possible correlation of the test failure with [JENKINS-69376](https://issues.jenkins.io/browse/JENKINS-69376), because one or more of the managed plugins for 2.361.1 requires a reboot after the initial installation, therefore a different tab is displayed during the initial installation, but the ATH expects to find `.btn-primary.install-done` to finish the installation, which doesn't exist on the reboot view.

jtnord:
> I sense a possible correlation of the test failure with [JENKINS-69376](https://issues.jenkins.io/browse/JENKINS-69376).
> 
> One or more of the managed plugins for 2.361.1 require a reboot after the initial installation (which one and why would be nice to know, because 2.361 as weekly does not ship any plugins requiring a reboot), hence a different view is displayed, but the ATH expects to find `.btn-primary.install-done` to finish the installation, which doesn't exist on the reboot view.

This occurs if you install a few plugins and one requires a newer version of a "detached" plugin (but the detached plugin was installed earlier due to the install of the previous plugin)

jtnord:
from the logs -> master00000|2022-08-17 19:45:19.751+0000 [id=107]	INFO	hudson.ClassicPluginStrategy#updateDependency: Updated dependency of bouncycastle-api

it was loaded as a detached plugin and then asked to be installed.

PR #900, opened 2022-08-18
Address IOUtils' charset deprecations

NotMyFault:


src/main/java/org/jenkinsci/test/acceptance/SshKeyPair.java
```diff
@@ -42,19 +43,19 @@ public SshKeyPair(File publicKeyFile, File privateKeyFile) {
     }
 
     public String readPublicKey() throws IOException {
-        return FileUtils.readFileToString(publicKey);
+        return FileUtils.readFileToString(publicKey, StandardCharsets.UTF_8);
```
jtnord, line 46: just came from my PR where this has caused conflicts.

These are files that are executied by the system, so assuming UTF8 here looks wrong and should rather have been the default charset, or the default file enconding.

I use a charset that is mostly going to be with this - but not sure that is the case for everyone?

NotMyFault: >  but not sure that is the case for everyone?

Tests were passing for me locally, as they did on the CI. Did I block your gitrepo PR?

If this is a considerable issue, I can move them to the default charset.

jtnord: > Did I block your gitrepo PR?

not at all - I just got notified of the conflicts as I had updated the `GitRepo` to use `Charset.default()` as we where writing files to be read by the OS and there where `IOUtils` charset deprecations ;)

> If this is a considerable issue, I can move them to the default charset.

Certainly not a showstopper for me or CI, but more a word of caution that we should not be assuming UTF-8 for things that are for the OS.

That said I noticed https://www.ietf.org/rfc/rfc4716.txt  sec 3.3 which is interesting as it defines the charset for the headers, but not the actual key part (which is base64).  So in this case `UTF-8` is probably less incorrect that the platform default and doing something "correct" would be "fun" that most likely uneeded for this specific case.

(anyone running the ATH on `EBDIC`/`JIS_X_0208` or `UTF-16` encoded systems will probably have a fun time anyway)

PR #905, opened 2022-08-22
chore(ATH Docker Image) build and deploy with infra.ci.jenkins.io

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/3084

This PR adds a 2nd pipeline which goal is to build/test/deploy the ATH Docker image on the Jenkins controller infra.ci.jenkins.io.

- ~[ ] Link to relevant pull requests, esp. upstream and downstream changes~

### DIFF
--- a/src/main/resources/ath-container/Dockerfile
+++ b/src/main/resources/ath-container/Dockerfile
@@ -3,7 +3,6 @@
 # see docs/DOCKER.md for usage
 
 FROM centos:7
-LABEL maintainer="ogondza@gmail.com"
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
 ## Always install the latest version of these packages.
--- a/src/main/resources/ath-container/Dockerfile
+++ b/src/main/resources/ath-container/Dockerfile
@@ -3,8 +3,11 @@
 # see docs/DOCKER.md for usage
 
 FROM centos:7
-MAINTAINER ogondza@gmail.com
+LABEL maintainer="ogondza@gmail.com"
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
+## Always install the latest version of these packages.
+# hadolint ignore=DL3033
 RUN yum -y install \
         curl \
         docker \
@@ -41,7 +44,7 @@ ENV gid $gid
 EXPOSE 5942
 
 RUN groupadd ath-user -g $gid && \
-    useradd ath-user -u $uid -g $gid -m -d /home/ath-user && \
+    useradd ath-user -l -u $uid -g $gid -m -d /home/ath-user && \
     # Give permission to modify the alternatives links to change the java version in use
     chown ath-user:ath-user /etc/alternatives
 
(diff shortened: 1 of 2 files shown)


timja:
The checks are a bit confusing in that there's soo many for the dockerfile, and then the overall build status is `continuous-integration/jenkins/pr-merge` which is for the docker build and not the main build.

Would it be possible to turn off the stage builds and configure the overall build status to be `Docker build` ?

src/main/resources/ath-container/Dockerfile
```diff
@@ -3,8 +3,11 @@
 # see docs/DOCKER.md for usage
 
 FROM centos:7
-MAINTAINER ogondza@gmail.com
+LABEL maintainer="ogondza@gmail.com"
```
timja, line 6:
```suggestion
```

dduportal:
> The checks are a bit confusing in that there's soo many for the dockerfile, and then the overall build status is `continuous-integration/jenkins/pr-merge` which is for the docker build and not the main build.
> 
> Would it be possible to turn off the stage builds and configure the overall build status to be `Docker build` ?

Yes of course, nice suggestion!

Thanks for the review.

This PR stays in draft, because the pipeline-library needs to be updated to allow specifying the registry (and adding the correct credentials) because `jenkinsciinfra` is used by default and there are no way to override it today. Should be easy.


I'll rebase this PR and set it as "ready to review" once done.

dduportal:
> The checks are a bit confusing in that there's soo many for the dockerfile, and then the overall build status is `continuous-integration/jenkins/pr-merge` which is for the docker build and not the main build.
> 
> Would it be possible to turn off the stage builds and configure the overall build status to be `Docker build` ?

@timja I can't easily find a way to disable the status check "per stage". I can find how to publish a custom one from pipeline though.

I'me playing around with the mulitbranch job configuration (manual changes for now) on infra.ci and just set it up like this:

Capture d’écran 2022-08-23 à 15 33 18

But it seems that the checks are kept over time. Do you have a way to clean them up or should I close this PR and open a new one with this setup?

timja:
if you push a new commit it should get rid of the old ones.

The stage one is probably ok, as long as the 'default' looking link doesn't point to infra.ci which most people can't access

dduportal:
> if you push a new commit it should get rid of the old ones.

Alas it does not :( I've rebased and updated the branch: still the same checks. It's not clear how to disable it.
Even renaming the check did not had any effect, I'm not sure if I did something wrong or what.

> 'default' looking link

Which one exactly (screenshot to clarify)?

dduportal:
> point to infra.ci which most people can't access

I'm confused between the "status checks" on the "Conversation" tabs and the "Checks" on the "Checks" tabs.
It seems that my changes of the job configuration are impacting the "Checks" tabs, which I would like to enable again so that users would be able to get the lint/build logs.

But I don't understand how to tune the "status checks" though 🤔

timja (approved):

PR #913, opened 2022-09-06
click 'check Now' when MockUpdateCenter is running 

Pldi23:
Fix update center usage in ATH

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
@@ -95,31 +95,34 @@ public class PluginManager extends ContainerPageObject {
      * Force update the plugin update center metadata.
      */
     public void checkForUpdates() {
-        mockUpdateCenter.ensureRunning(jenkins);
-        visit("index");
-        final String current = getCurrentUrl();
-        // The check now button is a form submit (POST) with a redirect to the same page only if the check is successful.
-        // We use the button itself to detect when the page has changed, which happens after the refresh has been done
-        // And we check for the presence of the button again
-        WebElement checkButton = find(by.link("Check now"));
-        checkButton.click();
-        // The wait criteria is: we have left the current page and returned to the same one
-        waitFor(checkButton).withTimeout(java.time.Duration.of(time.seconds(30), ChronoUnit.MILLIS)).until(webElement -> {
-            try {
-                // We interact with the element just to detect if it is stale
-                webElement.findElement(by.id("it does not matter"));
-            } catch(StaleElementReferenceException e) {
-                // with this exception we know we've left the original page
-                // we look for an element in the page to check for success
-                if (current.equals(getCurrentUrl())) {
-                    return true;
+        boolean isRunning = mockUpdateCenter.ensureRunning(jenkins);
+        if (isRunning) {
+            visit("index");
+            final String current = getCurrentUrl();
+            // The check now button is a form submit (POST) with a redirect to the same page only if the check is successful.
+            // We use the button itself to detect when the page has changed, which happens after the refresh has been done
+            // And we check for the presence of the button again
+            WebElement checkButton = find(by.link("Check now"));
+            checkButton.click();
+            // The wait criteria is: we have left the current page and returned to the same one
+            waitFor(checkButton).withTimeout(java.time.Duration.of(time.seconds(30), ChronoUnit.MILLIS)).until(webElement -> {
+                try {
+                    // We interact with the element just to detect if it is stale
+                    webElement.findElement(by.id("it does not matter"));
+                } catch(StaleElementReferenceException e) {
+                    // with this exception we know we've left the original page
+                    // we look for an element in the page to check for success
+                    if (current.equals(getCurrentUrl())) {
+                        return true;
+                    }
+                } catch(NoSuchElementException e) {
+                    return false;
                 }
-            } catch(NoSuchElementException e) {
                 return false;
-            }
-            return false;
-        });
-        updated = true;
+            });
+            updated = true;
+
+        }
     }
 
     public enum InstallationStatus {NOT_INSTALLED, OUTDATED, UP_TO_DATE}
--- a/src/main/java/org/jenkinsci/test/acceptance/update_center/MockUpdateCenter.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/update_center/MockUpdateCenter.java
@@ -84,20 +84,21 @@ public class MockUpdateCenter implements AutoCleaned {
 
     private HttpServer server;
 
-    public void ensureRunning(Jenkins jenkins) {
+    public boolean ensureRunning(Jenkins jenkins) {
         if (original != null) {
-            return;
+            return false;
         }
         JsonNode ucNode = new UpdateCenter(jenkins).getJson("tree=sites[url,id]");
         List<String> sites = ucNode.findValuesAsText("url");
         List<String> ids = ucNode.findValuesAsText("id");
+        LOGGER.log(Level.INFO, "update sites: {0}", sites);
         if (sites.size() != 1) {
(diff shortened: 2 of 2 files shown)


timja:
can you explain a bit more what this is fixing?

Pldi23:
@jtnord @jmdesprez If I remember everything correctly this is the solution we discussed last week, please review and correct me if I'm wrong.

During investigating I found few weird things and It's not clear to me if is it ok or not:
1. After our meeting I thought that we don't want to click 'Check Now' if MockUpdateCenter did not start, but with that change local server starts fine even if I ran test without internet connection (so we are still clicking 'Check Now' regardless of whether there is an internet connection or not). Is it a mistake in my expectations here?

2. When i'm running ATH test locally, but wifi is switch off, I expect MockUpdateCenter to work and list of available plugins should not be empty, but it is
Screenshot 2022-09-06 at 19 02 52  
Am I missed something here too?


3. Analysing failure logs it seems that  every ATH test annotated `WithPlugin` do not use MockUpdateCenter at all, every test log contain line 
`
Jul 20, 2022 9:28:10 AM org.jenkinsci.test.acceptance.update_center.MockUpdateCenter ensureRunning WARNING: found an unexpected number of update sites: [file:/root/ath/target/jenkins4850299043136569497home/war/WEB-INF/plugins/update-center.json, https://jenkins-updates.cloudbees.com/update-center/envelope-core-cm/update-center.json]` 
comes from https://github.com/jenkinsci/acceptance-test-harness/pull/913/files#diff-cfb2fe3a56bdb573bba0e66a1af029d251f2068c3b2e95f380c3eff8f9c59ec6R97 . It means that `ensureRunning ` returns and server does not start. Is it expected? (It also means that locally I'm running different ATH setup and can't fully reproduce the issue, could I somehow setup the same set of update sites locally (some property file or variable somewhere)?)

Pldi23:
> can you explain a bit more what this is fixing?

@timja I added description and a comment. For the context we have a meeting last week within FlakeBusters Team and James discussing that all ATH tests are flaky because everything using Update Center is flaky, and we share some thoughts how we could reduce flakiness when using `@WithPlugins`. This is a draft PR to discuss opened questions in details. Feel free to share your thoughts with us :)

src/main/java/org/jenkinsci/test/acceptance/update_center/MockUpdateCenter.java
```diff
@@ -84,20 +84,21 @@ public class MockUpdateCenter implements AutoCleaned {
...
-    public void ensureRunning(Jenkins jenkins) {
+    public boolean ensureRunning(Jenkins jenkins) {
         if (original != null) {
-            return;
+            return false;
         }
         JsonNode ucNode = new UpdateCenter(jenkins).getJson("tree=sites[url,id]");
         List<String> sites = ucNode.findValuesAsText("url");
         List<String> ids = ucNode.findValuesAsText("id");
+        LOGGER.log(Level.INFO, "update sites: {0}", sites);
```
Pldi23, line 94: I'll remove it later, just wanted to execute test in CB test selector to see that we are using two sites that prevent MockUpdateCenter to start

jtnord:
> 3. Analysing failure logs it seems that  every ATH test annotated `WithPlugin` do not use MockUpdateCenter at all, every test log contain line

because you are using a CloudBees' jenkins build which has multiple update centers and does not use the mock update center but its own ones.

In order to test the code change you will need to run this against *both* an open source jenkins build (which only has the one update site and will not show that line) and CloudBees' site.

src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
```diff
@@ -95,31 +95,34 @@ public PluginManager(Jenkins jenkins) {
...
-        waitFor(checkButton).withTimeout(java.time.Duration.of(time.seconds(30), ChronoUnit.MILLIS)).until(webElement -> {
-            try {
-                // We interact with the element just to detect if it is stale
-                webElement.findElement(by.id("it does not matter"));
-            } catch(StaleElementReferenceException e) {
-                // with this exception we know we've left the original page
-                // we look for an element in the page to check for success
-                if (current.equals(getCurrentUrl())) {
-                    return true;
+        boolean isRunning = mockUpdateCenter.ensureRunning(jenkins);
```
jtnord, line 98: `isRunning` is a misleading variable name.  the method will return `false` even when the `mockUpdateCenter` is correctly injected and running.  It only returns `true` the first time the mockUpdateCenter is injected.

The code is kind of right - but I think we need to maybe make `checkForUpdates` always check for updates, but possible all calling code should be changed to not call this before attempting to install plugins - and that the constructor is this should maybe not call it but call say another method.

jtnord:
> can you explain a bit more what this is fixing?

Jenkins (OSS) at startup refreshes the `update sites`.  This is asynchronous so may or may not have happened, but when testing OSS Jenkins the `mockUpdateCetner` will have been actived so refreshing the metadata is a good thing here.

In CloudBees products there is more than 1 update site so the mock update center is never used.
Additionally bundles plugins that we support inside the war for offline users, so that no network interaction is required and has an UpdateSite to reflect this.
This updateSite is refreshed as part of the product start up (in an `Initializer` so this updateSites data is always up to date).
However there is another online updateSite to provide plugins that we do no support.

When clicking refreshnow - the tests become falky as this online updatecenter is well online, so network issues or server issues can make `checkNow` fail due to a network or server error.  But as we do not support the plugins from this UpdateSite we do not care if the data it servers is out of date or not for the tests we run, as all the tests will use plugins that will come from the offline updateSite. 

So by removing the forced check, we can make the tests much much less flaky.

The forced check should still be used for the MockUpdateCenter, but where and when this is used can be enhanced.

src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
Pldi23: What are your thoughts if we split `checkForUpdates` into two parts, and call `ensureRunning` where always, but `clickCheckNow` before plugin installation if list of plugins was not refreshed before? see [cf72035](https://github.com/jenkinsci/acceptance-test-harness/pull/913/commits/cf72035f7920c813736e64bdf805dc92a74a699d)

Pldi23:
> > 3. Analysing failure logs it seems that  every ATH test annotated `WithPlugin` do not use MockUpdateCenter at all, every test log contain line
> 
> because you are using a CloudBees' jenkins build which has multiple update centers and does not use the mock update center but its own ones.
> 
> In order to test the code change you will need to run this against _both_ an open source jenkins build (which only has the one update site and will not show that line) and CloudBees' site.

I ran this changes locally against jenkins build, seems that changes does not brake nothing, but when I run without internet connection the test is failing in [tickPluginToInstall](https://github.com/jenkinsci/acceptance-test-harness/pull/913/files#diff-f76753d582dcf82b2c291976c6a97ae52bacf98c334428c8702a68495797822fR263) because list of available plugins is empty even after refresh. I also used test selector to run tests with internet connection against Cloudbees build (by the way how could I ran it against cloudbees locally?) and it also does not brake nothing.

Pldi23:
@jtnord , @jmdesprez Ok, I completed testing for both `jenkins war` and `cloudbees war`. With this change:

- on jenkins war ATH works as before, no regression found.

- on cloudbees war we most likely will never click ‘Check now’, because during my testing on different `cloudbees wars`, with and without connection the list of all available plugins are full and we don’t even need to refresh plugins. 

But when internet connection is off and `WithPlugins` need to install plugin which is not in the local offline update center the test will fail with [tickPluginToInstall_stacktrace.txt](https://github.com/jenkinsci/acceptance-test-harness/files/9498821/tickPluginToInstall_stacktrace.txt). 
What would be our next steps? do we have any initial idea how to deal with that types of plugins? (do we want additional UC for them or include somehow into a offline UC, or something other?)

I also would suggest to merge this  fix to reduce flakiness with ‘Check now’ while we thinking about `tickPluginToInstall` issue, what are your thoughts?

jtnord:
replaced by https://github.com/jenkinsci/acceptance-test-harness/pull/1185

PR #931, opened 2022-09-20
Replace slf4j-log4j12 with slf4j-jdk14

NotMyFault:


pom.xml
```diff
@@ -288,7 +288,7 @@
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-log4j12</artifactId>
+      <artifactId>slf4j-reload4j</artifactId>
```
timja, line 291: Would this work instead? not sure if it's being used as a bridge or as an implementation
```suggestion
      <artifactId>slf4j-jdk14</artifactId>
```

NotMyFault: According to the SLF4J manual, the preferred drop-in replacement is reload4j.

timja: but do we want this to stay on log4j 1?

NotMyFault: Quickly checked locally, jdk14 passes `verify`.

timja (approved):

PR #942, opened 2022-09-21
Add Java 17 to CI workflow

NotMyFault:


NotMyFault:
There's no jdk 17 openjdk devel package for rhel 7, therefore I'll wait until we switched to ubuntu, before continueing here.

PR #949, opened 2022-09-27
Update htmlunit-driver from 2.33.0 to 2.64.0

NotMyFault:
The selenium GAV is banned for any dependency updates by dependabot, but 2.64.0 of the htmlunit driver is the last version labeled as compatible with our Selenium version and should be safe to use.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,7 +34,6 @@
     <selenium.version>3.141.59</selenium.version>
     <guava.version>31.1-jre</guava.version> <!-- aligned with selenium -->
     <aether.version>1.1.0</aether.version>
-    <jackson.bom.version>2.12.1</jackson.bom.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -34,7 +34,6 @@
     <selenium.version>3.141.59</selenium.version>
     <guava.version>31.1-jre</guava.version> <!-- aligned with selenium -->
     <aether.version>1.1.0</aether.version>
-    <jackson.bom.version>2.12.1</jackson.bom.version>
```
NotMyFault, line 37: Unused, hence removed.

timja:
do we even use htmlunit here?

NotMyFault:
> do we even use htmlunit here?

In a couple of classes.

Issue #736, opened 2022-01-09
Use a fixed version of core in tests and automate updates to it


timja:
### What feature do you want to see added?

I would like to get a PR created by updatecli / whatever that bumps the version of core that ATH tests against, currently it just always runs against the last weekly but that means we don't recognise regressions when they happen and there's no notification to check the test suite.

In bom we do:

https://github.com/jenkinsci/bom/blob/master/.github/workflows/updatecli.yaml
https://github.com/jenkinsci/bom/blob/master/updatecli/updatecli.d/weekly-apply.sh

Currently the version is set in https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/main/resources/ath-container/run.sh#L48 where it just uses latest and 

I think what we want to do is change the [Jenkinsfile](https://github.com/jenkinsci/acceptance-test-harness/blob/master/Jenkinsfile#L55) to read a file with the version number we're testing against.

The reason to use a file is to allow 'untrusted' contributors to test against different versions without needing write access

PR #961, opened 2022-10-02
Run ATH on a weekly basis

NotMyFault:
The change proposed runs the ATH after the weekly release, to highlight possible regressions quicker, and outside regularly core PR testing.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,6 +3,13 @@
 
 properties([disableConcurrentBuilds(abortPrevious: true)])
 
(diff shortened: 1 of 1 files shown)


timja (approved):
I guess okay but doesn't mean people will check it, this would be a better solution I think:

https://github.com/jenkinsci/acceptance-test-harness/issues/736

NotMyFault:
> this would be a better solution I think:

Yeah, checked in before. But, I guess, this one should suffice for now.

PR #968, opened 2022-10-06
Setup Jenkins CodeQL

NotMyFault:
Let's see if we can use this here too...

### DIFF
--- a/.github/workflows/jenkins-security-scan.yml
+++ b/.github/workflows/jenkins-security-scan.yml
@@ -1,0 +1,20 @@
+name: Jenkins Security Scan
+on:
+  push:
+    branches:
+      - "master"
+  pull_request:
+    types: [ opened, synchronize, reopened ]
+  workflow_dispatch:
+
+permissions:
+  security-events: write
+  contents: read
+  actions: read
+
+jobs:
+  security-scan:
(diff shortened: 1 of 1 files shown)


.github/workflows/jenkins-security-scan.yml
```diff
@@ -0,0 +1,20 @@
...
+permissions:
+  security-events: write
+  contents: read
+  actions: read
+
+jobs:
+  security-scan:
+    uses: jenkins-infra/jenkins-security-scan/.github/workflows/jenkins-security-scan.yaml@v2
+    with:
+      java-cache: '' # Optionally enable use of a build dependency cache. Specify 'maven' or 'gradle' as appropriate.
```
timja, line 19:
```suggestion
      java-cache: 'maven' # Optionally enable use of a build dependency cache. Specify 'maven' or 'gradle' as appropriate.
```

timja (approved):

timja:
not sure if its terribly useful adding this to a testing framework...

jtnord:
> not sure if its terribly useful adding this to a testing framework...

Looks like one thing for the pooled Jenkins controller where it has a callable.  Otherwise looks like noise (and if the pooled callable works which @uhafner uses iirc then even that finding could be noise)

PR #1003, opened 2022-12-02
Removing unused and broken `JenkinsConfig.addCloud`

jglick:
Introduced in 6f6f371a84130888ead74f50cda778c4873de87a apparently with no use case, not currently used, and anyway broken as of https://github.com/jenkinsci/jenkins/pull/4339.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/JenkinsConfig.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/JenkinsConfig.java
@@ -1,9 +1,6 @@
 package org.jenkinsci.test.acceptance.po;
 
-import org.openqa.selenium.WebElement;
-
 import java.net.URL;
-import java.util.List;
 
 /**
  * Page object for the system configuration page.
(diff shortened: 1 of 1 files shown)


Vlatombe (approved):

jglick:
For a proprietary test I did manage to write the equivalent. In case anyone ever actually needs to write such a test in this repo, try something like

```java
public final class CloudConfiguration extends ConfigurablePageObject {
    public final Control addCloudButton = control("/hetero-list-add[cloud]");
    public CloudConfiguration(Jenkins jenkins) {
        super(jenkins, jenkins.url("configureClouds/"));
    }
    @Override
    public URL getConfigUrl() {
        return url;
    }
    public <T extends Cloud> T addCloud(Class<T> type) {
        ensureConfigPage();
        addCloudButton.selectDropdownMenu(type);
        List<WebElement> all = all(by.name("cloud"));
        WebElement last = all.get(all.size() - 1);
        return newInstance(type, this, last.getAttribute("path"));
    }
}
```

used like

```java
CloudConfiguration config = new CloudConfiguration(controller);
config.configure();
CloudType cloud = config.addCloud(CloudType.class);
// configure cloud
config.save();
```

PR #1028, opened 2023-01-31
Update SvnContainer from debian buster to bullseye

NotMyFault:


timja:
questionable if we even want to keep this anymore...

PR #973, opened 2022-10-14
Upgrade to Selenium 4

jtnord:
Based on #970 at least some tests work on my machine...

@Vlatombe 

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #1045, opened 2023-03-04
Adapt to Firefox 110 behaviour change

timja:


raul-arabaolaza:
I would say is better to get a fixed, known to work, version of firefox like 108 instead of trying to adapt here as we can not evaluate the impact in other browsers like chrome (where BTW everything works perfectly on the last version). Also https://github.com/jenkinsci/acceptance-test-harness/blob/200aafd9410b487aa184c5d92c639c3baacdc09d/src/main/resources/ath-container/Dockerfile#L30 needs to be changed as installing the last available version of firefox is clearly not deterministic and we can expect to face issues like this one continuously.

I am all in about update things to work properly with last browsers, but I would say the best place to do it is https://github.com/jenkinsci/acceptance-test-harness/pull/973 where we are also updating selenium. I am not feeling safe using an old selenium version with bleeding edge browser version

timja:
Sounds good it was just a quick POC to see what it looked like

timja:
Generally with browsers you can’t really install a fixed version. They remove them too quickly and we don’t want to get stuck on an old version, but in this case let’s get selenium upgraded first before looking at it

PR #1043, opened 2023-03-03
Use older 108 version of firefox in ath-container

raul-arabaolaza:
110 (current one available in ppa) seems to have some issues with Selenium making the existing code not work anymore

~So to validate is the firefox version I am using here firefox esr 108, if that solves problems I will rewrite to use the standard firefox instead, sadly 108 version is not on the ppa repository anymore, that is why for this tests I am using esr~

The last known good version was 108.0.2, which is now installed, and fixed,  in the docker image. Basically I copied and adapted the installation procedure from the selenium docker images.

- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #1048, opened 2023-03-06
Fix hadolint errors

NotMyFault:
https://github.com/jenkinsci/acceptance-test-harness/pull/1043 didn't run on infra.ci, therefore hadolint errors remained invisible.
This PR restores deployment if the build is green. Additionally, let's see if infra.ci picks up this PR.


src/main/resources/ath-container/Dockerfile
```diff
@@ -32,15 +32,16 @@ RUN apt-get update && \
     apt-get clean all && rm -rf /var/cache/apt
 
 # Install a fixed firefox version that is known to work with the current selenium version, copied from https://hub.docker.com/r/selenium/node-firefox/dockerfile
-ARG FIREFOX_VERSION=108.0.2
+ENV FIREFOX_VERSION 108.0.2
```
dduportal, line 35: nit: is this hadolint which proposed to switch from `ARG` to `ENV`?

I would have expected `ARG` to be used to avoid persisting the variable in the final image but I might have missed something here?

src/main/resources/ath-container/Dockerfile
```diff
@@ -32,15 +32,16 @@ RUN apt-get update && \
...
+# hadolint ignore=DL3008, DL3015
 RUN apt-get -y install libgtk-3-0 libasound2 libdbus-glib-1-2 libx11-xcb-dev libpci-dev libgl1-mesa-glx wget bzip2 \
   && rm -rf /var/lib/apt/lists/* /var/cache/apt/* \
-  && wget --no-verbose -O /tmp/firefox.tar.bz2 https://download-installer.cdn.mozilla.net/pub/firefox/releases/$FIREFOX_VERSION/linux-x86_64/en-US/firefox-$FIREFOX_VERSION.tar.bz2 \
+  && curl -sO https://download-installer.cdn.mozilla.net/pub/firefox/releases/${FIREFOX_VERSION}/linux-x86_64/en-US/firefox-${FIREFOX_VERSION}.tar.bz2 /tmp/firefox.tar.bz2 \
```
dduportal, line 39:
```suggestion
  && curl --silent https://download-installer.cdn.mozilla.net/pub/firefox/releases/${FIREFOX_VERSION}/linux-x86_64/en-US/firefox-${FIREFOX_VERSION}.tar.bz2 --output /tmp/firefox.tar.bz2 \
```

src/main/resources/ath-container/Dockerfile
NotMyFault: I've restored `ARG`, hadolint doesn't complain locally anymore. Maybe I falsely suppresed it here.

dduportal (approved):

src/main/resources/ath-container/Dockerfile
```diff
@@ -33,14 +33,15 @@ RUN apt-get update && \
 
 # Install a fixed firefox version that is known to work with the current selenium version, copied from https://hub.docker.com/r/selenium/node-firefox/dockerfile
 ARG FIREFOX_VERSION=108.0.2
+# hadolint ignore=DL3008, DL3015
```
dduportal, line 36:
```suggestion
# hadolint ignore=DL3008
```

DL3015 should not be ignored: it relies on a set of implicit additional dependencies (the "recommended" packages) that can change over time even on a fixed disribution.

Better to add the `--no-install-recommends` and add the required packages (as these packages... are needed).
Example is `ca-certificates` which is recommended by `curl` but should be mande explicit most of the time.

dduportal, line 36: About `DL3018`: do you mind adding a comment before the ignore rule to explain why you would want to ignore it? It helps to share knowledge of the "why".

NotMyFault: Do you mean 3008? We ignore it here the same way we ignore it above.

dduportal: my bad, I meant `DL3008ù` yes 👍 
No problem to not pin dependencies. But I really like adding a comment explaining "why" along with the inline ignore ;)

NotMyFault: for sure, added 👍🏻

src/main/resources/ath-container/Dockerfile
```diff
@@ -33,14 +33,16 @@ RUN apt-get update && \
 
 # Install a fixed firefox version that is known to work with the current selenium version, copied from https://hub.docker.com/r/selenium/node-firefox/dockerfile
 ARG FIREFOX_VERSION=108.0.2
-RUN apt-get -y install libgtk-3-0 libasound2 libdbus-glib-1-2 libx11-xcb-dev libpci-dev libgl1-mesa-glx wget bzip2 \
+# Package versions remain unpinned
```
dduportal, line 36:
```suggestion
# hadolint: Do not pin packages as they are requirements only for firefox installation (and not ofr the rest of the ATH) and the effort required to track each version would be too huge
```
WDYT?

- Prefix by hadolint for reader ot understand it's related to the below comment
- Explain the "why" instead of the "what" => are you ok with this "why" or is it another reason?

NotMyFault: Sounds good to me :)

src/main/resources/ath-container/Dockerfile
```diff
@@ -81,7 +83,11 @@ RUN echo " \
     " | tee /etc/apt/preferences.d/docker-ce
 
 # Install pinned version
-RUN  apt-get update --quiet && apt-get install --yes --no-install-recommends docker-ce docker-ce-cli docker-buildx-plugin
+# hadolint: Do not pin packages as they are requirements only for docker installation (and not of the rest of the ATH) and the effort required to track each version would be too huge
+# hadolint ignore=DL3008
+RUN apt-get update --quiet && apt-get install --yes --no-install-recommends docker-ce docker-ce-cli docker-buildx-plugin \
+    && apt-get clean \
+    && rm -rf /var/cache/apt
```
dduportal, line 90: The Docker CE version is pinned by the instruction above, which is a good thing (Docker actually have breaking changes some time to time: 22.x to 23.x did break a lot of things in a good way, but broke nevertheless).

But the "pinning" on APT is useful in a "long running instance" such as a VM. We can improve the current pinning by using pure-Docker form to make it easier to read and maintain.

My proposal is a refinment of the existing code and include removing the whole instruction above (`RUN echo "..." | tee /etc/apt/preferences.d/docker-ce` because I can't make a suggestion on unchanged code):

```suggestion
# Ensure that version is fixed
ARG DOCKER_VERSION=23.0.1
ARG DOCKER_BUILDX_VERSION=0.10.2
RUN apt-get update --quiet \
  && apt-get install --yes --no-install-recommends \
    docker-ce="5:${DOCKER_VERSION}*" \
    docker-ce-cli="5:${DOCKER_VERSION}*" \
    docker-buildx-plugin="${DOCKER_BUILDX_VERSION}*" \
  && apt-get clean \
  && rm -rf /var/cache/apt
```

=> no need to ignore package pinning (in this particular case, Docker-CE and Docker BuildX, pinning is important for the ATH usage otherwise it will break in 3 month with the next Docker CE update) in hadolint


WDYT?

NotMyFault: Sounds fine, I wasn't aware of that many breaking changes here 👀

dduportal: no problem, it was a good opportunity to point it out so the time you spent on fixing hadolint would be beneficial for you and the ATH

Issue #1053, opened 2023-03-08
Additional test coverage


jtnord:
### What feature do you want to see added?

CloudBees has maintained it's own extended set of tests for some open source plugins.

These internal tests don't benefit the community as much as they could as there is no visibility of them outside of CloudBees.

They are run on every build of our product (many times per day) without any surefire re-runs enabled and so are considered as stable as the ATHs in this repository.

We intend to open source these tests by migrating them to this repository where the tests do not significantly overlap.

This ticket is so you are aware of this upcoming work @timja @NotMyFault

uhafner:
I would recommend to move those tests to the corresponding plugins and not to the ATH. At least that was the consensus that we had quite a couple of years ago. It would only make sense to add more tests here that actually test core functionality that is not tested here yet. Otherwise we are bloating up the ATH again. @olivergondza introduced a successful concept for UI tests in plugins that should work here as well. I am also using this concept in all of my plugins very successfully.

timja:
cc @jtnord @rsandell @olamy @ampuscas ^^

uhafner:
> > I would recommend to move those tests to the corresponding plugins and not to the ATH. At least that was the consensus that we had quite a couple of years ago
> 
> not to my knowledge or recollection. 

It started with https://github.com/jenkinsci/acceptance-test-harness/pull/556

> In fact I would say the ATH tests for the warnings-ng are only in the warnings-ng repo is an anti-pattern as they are not picking up regressions caused by changes in Jenkins in a reasonably timely fashion (they are not run as part of the PCT).
> Having some UI tests in the plugin code is reasonable a it allows the testing for regressions in plugin code when developing the plugin - but there also needs to be some coverage in the ATH.

Yes, I think a single test per important plugin would be helpful. 

> 
> Additionally and this is a not the current state - the ATH tests here should really be exercising human flow now individual plugins - however that ship has sailed and is a larger effort that can be done in the future.
> 
> > It would only make sense to add more tests here that actually test core functionality that is not tested here yet.
> 
> core to what Jenkins or plugins?
> 
> if Jenkins then I disagree strongly.
> 

Why? There are so many areas in Jenkins core that are not UI tested. 

And for instance, the new Ant tests add new integration tests for Ant here in the ATH, these should be part of the Ant plugin, not ATH. (They even can be implemented as integration tests without UI layer).

uhafner:
[...]

PR #1064, opened 2023-03-14
[#1053] Add test for windows agents plugin

rsandell:
See https://github.com/jenkinsci/acceptance-test-harness/issues/1053
The Jenkinsfile doesn't seem to be running the ATH on windows agents, so this test is disabled for now.

- [X] Link to relevant issues in GitHub or Jira
- [X] Link to relevant pull requests, esp. upstream and downstream changes

### DIFF
--- a/src/test/java/plugins/WindowsAgentsTest.java
+++ b/src/test/java/plugins/WindowsAgentsTest.java
@@ -131,7 +131,7 @@ public class WindowsAgentsTest extends AbstractJUnitTest {
     private String getWindowsProperty(String property) {
         String p = System.getProperty(property);
         if (p == null || p.isEmpty()) {
-            throw new AssumptionViolatedException(String.format("No windows property (%s) found, ignoring windows slaves test", property));
+            throw new AssumptionViolatedException(String.format("No windows property (%s) found, ignoring windows test", property));
         }
         return p;
     }
--- a/src/test/java/plugins/WindowsAgentsTest.java
+++ b/src/test/java/plugins/WindowsAgentsTest.java
@@ -1,0 +1,146 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2023 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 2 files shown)


src/test/java/plugins/WindowsAgentsTest.java
```diff
@@ -0,0 +1,146 @@
...
+import org.junit.internal.AssumptionViolatedException;
+import org.openqa.selenium.TimeoutException;
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.Assert.fail;
+
+@WithPlugins({ "windows-slaves" }) @Ignore("Enable when ATH testing on windows is available.")
```
timja, line 43: is it possible to do this with an assume rather than ignore? so that anyone running on windows can test it without modifying the code?

rsandell: There are assumtions specified in the test as well as it needs some information from the one who runs it via system properties. If those are not there the test will be skipped.
So this `@Ignore` was mostly added for emphasis :shrug:

src/test/java/plugins/WindowsAgentsTest.java
```diff
@@ -0,0 +1,146 @@
...
+            fail("Slave has become online when it should remain offline");
+        } catch (TimeoutException timeout) {
+            //empty by design this is the expected behaviour
+        }
+    }
+
+    private String getWindowsProperty(String property) {
+        String p = System.getProperty(property);
+        if (p == null || p.isEmpty()) {
+            throw new AssumptionViolatedException(String.format("No windows property (%s) found, ignoring windows slaves test", property));
```
timja, line 134:
```suggestion
            throw new AssumptionViolatedException(String.format("No windows property (%s) found, ignoring windows test", property));
```

timja (commented):
bit wary of adding a test that isn't being run, is there a plan to add running of tests on windows / running of windows only tests here?

or are you planning to run these tests downstream somewhere?

rsandell:
I guess I will need some infra help to understand how to add running the tests on Windows..

jtnord (changes requested):
this relies (without specifyin) on the windows-slaves-plugin which intern relies on a feature that is being phased out by MSFT.

At this point - the registry tweaks to override micorosft secrutiy fix to keep this working are no longer available and so the plugin should be considered dead in the water...

https://github.com/jenkinsci/windows-slaves-plugin/tree/91ab0e4cca1f4eb9cd6cadd051fc1399deaf931d#notice-of-deprecation

I would suggest we just close this PR.

PR #1066, opened 2023-03-15
[#1053] Add tests for Ant Plugin

ampuscas:
related to #1053 

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer.java
@@ -42,10 +42,10 @@ public class SshAgentContainer extends JavaContainer {
         return resource("ed25519.pass").asText();
     }
 
-    public static SshSlaveLauncher configureSSHSlaveLauncher(DumbSlave agent, String host, int port) {
+    public SshSlaveLauncher configureSSHSlaveLauncher(DumbSlave agent) {
         SshSlaveLauncher launcher = agent.setLauncher(SshSlaveLauncher.class);
-        launcher.host.set(host);
-        launcher.port(port);
+        launcher.host.set(ipBound(22));
+        launcher.port(port(22));
         launcher.setSshHostKeyVerificationStrategy(SshSlaveLauncher.NonVerifyingKeyVerificationStrategy.class);
         return launcher;
     }
--- a/src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer.java
@@ -25,6 +25,8 @@
 package org.jenkinsci.test.acceptance.docker.fixtures;
 
 import org.jenkinsci.test.acceptance.docker.DockerFixture;
+import org.jenkinsci.test.acceptance.plugins.ssh_slaves.SshSlaveLauncher;
+import org.jenkinsci.test.acceptance.po.DumbSlave;
 
 /**
  * Jenkins agent with various login methods.
@@ -40,4 +42,11 @@ public class SshAgentContainer extends JavaContainer {
         return resource("ed25519.pass").asText();
     }
 
+    public static SshSlaveLauncher configureSSHSlaveLauncher(DumbSlave agent, String host, int port) {
+        SshSlaveLauncher launcher = agent.setLauncher(SshSlaveLauncher.class);
+        launcher.host.set(host);
+        launcher.port(port);
+        launcher.setSshHostKeyVerificationStrategy(SshSlaveLauncher.NonVerifyingKeyVerificationStrategy.class);
+        return launcher;
+    }
 }
--- a/src/test/java/plugins/AntPluginTest.java
+++ b/src/test/java/plugins/AntPluginTest.java
@@ -42,15 +42,7 @@ public class AntPluginTest extends AbstractJUnitTest {
     private static final String OPTS = "-showversion";
     private static final String BUILD_FILE = "custom-build-file.xml";
     private static final String FAKE_BUILD_FILE = "fake.xml";
-    private static final String SCRIPT_PIPELINE_ANT = "node {\n" +
-            "    withAnt(installation: '" + NATIVE_ANT_NAME + "') {\n" +
-            "        if (isUnix()) {\n" +
-            "            sh \"ant -version\"\n" +
-            "        } else {\n" +
-            "            bat \"ant -version\"\n" +
-            "        }\n" +
-            "    }\n" +
-            "}";
+
     FreeStyleJob job;
     private AntBuildStep step;
 
@@ -145,6 +137,15 @@ public class AntPluginTest extends AbstractJUnitTest {
     @WithPlugins({"workflow-job", "workflow-cps", "workflow-basic-steps", "workflow-durable-task-step"})
     public void testAntWrapper() {
         useSlave();
+        String SCRIPT_PIPELINE_ANT = "node (\"" + slave.getName() + "\") {\n" +
+                "    withAnt(installation: '" + NATIVE_ANT_NAME + "') {\n" +
+                "        if (isUnix()) {\n" +
+                "            sh \"ant -version\"\n" +
+                "        } else {\n" +
+                "            bat \"ant -version\"\n" +
+                "        }\n" +
+                "    }\n" +
+                "}";
         String antHome = setUpAntInstallation();
 
         String expectedVersion = "1.10.5"; // this is the version installed in the java container by the ubuntu bionic;
--- a/src/test/java/plugins/AntPluginTest.java
+++ b/src/test/java/plugins/AntPluginTest.java
@@ -2,7 +2,6 @@ package plugins;
 
 import com.google.inject.Inject;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.SystemUtils;
 import org.jenkinsci.test.acceptance.Matchers;
 import org.jenkinsci.test.acceptance.docker.DockerContainerHolder;
 import org.jenkinsci.test.acceptance.docker.fixtures.SshAgentContainer;
@@ -10,7 +9,6 @@ import org.jenkinsci.test.acceptance.junit.AbstractJUnitTest;
 import org.jenkinsci.test.acceptance.junit.WithPlugins;
 import org.jenkinsci.test.acceptance.plugins.ant.AntBuildStep;
 import org.jenkinsci.test.acceptance.plugins.ant.AntInstallation;
-import org.jenkinsci.test.acceptance.plugins.ssh_slaves.SshSlaveLauncher;
 import org.jenkinsci.test.acceptance.po.*;
 import org.junit.Before;
 import org.junit.Test;
@@ -43,7 +41,6 @@ public class AntPluginTest extends AbstractJUnitTest {
 
     private SshAgentContainer sshd;
     private DumbSlave agent;
-
     @Before
     public void setUp() {
         job = jenkins.jobs.create(FreeStyleJob.class);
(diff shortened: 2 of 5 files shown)


src/test/java/plugins/AntPluginTest.java
```diff
@@ -72,6 +99,67 @@ public void locallyInstalledAnt() {
...
+        assertThat(console, containsString("-D" + OK_PROP2));
+        assertThat(console, containsString("-D" + NOK_PROP1));
+        assertThat(console, containsString("-D" + NOK_PROP2));
+        assertThat(console, containsString("[echoproperties] " + OK_PROP1));
+        assertThat(console, containsString("[echoproperties] " + OK_PROP2));
+        assertThat(console, not(Matchers.containsRegexp("[echoproperties] " + NOK_PROP1, Pattern.MULTILINE)));
+        assertThat(console, not(Matchers.containsRegexp("[echoproperties] " + NOK_PROP2, Pattern.MULTILINE)));
+    }
+
+    @Test @Native("ant")
```
rsandell, line 149: From [looking at the code](https://github.com/jenkinsci/ant-plugin/blob/master/src/main/java/hudson/tasks/Ant.java#L505), there seems to be a download installer option for Ant. Perhaps the ant setup could be rewritten to use that instead of requiring it on the agent natively?

jtnord: that would change the intent of the test.

You can for always use a docker container with `ant` and `ssh` for the agent. and use a specific node (that will start the container)

e.g. something like https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer/Dockerfile with the addition of `ant`

https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer.java

jtnord: `SSHAgentContainer` already has `ant` installed 🥳 

https://github.com/jenkinsci/acceptance-test-harness/blob/608f30703d2b8377e251fcb7e2a7637d3c128a2d/src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer/Dockerfile

https://raw.githubusercontent.com/jenkinsci/docker-fixtures/master/src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/JavaContainer/Dockerfile

src/test/java/plugins/AntPluginTest.java
```diff
@@ -23,15 +32,74 @@ public class AntPluginTest extends AbstractJUnitTest {
...
+
+    private SshSlaveLauncher configureSSHSlaveLauncher(String host, int port) {
+        SshSlaveLauncher launcher = slave.setLauncher(SshSlaveLauncher.class);
+        launcher.host.set(host);
+        launcher.port(port);
+        launcher.setSshHostKeyVerificationStrategy(SshSlaveLauncher.NonVerifyingKeyVerificationStrategy.class);
+        return launcher;
+    }
+
+    private void useSlave() {
```
uhafner, line 83: Please rethink the naming, the term `slave` is discourage for several years now

src/test/java/plugins/AntPluginTest.java
```diff
@@ -23,15 +32,74 @@ public class AntPluginTest extends AbstractJUnitTest {
...
         job = jenkins.jobs.create(FreeStyleJob.class);
+        sshd = docker.get();
+
+    }
+
+    private SshSlaveLauncher configureDefaultSSHSlaveLauncher() {
+        return configureSSHSlaveLauncher(sshd.ipBound(22), sshd.port(22));
+    }
+
+    private SshSlaveLauncher configureSSHSlaveLauncher(String host, int port) {
```
jtnord, line 75: as this (and its caller) are probably used elsewhere - it would probably be refactored to be a helper method in the `SshAgentContainer` so it is easier to reuse across other tests.

ie - this plugins and probably the SSHPluginTest should call something like `sshd.configureSSHLauncher(agent)` at least for the host/port combination - the verification strategy and credentials may be a bit more test specific?

this way any changes to the ssh launcher can be updated in a single place and not in every test.

src/test/java/plugins/AntPluginTest.java
```diff
@@ -68,10 +137,69 @@ public void locallyInstalledAnt() {
...
+    @Test
+    @WithPlugins({"workflow-job", "workflow-cps", "workflow-basic-steps", "workflow-durable-task-step"})
+    public void testAntWrapper() {
+        useSlave();
+        String antHome = setUpAntInstallation();
+
+        String expectedVersion = "1.10.5"; // this is the version installed in the java container by the ubuntu bionic;
+
+        WorkflowJob workflowJob = jenkins.jobs.create(WorkflowJob.class);
+        workflowJob.script.set(SCRIPT_PIPELINE_ANT);
```
jtnord, line 153: you will need to tell the pipeline to use the ssh agent you have created - otherwise it will run anywhere.

this is inline in the pipeline definition `node(_the_agent_label) {`  or you would need to remove the ability to execute on the controller by setting the number of executors there to zero.

As the pipeline definition is not re-used in any other test it it would probably make sense to move the definition of it inline here.

src/test/java/plugins/AntPluginTest.java
```diff
@@ -23,15 +32,74 @@ public class AntPluginTest extends AbstractJUnitTest {
     private static final String INSTALL_VERSION_1_10 = "1.10.5";
     private static final String INSTALL_NAME_1_10 = "ant_" + INSTALL_VERSION_1_10;
 
+    private static final String NATIVE_ANT_NAME = "native_ant";
+
+    private static final String OK_PROP1 = "okPROP1=foo_bar_ok_1";
+    private static final String OK_PROP2 = "okPROP2=foo_bar_ok_2";
+    private static final String NOK_PROP1 = "nokPROP1=foo_bar_nok_1";
+    private static final String NOK_PROP2 = "nokPROP2=foo_bar_nok_2";
+    private static final String PROPERTIES = OK_PROP1+"\n"+OK_PROP2+"\n"+NOK_PROP1+"\n"+NOK_PROP2;
```
jtnord, line 41: as these appear to be specific to a single test would they be better off defined in that single test method?

src/test/java/plugins/AntPluginTest.java
```diff
@@ -68,10 +137,69 @@ public void locallyInstalledAnt() {
...
         job.save();
 
-        String expectedVersion = localAntVersion(antHome);
+        String expectedVersion = "1.10.5"; // this is the version installed in the java container by the ubuntu bionic
         job.startBuild().shouldSucceed().shouldContainsConsoleOutput(Pattern.quote(expectedVersion));
     }
 
+    @Test
+    @WithPlugins({"workflow-job", "workflow-cps", "workflow-basic-steps", "workflow-durable-task-step"})
+    public void testAntWrapper() {
```
jtnord, line 146:
```suggestion
    public void testWithAntPipelineBlock() {
```

src/test/java/plugins/AntPluginTest.java
```diff
@@ -23,15 +32,74 @@ public class AntPluginTest extends AbstractJUnitTest {
...
+    private static final String OK_PROP2 = "okPROP2=foo_bar_ok_2";
+    private static final String NOK_PROP1 = "nokPROP1=foo_bar_nok_1";
+    private static final String NOK_PROP2 = "nokPROP2=foo_bar_nok_2";
+    private static final String PROPERTIES = OK_PROP1+"\n"+OK_PROP2+"\n"+NOK_PROP1+"\n"+NOK_PROP2;
+    private static final String OPTS = "-showversion";
+    private static final String BUILD_FILE = "custom-build-file.xml";
+    private static final String FAKE_BUILD_FILE = "fake.xml";
+    private static final String SCRIPT_PIPELINE_ANT = "node {\n" +
+            "    withAnt(installation: '" + NATIVE_ANT_NAME + "') {\n" +
+            "        if (isUnix()) {\n" +
```
jtnord, line 47: as this is only used in one test and as that test should be excersized on the SSH agent only (never the controller) we know that this will be a unix docker agent (as we do not support windows containers here - and if we did that would be a lot of work!)  
so you can remove the `isUnix` if/esel and just assume it is always unix (after fixing the `node` block to use the approprate agent!

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer.java
```diff
@@ -40,4 +42,11 @@ public String getEncryptedEd25519PrivateKeyPassphrase() {
         return resource("ed25519.pass").asText();
     }
 
+    public static SshSlaveLauncher configureSSHSlaveLauncher(DumbSlave agent, String host, int port) {
```
jtnord, line 45:
```suggestion
    public SshSlaveLauncher configureSSHSlaveLauncher(DumbSlave agent) {
```
this should not be static and obtain the host and port from the current instance, callers will need to be updated.

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer.java
```diff
@@ -40,4 +42,11 @@ public String getEncryptedEd25519PrivateKeyPassphrase() {
...
+        SshSlaveLauncher launcher = agent.setLauncher(SshSlaveLauncher.class);
+        launcher.host.set(host);
+        launcher.port(port);
```
jtnord, line 48:
```suggestion
        launcher.port(sshd.port(22));
```

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer.java
```diff
@@ -40,4 +42,11 @@ public String getEncryptedEd25519PrivateKeyPassphrase() {
         return resource("ed25519.pass").asText();
     }
 
+    public static SshSlaveLauncher configureSSHSlaveLauncher(DumbSlave agent, String host, int port) {
+        SshSlaveLauncher launcher = agent.setLauncher(SshSlaveLauncher.class);
+        launcher.host.set(host);
```
jtnord, line 47:
```suggestion
        launcher.host.set(ipBound(22));
```

src/test/java/plugins/AntPluginTest.java
```diff
@@ -1,20 +1,29 @@
 package plugins;
 
+import com.google.inject.Inject;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.SystemUtils;
```
jtnord, line 5: unised import?

src/test/java/plugins/AntPluginTest.java
```diff
@@ -23,15 +32,46 @@ public class AntPluginTest extends AbstractJUnitTest {
...
+    private AntBuildStep step;
+
+    @Inject private DockerContainerHolder<SshAgentContainer> docker;
+
+    private SshAgentContainer sshd;
+    private DumbSlave agent;
 
     @Before
     public void setUp() {
         job = jenkins.jobs.create(FreeStyleJob.class);
```
jtnord, line 49: this is creating a Freestyle job even for the jobs that tests.
would make the tests quicker if this (and the probably `FreeStyle job filed) was just inlined in the tests that need it.

src/test/java/plugins/AntPluginTest.java
```diff
@@ -54,24 +94,96 @@ public void autoInstallMultipleAnt() {
         );
     }
 
-    @Test @Native("ant")
+    @Test
     public void locallyInstalledAnt() {
+        useCustomAgent();
```
jtnord, line 99:
```suggestion

        final String NATIVE_ANT_NAME = "native_ant"
        useCustomAgent();
```

src/test/java/plugins/AntPluginTest.java
```diff
@@ -54,24 +94,96 @@ public void autoInstallMultipleAnt() {
...
         AntInstallation ant = ToolInstallation.addTool(jenkins, AntInstallation.class);
```
jtnord, line 100: why not use `setUpAntInstallation` here?

src/test/java/plugins/AntPluginTest.java
ampuscas: `NATIVE_ANT_NAME` is used in several methods and tests, I thought not to multiply the declaration

jtnord (approved):

raul-arabaolaza (approved):

NotMyFault (approved):

PR #1080, opened 2023-03-27
Forward compatibility for refresh update sites button change

janfaracik:
Testing https://github.com/jenkinsci/jenkins/pull/7770


janfaracik:
plugins.WorkflowPluginTest.sshGitInsideDocker is the only test failing now, reckon that's related to the change? @timja

timja:
> plugins.WorkflowPluginTest.sshGitInsideDocker is the only test failing now, reckon that's related to the change? @timja

Unsure, but it is failing in plugin manager code so better safe than sorry.

I've reran it in case there was an upstream failure

src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
```diff
@@ -88,7 +88,7 @@ public void checkForUpdates() {
         // The check now button is a form submit (POST) with a redirect to the same page only if the check is successful.
         // We use the button itself to detect when the page has changed, which happens after the refresh has been done
         // And we check for the presence of the button again
-        WebElement checkButton = find(by.link("Check now"));
+        WebElement checkButton = find(by.css("form[action='checkUpdatesServer'] .jenkins-button"));
```
timja, line 91: better to use the id?

janfaracik: The button prior to this PR didn't have an ID

timja (approved):

PR #1091, opened 2023-04-03
Add test for declarative migration assistant and pipeline with unstable tests

ampuscas:
related to #1053 

### DIFF
--- a/src/test/java/plugins/DeclarativeAssistantMigrationTest.java
+++ b/src/test/java/plugins/DeclarativeAssistantMigrationTest.java
@@ -1,0 +1,59 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2023 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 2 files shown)


src/test/java/plugins/DeclarativeAssistantMigrationTest.java
```diff
@@ -0,0 +1,59 @@
...
+import org.jenkinsci.test.acceptance.po.FreeStyleJob;
+import org.jenkinsci.test.acceptance.po.ShellBuildStep;
+import org.junit.Test;
+import org.openqa.selenium.By;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.jenkinsci.test.acceptance.Matchers.containsString;
+import static org.jenkinsci.test.acceptance.Matchers.hasElement;
+
+@WithPlugins({"declarative-pipeline-migration-assistant@1.0.3","declarative-pipeline-migration-assistant-api@1.0.3"})
```
jtnord, line 37: the versions are ancient - so can be ommitted

```suggestion
@WithPlugins({"declarative-pipeline-migration-assistant","declarative-pipeline-migration-assistant-api"})
```

src/test/java/plugins/DeclarativeAssistantMigrationTest.java
```diff
@@ -0,0 +1,59 @@
...
+    public void basicDeclarativeTests() throws Exception {
+        FreeStyleJob j = jenkins.jobs.create( FreeStyleJob.class, "simple-job-to-declarative");
+        j.configure();
+        ShellBuildStep shell = j.addBuildStep( ShellBuildStep.class);
+        shell.command("echo 1");
+
+        j.apply();
+        j.save();
+
+        clickLink("To Declarative");
```
jtnord, line 52: clicking something can be asynchronous - webdriver attempts to wait but may not actually wait depending on how the link is.  As this seems like it should be a post the clicking will invoke javascript which means that this test is likely to be flaky.  in order to resolve that the link Element should be obtained, then clicked, then we should wait for it to become stale.  (or it is a in page migration with a dialog - which means we then also need to have some form of wait!)

src/test/java/plugins/DeclarativeAssistantMigrationTest.java
jtnord, line 1: this appears to be testing dec;atative pipeline migration not pipeline itself?
shoudl this be part of the PR or is the title wrong in the PR?

src/test/java/plugins/PipelineTest.java
```diff
@@ -0,0 +1,113 @@
...
+@WithPlugins({"git", "junit", "javadoc"})
+public class PipelineTest extends AbstractPipelineTest {
+
+    @Before
+    public void setup() {
+        MavenInstallation.installMaven(jenkins, "M3", "3.3.9");
+    }
+
+    @Test
+    public void testCompletePipeline() {
```
jtnord, line 57: whilst it appears as though the functionality is covered in [`WorkflowPluginTest#linearFlow()`](https://github.com/jenkinsci/acceptance-test-harness/blob/3bb860d86858dd44d7500e3b51776dc28e408f22/src/test/java/plugins/WorkflowPluginTest.java#L122-L175} the test is subtly different and is exercising a useful path of a build with failing tests (where the failing tests are recorded and checked)

jtnord (approved):

jtnord (commented):
looks ok - however ` clickLink("To Declarative");` would appear to be a source of flakyness?

jtnord (approved):

Issue #963, opened 2022-10-03
Use Jenkins project mirrors / caches


timja:
### What feature do you want to see added?

Ideally we should utilise maven caches if available (without forcing people to use them who use the framework)

This build failed:
https://ci.jenkins.io/blue/organizations/jenkins/Core%2Facceptance-test-harness/detail/master/601/pipeline/

Because of:
```
[2022-10-02T22:45:56.251Z] [ERROR] Error executing Maven.
[2022-10-02T22:45:56.251Z] [ERROR] Extension io.jenkins.tools.incrementals:git-changelist-maven-extension:1.4 or one of its dependencies could not be resolved: Could not transfer artifact org.apache.maven.resolver:maven-resolver-util:jar:1.0.3 from/to central (https://repo.maven.apache.org/maven2): transfer failed for https://repo.maven.apache.org/maven2/org/apache/maven/resolver/maven-resolver-util/1.0.3/maven-resolver-util-1.0.3.jar
[2022-10-02T22:45:56.251Z] [ERROR] Caused by: Could not transfer artifact org.apache.maven.resolver:maven-resolver-util:jar:1.0.3 from/to central (https://repo.maven.apache.org/maven2): transfer failed for https://repo.maven.apache.org/maven2/org/apache/maven/resolver/maven-resolver-util/1.0.3/maven-resolver-util-1.0.3.jar
[2022-10-02T22:45:56.251Z] [ERROR] Caused by: Could not transfer artifact org.apache.maven.resolver:maven-resolver-util:jar:1.0.3 from/to central (https://repo.maven.apache.org/maven2): transfer failed for https://repo.maven.apache.org/maven2/org/apache/maven/resolver/maven-resolver-util/1.0.3/maven-resolver-util-1.0.3.jar
[2022-10-02T22:45:56.251Z] [ERROR] Caused by: transfer failed for https://repo.maven.apache.org/maven2/org/apache/maven/resolver/maven-resolver-util/1.0.3/maven-resolver-util-1.0.3.jar
[2022-10-02T22:45:56.251Z] [ERROR] Caused by: Connect to repo.maven.apache.org:443 [repo.maven.apache.org/146.75.32.215] failed: Connection timed out (Connection timed out)
[2022-10-02T22:45:56.251Z] [ERROR] Caused by: Connection timed out (Connection timed out)
```

lemeurherve:
[...]

PR #1097, opened 2023-04-05
Revamp build/CI

### LINKED ISSUE #736: Use a fixed version of core in tests and automate updates to it

basil:
Fixes #736. Preparatory work for #963. See self-review for details.

High-level changes:

- Retry CI stage on agent failure, consistent with other jobs.
- Publish incrementals, just like other jobs.
- Set `maven.repo.local` inside `WORKSPACE_TMP` in CI, just like jenkinsci/bom#1842.
- Define `jenkins.version` in `pom.xml`, where user-submitted PRs may set it to an incremental build. This version will be updated by Dependabot.
- Define an LTS Maven profile with the LTS version. This will need to be updated by the release lead after shipping an LTS.
- When `latest`, `lts`, or a version number are provided, have `run.sh` delegate to Maven rather than trying to do the download itself. Continue to support usages of a URL or a filesystem path to the WAR.
- Update documentation.
- Observe `MAVEN_SETTINGS`, if set. This is preparatory work for #963.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -34,9 +34,19 @@ if (needSplittingFromWorkspace) {
 
 branches['CI'] = {
   stage('CI') {
-    node('maven-11') {
-      checkout scm 
-      sh "mvn verify --no-transfer-progress -DskipTests -P jenkins-release"
+    retry(count: 2, conditions: [kubernetesAgent(handleNonKubernetes: true), nonresumable()]) {
+      node('maven-11') {
+        checkout scm
+        def mavenOptions = [
+          '-Dset.changelist',
+          '-DskipTests',
+          '-U',
+          'clean',
+          'install',
+        ]
+        infra.runMaven(mavenOptions, 11)
+        infra.prepareToPublishIncrementals()
+      }
     }
   }
 }
@@ -71,7 +81,7 @@ for (int i = 0; i < splits.size(); i++) {
                     set-java.sh $javaVersion
                     eval \$(vnc.sh)
                     java -version
-                    run.sh firefox ${jenkinsUnderTest} -Dmaven.test.failure.ignore=true -DforkCount=1 -B
+                    run.sh firefox ${jenkinsUnderTest} -Dmaven.repo.local=${WORKSPACE_TMP}/m2repo -Dmaven.test.failure.ignore=true -DforkCount=1 -B
                 """
               }
             }
@@ -83,3 +93,4 @@ for (int i = 0; i < splits.size(); i++) {
   }
 }
 parallel branches
+infra.maybePublishIncrementals()
--- a/pom.xml
+++ b/pom.xml
@@ -355,13 +355,6 @@
       <artifactId>zt-zip</artifactId>
       <version>1.15</version>
     </dependency>
-    <!-- Keep in sync with mockito-core's version -->
-    <dependency>
-      <groupId>net.bytebuddy</groupId>
-      <artifactId>byte-buddy</artifactId>
-      <version>1.12.14</version>
-      <scope>test</scope>
-    </dependency>
     <!--
       This is just for dependency updates.
 
--- a/pom.xml
+++ b/pom.xml
@@ -32,6 +32,7 @@
     <trimStackTrace>false</trimStackTrace> <!-- surefire -->
     <spotbugs.skip>true</spotbugs.skip>
 
+    <jenkins.version>2.399</jenkins.version>
     <selenium.version>4.8.1</selenium.version>
     <guava.version>31.1-jre</guava.version> <!-- aligned with selenium -->
     <aether.version>1.1.0</aether.version>
@@ -361,6 +362,24 @@
       <artifactId>byte-buddy</artifactId>
       <version>1.12.14</version>
       <scope>test</scope>
+    </dependency>
+    <!--
+      This is just for dependency updates.
+
+      TODO: Delete this dependency and figure out how to automatically update
+      this dependency some other way.
+    -->
+    <dependency>
+      <groupId>org.jenkins-ci.main</groupId>
+      <artifactId>jenkins-core</artifactId>
+      <version>${jenkins.version}</version>
+      <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>*</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.mockito</groupId>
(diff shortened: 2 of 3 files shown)


Jenkinsfile
```diff
@@ -34,9 +34,19 @@ if (needSplittingFromWorkspace) {
 
 branches['CI'] = {
   stage('CI') {
-    node('maven-11') {
-      checkout scm 
-      sh "mvn verify --no-transfer-progress -DskipTests -P jenkins-release"
+    retry(count: 2, conditions: [kubernetesAgent(handleNonKubernetes: true), nonresumable()]) {
```
basil, line 37: Consistent with core and BOM: even though in practice these are Kubernetes agents, we respect the contract provided by the infrastructure team by not relying on this implementation detail.

Jenkinsfile
```diff
@@ -34,9 +34,19 @@ if (needSplittingFromWorkspace) {
...
+      node('maven-11') {
+        checkout scm
+        def mavenOptions = [
+          '-Dset.changelist',
```
basil, line 41: Needed for producing incrementals.

Jenkinsfile
```diff
@@ -34,9 +34,19 @@ if (needSplittingFromWorkspace) {
...
+          '-DskipTests',
+          '-U',
+          'clean',
+          'install',
```
basil, line 45: Needed for `infra.prepareToPublishIncrementals()` (as opposed to `verify`).

Jenkinsfile
```diff
@@ -34,9 +34,19 @@ if (needSplittingFromWorkspace) {
...
+        ]
+        infra.runMaven(mavenOptions, 11)
```
basil, line 47: Internally respects `MAVEN_SETTINGS`, preparing for the potential use of the artifact caching proxy.

Jenkinsfile
```diff
@@ -71,7 +81,7 @@ for (int i = 0; i < splits.size(); i++) {
                     set-java.sh $javaVersion
                     eval \$(vnc.sh)
                     java -version
-                    run.sh firefox ${jenkinsUnderTest} -Dmaven.test.failure.ignore=true -DforkCount=1 -B
+                    run.sh firefox ${jenkinsUnderTest} -Dmaven.repo.local=${WORKSPACE_TMP}/m2repo -Dmaven.test.failure.ignore=true -DforkCount=1 -B
```
basil, line 84: See jenkinsci/bom#1842.

pom.xml
```diff
@@ -362,6 +363,24 @@
...
+      <artifactId>jenkins-core</artifactId>
+      <version>${jenkins.version}</version>
+      <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>*</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
```
basil, line 383: Copying the trick I did in `plugin-compat-tester` to define an unused test-scoped dependency to get Dependabot to automatically update `jenkins.version` (actually used further below by the `copy` mojo, which Dependabot can't update). I have confirmed this trick works. I am not interested in learning how to use any other dependency updating tools. I trust that those who are enthusiastic about other tools can find a better way to do this after this PR lands (or stick with this solution, which is "good enough", if they are not interested).

pom.xml
```diff
@@ -788,5 +807,48 @@ and
...
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+    <profile>
+      <id>lts</id>
+      <properties>
+        <jenkins.version>2.387.1</jenkins.version>
```
basil, line 850: This will need to be updated by the release lead after shipping an LTS release, just like in BOM.

src/main/resources/ath-container/run.sh
```diff
@@ -7,22 +7,30 @@ if [[ $# -lt 2 ]]; then
...
+		# Run full suite in Firefox against LTS:
+		$ ./run firefox lts
 	USAGE
 	exit 2
 fi
 
+MVN='mvn -V -e -ntp'
+if [[ -n ${MAVEN_SETTINGS-} ]]; then
+	MVN="${MVN} -s ${MAVEN_SETTINGS}"
+fi
```
basil, line 32: Observe `MAVEN_SETTINGS` if defined in preparation for #963.

pom.xml
```diff
@@ -788,5 +807,48 @@ and
...
+                  </artifactItems>
+                  <outputDirectory>${project.build.directory}</outputDirectory>
+                  <stripVersion>true</stripVersion>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
```
basil, line 846: If you are reading this PR from top to bottom, this hunk will only make sense after you have read the changes to `run.sh`.

src/main/resources/ath-container/run.sh
```diff
@@ -39,50 +47,33 @@ fi
...
-		war=jenkins-lts.war
-		url=https://updates.jenkins.io/stable/latest/jenkins.war
+		# Maven will download this in the process-test-resources phase
+		war=target/jenkins-war.war
+		extra_args='-Plts'
+		;;
+	*.war)
+		download "$war" "jenkins.war" || exit 1
+		war=jenkins.war
+		;;
```
basil, line 66: Due to the `[[ ! -f $war ]]; then` a few lines above, this is the case statement for URLs, not local file paths.

basil:
Used my admin powers to replay the build with the `Jenkinsfile` changes here: https://ci.jenkins.io/job/Core/job/acceptance-test-harness/job/PR-1097/12/

timja (approved):

pom.xml
timja: Filed https://github.com/jenkinsci/acceptance-test-harness/issues/1098 for later improvement

PR #1099, opened 2023-04-05
Adding varargs test for ScriptSecurityPluginTest

julieheard:
Related to https://github.com/jenkinsci/acceptance-test-harness/issues/1053

This adds the test for JENKINS-48364 - varargs not recognized when first arg is null

### DIFF
--- a/src/test/java/plugins/ScriptSecurityPluginTest.java
+++ b/src/test/java/plugins/ScriptSecurityPluginTest.java
@@ -24,16 +24,24 @@
 
 package plugins;
 
+import org.jenkinsci.test.acceptance.Matchers;
 import org.jenkinsci.test.acceptance.junit.AbstractJUnitTest;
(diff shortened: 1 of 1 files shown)


timja (commented):
What is special about this test that means it belongs in this repository?

This seems like something very specific (which already has coverage around it for similar things already in this test class)

julieheard:
Hi @timja,

This test is testing for a pipeline having null parameters in it. In [JENKINS-48364](https://issues.jenkins.io/browse/JENKINS-48364) the issue was that some methods were no longer recognised as valid methods if the first parameters value is null. This test inclusion will allow us to see if there is a regression in this plugin.

I do see your point, as this is only testing script-security version 1.39, although looking at [usages of this plugin](https://stats.jenkins.io/pluginversions/script-security.html), it looks like it is still being used.

PR #1108, opened 2023-04-14
[#1053] More git plugin tests

rsandell:
See #1053 

- [X] Link to relevant issues in GitHub or Jira
- [X] Link to relevant pull requests, esp. upstream and downstream changes


src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/GitListContainer/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+
+# install SSHD, Git and zip
+RUN apt-get update && apt-get install -y \
+    openssh-server \
+    git \
+    zip \
+    tar \
+    curl \
+    nginx-full \
+    php8.1-fpm \
```
rsandell, line 21: php-fqm fails to start for unknown reason. Still investigating.

src/test/java/plugins/GitPluginRepositoryBrowserTest.java
```diff
@@ -0,0 +1,335 @@
...
+        );
+
+        String revision = getRevisionFromConsole(b.getConsole());
+        RepositoryBrowserUtils rbUtils = new RepositoryBrowserUtils(driver, hostHTTP, portHTTP);
+        String revisionUrl = rbUtils.getCommitUrl(revision, RepositoryBrowserUtils.RepositoryBrowserType.DEFAULT_BROWSER);
+        assertThat(
+                visit(changesUrl).getPageSource(),
+                Matchers.containsRegexp(revisionUrl, Pattern.MULTILINE)
+        );
+        String pageSource = rbUtils.visit(revision).getPageSource();
```
rsandell, line 186: Fails because php-fqm hasn't started correctly in the gitlist container.

src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/GitListContainer/Dockerfile
rsandell: The original dockerfile was based on Ubuntu 14, so I have done tons of changes to upgrade it to Ubuntu 22. U14 didn't fully work locally for me so that's why I started this U22 upgrade work.

rsandell:
But I see there are more fundamental errors on CI than there is on my machine :thinking:

timja:
cc @MarkEWaite

MarkEWaite:
I think that we should reconsider whether it is worth adding this repository browser test to the acceptance tests.

It is testing a relatively low usage browser (Gitlist).

There is a [JenkinsRule based test](https://github.com/jenkinsci/git-plugin/blob/master/src/test/java/hudson/plugins/git/CredentialsUserRemoteConfigTest.java) inside the git plugin repository that checks Pipeline definitions are accepted for [many different repository browser configurations](https://github.com/jenkinsci/git-plugin/blob/6d9979c480daa1d0e89d694bdcbebcc77b122c73/src/test/java/hudson/plugins/git/CredentialsUserRemoteConfigTest.java#L207).  If we need more testing of the repository browser, I'd rather extend the existing test than create a new test in the acceptance test harness.

The [test coverage report](https://ci.jenkins.io/job/Plugins/job/git-plugin/job/master/1158/coverage/) for the `browser` package shows that most of the repository browser classes have at least 50% branch coverage and have over 50% line coverage from the existing tests within the plugin.  I don't think there is enough benefit from this added test to justify the cost to run it and maintain it in the future.

rsandell:
@MarkEWaite Is it only the GitList test that you are concerned about or all of them?
I can understand that that test is complex and a burden to maintain and I would rather burn it with fire. 
But the other browser tests in the class are relatively simple in nature.

MarkEWaite:
> @MarkEWaite Is it only the GitList test that you are concerned about or all of them?

All of them.  Testing Phabricator is irrelevant because the Phabricator has been [unmaintained since July 2021](https://www.phacility.com/phabricator/).  The other cases are already tested in the git plugin.   If more tests are needed, they should be performed in the git plugin without the overhead of a web browser driving the test.  The JavaScript in these cases is not complicated enough to justify the overhead of a web browser.

rsandell:
Well none of the unit tests you linked to are actually testing clicking the drop down and selecting a different broswer and looking that the browser is applied.
I suggest we keep at least one simple test that checks the UI works as intended.

MarkEWaite:
> Well none of the unit tests you linked to are actually testing clicking the drop down and selecting a different browser and looking that the browser is applied.

I think that is an action using components that are used widely throughout Jenkins.  It seems to me that clicking a dropdown menu is already very well tested and won't benefit us much by adding a test that is specific to the repository browser dropdown menu.  However, I could accept a single test of the dropdown menu without the overhead of the Gitlist container setup.  I'd prefer that it test a commonly used value like GitLab or GitHub rather than a rarely used git provider like Phabricator.

MarkEWaite:
> Mark, I have a bit of trouble understanding your motivation for not adding more ATH tests? It is the best end to end test suite we've got that actually runs automated tests in as close to a real setup as we can. 

I think that the maintenance cost of end to end testing is high enough that we owe it to ourselves to assure that the tested functionality really needs an end to end test.  In this case, I believe that the benefits gained from the end to end test do not justify the maintenance cost of the test.

I don't object to end to end tests that confirm something that is not readily tested by lower level tests.  Those end to end tests would be adding something that can't be done with a lower level test.  I'm struggling to see that UI serialization / deserialization and other elements of the dropdown are risky enough that they need an end to end test in the git plugin.

> HtmlUnit tests are plagued with bad classpath setups and isn't running in a real browser and the PCT is inheriting all that plus flakes due to development setup changes that shouldn't affect runtimes. 

I agree that HtmlUnit tests are not running in a real browser.  I believe that is a benefit, not a detriment.  It makes the tests more reliable and more readily executed on many different systems.

As far as I can tell, the plugin bill of materials execution of automated tests has far fewer flaky tests than the acceptance test harness.  I think that test reliability is more of an argument for testing without a real web browser than an argument for testing with a real web browser.

> Yes the ATH is not perfect but from an end to end testing environment I believe we should extend it and not hinder it :)

I don't remember the last time that an automated test from a real web browser found something that benefited me in maintaining the git plugin.  I do remember multiple times when there were reports of test failures that were due to brittle end to end tests rather than real issues with the code.  I don't want to spend time investigating brittle end to end test failures.

rsandell:
You both are making excellent points :) I see more clearly now.
So let's close this one and I'll make a "simple" configRoundtrip test to the git plugin instead :)

PR #1113, opened 2023-04-19
Collect Launchable test results for ATH workspace in ATH builds

basil:
See [the design document](https://docs.google.com/document/d/12LdoAFA566P4LgHhu1L-fuOQuhipxk77tUDXQdNmhss/edit) for a thorough explanation of the design. This PR begins to implement the design described in that document by implementing test result collection for the ATH workspace in ATH build. (Test result collection for the ATH workspace in core builds is being done in https://github.com/jenkinsci/jenkins/pull/7857.) Test result collection is the easy part of integrating Launchable into our CI system; the difficult part (as explained in the design document) is commit collection, which I plan to work on after this. (As part of commit collection, I have various PRs open to add Git hash information to `MANIFEST.MF` for core, core components, and all plugins, so that change will need a little more time.) The ultimate goal, once the model is trained, is to increase velocity, decrease costs, and shift left. For example, once the model is trained, we could query Launchable for a subset of relevant ATH tests and run them within core PRs labelled with `web-ui` (for example). The purpose of this PR is not to discuss what to do with Launchable subsets in the long term but to begin to train the model so that we can have that discussion a month or so after commit collection and test result collection have been enabled for each Launchable workspace. To test this PR I re-ran the build with the `Jenkinsfile` changes, using my administrative privileges.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -63,8 +63,11 @@ for (int i = 0; i < splits.size(); i++) {
          retry(count: 2, conditions: [agent(), nonresumable()]) {
           node('docker-highmem') {
             checkout scm
+            def browser = 'firefox'
             def image = docker.build('jenkins/ath', '--build-arg uid="$(id -u)" --build-arg gid="$(id -g)" ./src/main/resources/ath-container/')
-            image.inside('-v /var/run/docker.sock:/var/run/docker.sock --shm-size 2g') {
+            sh 'mkdir -p target/ath-reports && chmod a+rwx target/ath-reports'
+            def cwd = pwd()
+            image.inside("-v /var/run/docker.sock:/var/run/docker.sock -v '${cwd}/target/ath-reports:/reports:rw' --shm-size 2g") {
               def exclusions = splits.get(index).join('\n')
               writeFile file: 'excludes.txt', text: exclusions
               realtimeJUnit(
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -63,8 +63,11 @@ for (int i = 0; i < splits.size(); i++) {
          retry(count: 2, conditions: [agent(), nonresumable()]) {
           node('docker-highmem') {
             checkout scm
+            def browser = 'firefox'
```
basil, line 66: Pulling out the browser into a local variable so I can pass it both to `run.sh` and to `launchable record tests` as the [flavor][(https://www.launchableinc.com/docs/sending-data-to-launchable/using-the-launchable-cli/recording-test-results-with-the-launchable-cli/use-flavors-to-run-the-best-tests-for-an-environment/), along with platform (Linux or Windows) and JDK version. It is good to start sending this information to Launchable it allows Launchable to produce better subsets in the future. For example, when producing a subset of tests for Windows test runs it will try to pick tests that are more relevant to Windows builds. The idea here is that if we ever change the matrix to include more browsers, the Launchable plumbing for browser flavor would already be taken care of.

Jenkinsfile
```diff
@@ -63,8 +63,11 @@ for (int i = 0; i < splits.size(); i++) {
...
             def image = docker.build('jenkins/ath', '--build-arg uid="$(id -u)" --build-arg gid="$(id -g)" ./src/main/resources/ath-container/')
-            image.inside('-v /var/run/docker.sock:/var/run/docker.sock --shm-size 2g') {
+            sh 'mkdir -p target/ath-reports && chmod a+rwx target/ath-reports'
```
basil, line 68: This change and the lines below are to create a directory on the Docker host for ATH reports so that we can later send them to Launchable. Why not install Launchable inside the ATH container and send the data to Launchable from inside the container, you might ask. Well, that would work as well, but this method is consistent with how test result collection is taking place in Jenkins core ATH builds, and it also enables us to use the common Launchable infrastructure used in other places like core and BOM. So, for example, when https://github.com/jenkins-infra/helpdesk/issues/3484 is implemented, we would get that for free here rather than having to maintain a parallel implementation of Launchable package installation in this repository.

Jenkinsfile
```diff
@@ -81,10 +84,23 @@ for (int i = 0; i < splits.size(); i++) {
...
+            launchable.install()
+            withCredentials([string(credentialsId: 'launchable-jenkins-acceptance-test-harness', variable: 'LAUNCHABLE_TOKEN')]) {
+              launchable('verify')
+              /*
+               * TODO Create a Launchable build and session earlier, and replace "--no-build" with
+               * "--session" to associate these test results with a particular build. The commits
+               * associated with the Launchable build should be the commits of the transitive closure of
+               * the Jenkins WAR under test in this build as well as the commits of the transitive closure
+               * of the ATH JAR.
+               */
```
basil, line 101: A bit of foreshadowing as to what is involved in the next step, commit collection. At present, this repository only pins the Jenkins core version, not plugin versions, so the relevant commits are the transitive closure of the core under test and the ATH under test. Collecting this with mathematical precision will require creating a new tool, but for the short term I plan on at least feeding Launchable the commits from Jenkins core itself (omitting its transitive dependencies like Stapler) and ATH itself (omitting its transitive dependencies like Remoting) since that is easy to implement (see my other PRs to `pom` for example). (Anything less than that, and the model would be so deprived of useful data that it would be pointless.) The idea is that this prototype will take shape incrementally, the model becoming more accurate as more and more of the design document is implemented. Obviously a fully transitive dependency analysis is needed to give the model the highest degree of confidence, but we don't have to wait for that to begin training the model with a fair degree of confidence.

timja (approved):

PR #1118, opened 2023-04-22
Add Jenkins core commit to Launchable build

basil:
This PR fetches the Jenkins WAR and passes its Git hash to Launchable when recording each build, thus defining each Launchable build with the most basic information it needs: the commit hashes of the primary repositories being tested. This is done with a fairly basic shell script that fishes the value out of `MANIFEST.MF` in the WAR file: transitive dependencies are not handled. This is intended to be a prototype solution that allows us to get up and running but should be improved later to cover transitive dependencies. In my conversations with Launchable product management, I was told that dealing with transitive dependencies is often a matter of diminishing returns, so it is better to get started with basic information to begin model training and go back and improve the data later if needed. If we keep going with Launchable, I intend to write some sort of tool that can collect the transitive dependency information we need for core, BOM, ATH, and plugin builds and send it to Launchable. That would be more of a medium to long term effort, while this PR covers the short term need to send basic commit information to Launchable.


Jenkinsfile
```diff
@@ -56,13 +56,17 @@ stage('Record builds and sessions') {
           // TODO Launchable not yet supported on LTS line
           return
         }
+        infra.withArtifactCachingProxy {
+          sh "DISPLAY=:0 ./run.sh firefox ${jenkinsVersion} -Dmaven.repo.local=${WORKSPACE_TMP}/m2repo -B clean process-test-resources"
+        }
```
basil, line 61: Using this script to fetch the WAR file, the same way we would a few lines down below. The `DISPLAY=:0` and Firefox arguments aren't needed for downloading the WAR file, but they are needed for running this script.

src/main/resources/ath-container/run.sh
```diff
@@ -74,6 +74,10 @@ fi
...
 shift 2
 
+if [[ $# -eq 0 ]]; then
+	extra_args+=' test'
+fi
+
 set -x
 
-BROWSER=$browser JENKINS_WAR=$war $MVN test $extra_args "$@"
+BROWSER=$browser JENKINS_WAR=$war $MVN $extra_args "$@"
```
basil, line 83: I needed to invoke this script with `process-test-resources` instead of `test` to download the WAR file, so making the `test` argument optional. To preserve compatibility I am setting it if it was not set by the caller, so the behavior is the same as before when no argument is passed in.

timja (approved):

NotMyFault (approved):

PR #1119, opened 2023-04-24
Address `#withTimeout()` deprecations

NotMyFault:
Addresses the deprecation of `#withTimeout()` using with `TimeUnit` by switching to `Duration#of*`, according to the javadocs.


basil (approved):
Thanks for the PR!

PR #1154, opened 2023-05-04
Forward compatibility for core-7911

mawinter69:
https://github.com/jenkinsci/jenkins/pull/7911 replaces the dedicated delete pages with a confirmation dialog.
Adjust the tests so it is able to handle both variants
See [JENKINS-13545](https://issues.jenkins.io/browse/JENKINS-13545)

### Submitter checklist


src/main/java/org/jenkinsci/test/acceptance/po/User.java
```diff
@@ -91,8 +91,12 @@ public User fullName(String fullName) {
     }
 
     public void delete() {
-        visit("delete");
-        clickButton("Yes");
+        try {
+            visit("delete");
+            clickButton("Yes");
+        } catch (NoSuchElementException nse) {
+            runThenConfirmAlert(() -> clickLink("Delete"),2);
```
timja, line 98: generally we would put the new behaviour first and then the old behaviour would be a fallback, it makes it easier when debugging tests as you aren't jumping into exception cases all the time and getting confused it's expected

mawinter69: I wonder what is the main purpose if this method. If it is just to delete the view we can also just directly call doDelete which hasn't changed. If it is to test functionality of the delete link and dialog then of course we need to keep. But such code would then belong to the test tree and not the main tree

timja: purpose would be to delete the user. but generally i think most of this is done through UI so delete link would make sense

NotMyFault (approved):

timja (approved):

Issue #988, opened 2022-11-05
Don't build ATH image on every split


timja:
It takes ~3.5 minutes to build and has flaked at least once.

It would be good to build it once and run a smoke test on it and then use a published image for the rest, or something like that.
Or even build and publish and then use the published version.

dduportal:
Make sense: since https://github.com/jenkinsci/acceptance-test-harness/pull/905 the image is prebuilt and pushed to an external registry.

Since this build uses expensive highmem instances, these minutes are good to avoid spending.

For information, nthe infra team always provides the same `jenkins` user with an UID of `1001`as per https://github.com/jenkins-infra/packer-images/blob/977ec35a3cba91abc7c0d3fe704467d450254005/provisioning/ubuntu-provision.sh#L31-L35 so that should be a nice improvement.

dduportal:
https://github.com/jenkinsci/acceptance-test-harness/pull/1168

PR #1168, opened 2023-05-11
chore(Jenkinsfile) Use `jenkins/ath` instead of rebuilding it

dduportal:
Fixes https://github.com/jenkinsci/acceptance-test-harness/issues/988

### Submitter checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -107,10 +107,9 @@ for (int i = 0; i < splits.size(); i++) {
         retry(count: 2, conditions: [agent(), nonresumable()]) {
           node('docker-highmem') {
             checkout scm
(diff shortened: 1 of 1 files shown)


timja (commented):
Nice!

It's switched from previously a consistent  range of 
2m50 to 3m building the image

to
1m to 1m 10  pulling the image.

Further optimisation could be done to pre-pull the image to the VM image or use a container registry local to the cloud provider I suspect, but for now that's a lot better

timja (approved):

basil:
But does this not use the last published version of the image rather than the one defined in code under test? That is not OK from a correctness perspective, because that means a change to the image made in a PR (the code under test) will not take effect in that PR. This also makes it impossible to change both the image and the consumer of the image in the same PR, which was possible before this. A better approach from a correctness perspective would be to build the image from the code under test once and then consume it from each parallel branch.

basil:
Causes #1193

PR #1171, opened 2023-05-12
Restore compatibility with the older Jenkins versions

c00ler:
This PR adds a fallback selector for the `Check now` button that was used in the older Jenkins versions. See: https://github.com/jenkinsci/acceptance-test-harness/pull/1080

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/PluginManager.java
@@ -89,7 +89,12 @@ public class PluginManager extends ContainerPageObject {
         // The check now button is a form submit (POST) with a redirect to the same page only if the check is successful.
         // We use the button itself to detect when the page has changed, which happens after the refresh has been done
         // And we check for the presence of the button again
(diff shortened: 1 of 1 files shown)


timja (commented):
What version are you adding compatibility for?

We generally maintain compatibility for weekly and LTS only.

c00ler:
> What version are you adding compatibility for?
> 
> We generally maintain compatibility for weekly and LTS only.

@timja Thank you for the quick reply. We are testing against the version `2.356`

timja:
Why? Jenkins weekly versions aren't supported outside of currently weekly normally.

jtnord:
`2.387` via the LTS line is currently the oldest supported version in this project. (Soon to be a newer version due to the upcoming LTS.

Whilst we have been poor at removing workaround code in the past this is something we are aiming to improve (the xpath expressions are getting extremely long) so I would expect more breakage as we try and remove debt.

If you need support for an old version (why)then I would look at an older release or a git sha depending on how you consume this project.

c00ler:
@timja @jtnord Thank you for the explanation. I've updated the Jenkins version on our side.

PR #1177, opened 2023-05-15
Move to Guice 6

NotMyFault:
Closes https://github.com/jenkinsci/acceptance-test-harness/pull/1175

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -10,7 +10,8 @@
     "org.apache.maven:maven-settings-builder",
     "org.apache.maven.resolver:maven-resolver-connector-basic",
     "org.apache.maven.resolver:maven-resolver-transport-file",
-    "org.apache.maven.resolver:maven-resolver-transport-http"
+    "org.apache.maven.resolver:maven-resolver-transport-http",
+    "com.google.inject:guice"
   ],
   "labels": [
     "dependencies"
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -16,8 +16,7 @@
     {
       "allowedVersions": "<7.0.0",
       "matchPackageNames": ["com.google.inject:guice"],
-      "description": "We focus on Guice 6 until core adapts 7",
-      "enabled": true
+      "description": "We focus on Guice 6 until core adapts 7"
     }
   ],
   "labels": [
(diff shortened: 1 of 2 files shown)


.github/renovate.json
```diff
@@ -10,7 +10,8 @@
     "org.apache.maven:maven-settings-builder",
     "org.apache.maven.resolver:maven-resolver-connector-basic",
     "org.apache.maven.resolver:maven-resolver-transport-file",
-    "org.apache.maven.resolver:maven-resolver-transport-http"
+    "org.apache.maven.resolver:maven-resolver-transport-http",
+    "com.google.inject:guice"
```
basil, line 14: Better to only allow versions < 7.0 with [`allowedversions`](https://docs.renovatebot.com/configuration-options/#allowedversions) than to block all updates, I think. Otherwise, we won't get PRs for new 6.x versions that we would presumably want to have.

NotMyFault: Oh right, good catch!

.github/renovate.json
```diff
@@ -12,6 +12,14 @@
     "org.apache.maven.resolver:maven-resolver-transport-file",
     "org.apache.maven.resolver:maven-resolver-transport-http"
   ],
+  "packageRules": [
+    {
+      "allowedVersions": "<7.0.0",
+      "matchPackageNames": ["com.google.inject:guice"],
+      "description": "We focus on Guice 6 until core adapts 7",
+      "enabled": false
```
basil, line 20: Why is this rule disabled? I could understand disabling it if we did not want to get PRs for this package, but in fact we do want to get PRs for version 6.x of this package.

NotMyFault, line 20:
```suggestion
      "enabled": true
```

.github/renovate.json
```diff
@@ -12,6 +12,14 @@
     "org.apache.maven.resolver:maven-resolver-transport-file",
     "org.apache.maven.resolver:maven-resolver-transport-http"
   ],
+  "packageRules": [
+    {
+      "allowedVersions": "<7.0.0",
+      "matchPackageNames": ["com.google.inject:guice"],
+      "description": "We focus on Guice 6 until core adapts 7",
+      "enabled": true
```
basil, line 20: This is enabled by default, so this line is unnecessary.

```suggestion
```

basil (approved):

.github/renovate.json
basil: (With the removal of the comma in the preceding line as well)

timja, line 20:
```suggestion
      "description": "We focus on Guice 6 until core adapts 7"
```

PR #1203, opened 2023-05-31
Update LTS profile

NotMyFault:


basil:
> Am I missing something @basil?

Nope, what you wrote seems accurate.

NotMyFault:
Does your backport need to be released? This PR fails for the same reason as before.

basil:
Yes, it does.

PR #1221, opened 2023-06-05
Migrate from mock security realm

jtnord:
the `mock-security-realm` is not a production quality realm and the tests here can trivially use Jenkins' internal database realm.

For CLoudBees' internal use this also means we do not need to hit the update center as the other plugins being part of the product will come from the war.

`JobDslPluginTest` and `AuthorizeProjectTest` have not been changed (leaving the only users of `mock-security-realm` in this project (there is no test of mock-security-realm here!).  

I am not intending to update these - in both cases the plugins are up for adoption - so if we feel the need to update them I would propose that we delete them instead (in a different PR)

### Submitter checklist


jglick (approved):
Makes sense.

PR #1225, opened 2023-06-06
remove IE and HTMLUnit support

jtnord:
IE is not supported by Jenkins so bringin in the dependency for testing is is sub par.

Additionally HTMLUnit is not supported here, we are testing end-to-end with real browsers.  anything HTMLUnit could be done with RealJenkinsRule and HTMLUnit instead.

### Submitter checklist


pom.xml
```diff
@@ -160,6 +160,11 @@
       <!--  to match upper bounds -->
       <version>1.2</version>
     </dependency>
+    <dependency>
```
jtnord, line 163: used by [`FTPdContainer`](https://github.com/jenkinsci/acceptance-test-harness/blob/0bbafeca555b1485383705577ec4c8b38ec28bc5/src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/FtpdContainer.java#L3-L5) was previously transitive

src/main/java/org/jenkinsci/test/acceptance/FallbackConfig.java
```diff
@@ -94,11 +92,6 @@ public class FallbackConfig extends AbstractModule {
     /** Browser property to set the default locale. */
     private static final String LANGUAGE_SELECTOR = "intl.accept_languages";
 
-    /**
-     * PhantomJS browser property to set the default locale.
-     */
-    private static final String LANGUAGE_SELECTOR_PHANTOMJS = "phantomjs.page.customHeaders.Accept-Language";
```
jtnord, line 100: phantomjs was previously removed but this was missed.

jglick (approved):
My reservation about removing HtmlUnit support is that this was the sole option for reproducing test failures in a headless environment without using something like Xvfb. Not sure if it even worked with current versions of core and all the refreshed JavaScript, though.

jtnord:
Chrome (and Firefox?) have headless versions now should we need it.

timja (approved):

pom.xml
```diff
@@ -160,6 +160,10 @@
...
+      <groupId>commons-net</groupId>
+      <artifactId>commons-net</artifactId>
+      <version>3.9.0</version>
```
jtnord, line 166: </dependency>

jtnord, line 166:
```suggestion
      <version>3.9.0</version>
    </dependency>
```

PR #1273, opened 2023-07-05
chore(Jenkinsfile) discard old builds (5/PR and 10 on primary branch)

dduportal:
Each build weight ~100 Mb (mainly due to junit-attachment which is 80 to 90 Mb per build).

This PR proposes to discard old builds: no more than 10 builds on the principal branch (~1 month during the past 6 months) instead of 50 and no more than 5 builds per PR.

The goal is to decrease the amount of storage used on ci.jenkins.io and the I/O pressure.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,11 +1,14 @@
 // For ci.jenkins.io
 // https://github.com/jenkins-infra/documentation/blob/master/ci.adoc
 
-properties([disableConcurrentBuilds(abortPrevious: true)])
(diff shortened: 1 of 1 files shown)


dduportal:
Cancelling the build as this change does not need to run a full ATH.

- Jenkinsfile was parsed properly (no error introduced by my change).
- The build discarder is applied in this PR as expected, with a value  of 5:

Capture d’écran 2023-07-05 à 16 05 48

PR #1324, opened 2023-08-12
Test on Java 21

### LINKED ISSUE #1320: Test on Java 21
As part of our goal of delivering Java 21 support, add ATH test coverage for Java 21. To preserve test coverage for Java 17, we could test weeklies on Java 21 and LTS lines on Java 17, but I do not feel strongly about this.

basil:
Fixes #1320. I have retained a Java 11 build for the LTS line in order to alleviate concerns that we may accidentally begin to start relying on the behavior of newer Java versions in plugins. If this occurs, then shortly after the plugin is released its Java 11 tests would begin to fail on the LTS line (assuming the plugin does not require a weekly release, which is only the case for a small number of plugins). True, the plugin would have to be released to users before the issue is detected in this repository, but that is already a problem even today since there is no convenient way to run ATH or PCT from within a plugin PR build.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -41,13 +41,13 @@ if (needSplittingFromWorkspace) {
 def axes = [
   jenkinsVersions: ['lts', 'latest'],
   platforms: ['linux'],
-  jdks: ['17'],
+  jdks: [11, 21],
   browsers: ['firefox'],
 ]
 
 stage('Record builds and sessions') {
   retry(conditions: [kubernetesAgent(handleNonKubernetes: true), nonresumable()], count: 2) {
-    node('maven-17') {
+    node('maven-21') {
       infra.checkoutSCM()
       def athCommit = sh(script: 'git rev-parse HEAD', returnStdout: true).trim()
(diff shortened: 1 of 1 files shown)


timja:
Looks like adding `JENKINS_OPTS=--enable-future-java` is needed.

basil:
Note that this PR is not yet ready for review.

basil:
Depends on https://github.com/jenkinsci/jenkins/pull/8365 and https://github.com/jglick/simple-maven-project-with-tests/pull/34.

NotMyFault:
> Depends on [jenkinsci/jenkins#8365](https://github.com/jenkinsci/jenkins/pull/8365) and [jglick/simple-maven-project-with-tests#34](https://github.com/jglick/simple-maven-project-with-tests/pull/34).

If we just need a repository with tests, we could replace Jesse's repo in favor of https://github.com/jenkinsci/hello-world-maven-builder, which is already used in another test case.
I created that repo under the org for ease of accessibility when updating a test case to support Java 17.

basil:
@NotMyFault Note that this PR is not yet ready for review.

timja (approved):

PR #1327, opened 2023-08-13
Update maven in tests

NotMyFault:
Update maven to 3.9.4 in test cases, matching the maven version used in the dockerfile


timja (approved):

jtnord (commented):
This causes regressions or flaky tests depending on your environment.

As the version is not in the [installed](https://github.com/jenkinsci/acceptance-test-harness/blob/8ae68f6f5dbc767d560f0c5f10b2e09be2a22f33/src/main/java/org/jenkinsci/test/acceptance/controller/JenkinsController.java#L78-L85) [metadata](https://github.com/jenkinsci/acceptance-test-harness/blob/8ae68f6f5dbc767d560f0c5f10b2e09be2a22f33/src/main/tool_installers/updates/hudson.tasks.Maven.MavenInstaller) - it relies on a successful update having being retrieved from the update center at the time it was done.

PR #1331, opened 2023-08-14
use custom DNS in the ath container to enable HAR recording

jtnord:


src/main/resources/ath-container/run.sh
```diff
@@ -80,4 +80,4 @@ fi
 
 set -x
 
-BROWSER=$browser JENKINS_WAR=$war $MVN $extra_args "$@"
+BROWSER=$browser JENKINS_LOCAL_HOSTNAME=jenkins.127.0.0.1.sslip.io JENKINS_WAR=$war $MVN $extra_args "$@"
```
timja, line 83: should this be configurable?

timja (commented):
please file PRs for any consumers, e.g.

https://github.com/jenkinsci/jenkins/blob/master/ath.sh#L29

jtnord:
> please file PRs for any consumers, e.g.
> 
> https://github.com/jenkinsci/jenkins/blob/master/ath.sh#L29

do you know of any more consumers than this https://github.com/search?q=org%3Ajenkinsci%20jenkins%2Fath&type=code shows the only otehr consumer of the container as the xvnc plugin - but that does not use `run.sh` so will be unaffected.

src/main/resources/ath-container/run.sh
jtnord: we could do it a different way - not modify the files and just start docker with the environment variable and the host I guess.

timja:
not that I know of

Issue #1322, opened 2023-08-11
ATH container is broken on AArch64


timja:
Fails to build on arm64:

```
➜  acceptance-test-harness git:(openjdk21) docker build -t ath src/main/resources/ath-container
[+] Building 55.2s (9/27)
 => [internal] load build definition from Dockerfile                                                                                                                                                 0.0s
 => => transferring dockerfile: 6.47kB                                                                                                                                                               0.0s
 => [internal] load .dockerignore                                                                                                                                                                    0.0s
 => => transferring context: 2B                                                                                                                                                                      0.0s
 => [internal] load metadata for docker.io/library/ubuntu:22.04                                                                                                                                      0.0s
 => [ 1/23] FROM docker.io/library/ubuntu:22.04                                                                                                                                                      0.0s
 => [internal] load build context                                                                                                                                                                    0.0s
 => => transferring context: 3.92kB                                                                                                                                                                  0.0s
 => [ 2/23] RUN apt-get update &&   apt-get install -y --no-install-recommends     gpg-agent     software-properties-common     &&   apt-get clean all && rm -rf /var/cache/apt                     15.0s
 => [ 3/23] RUN apt-get update &&     DEBIAN_FRONTEND=noninteractive     TZ=Etc/UTC     apt-get install -y --no-install-recommends         ca-certificates         gnupg         lsb-release        30.9s
 => [ 4/23] RUN curl -sSLo /tmp/jdk21.tar.gz "https://github.com/adoptium/temurin21-binaries/releases/download/jdk21-${JDK21_VERSION}/OpenJDK21U-jdk_x64_linux_hotspot_${JDK21_VERSION//-beta/}.tar  9.0s
 => ERROR [ 5/23] RUN update-alternatives --install /usr/bin/java java /usr/lib/jvm/java-21-openjdk-amd64/bin/java 21   && update-alternatives --install /usr/bin/javac javac /usr/lib/jvm/java-21-  0.2s
------
 > [ 5/23] RUN update-alternatives --install /usr/bin/java java /usr/lib/jvm/java-21-openjdk-amd64/bin/java 21   && update-alternatives --install /usr/bin/javac javac /usr/lib/jvm/java-21-openjdk-amd64/bin/javac 21   && update-alternatives --set java /usr/lib/jvm/java-17-openjdk-amd64/bin/java   && update-alternatives --set javac /usr/lib/jvm/java-17-openjdk-amd64/bin/javac:
[...]

PR #1321, opened 2023-08-11
Install OpenJDK 21 in ATH image

basil:
Prerequisite for #1320. Without changing the existing configuration (Java 17 as default, Java 11 available), add OpenJDK 21 to the image.

### Testing done

Verified OpenJDK 17 was still the default. Verified I could switch to OpenJDK 21 with `set-java.sh` and ran `java -version` successfully.

PR #1334, opened 2023-08-15
Support Java 21 on AArch64

basil:
Without fixing #1322, restore the status quo as of before #1321 by allowing the image to build (but not run) on AArch64.

### Testing done

Built the image on both amd64 and AArch64.


timja (approved):
Thanks!

PR #1335, opened 2023-08-16
Gitlab container and tests

ampuscas:
### Testing done

```[tasklist]
### Submitter checklist
```


src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer/Dockerfile
```diff
@@ -0,0 +1,5 @@
+# Use the official GitLab Community Edition image as the base
+FROM gitlab/gitlab-ce:latest
```
timja, line 2: pin to a version

julieheard: Fixed, pinned to gitlab-ce:16.2.4-ce.0 and renovate set up to automate pr bumps to newer versions

pom.xml
```diff
@@ -256,6 +256,11 @@
       <artifactId>wordnet-random-name</artifactId>
       <version>1.5</version>
     </dependency>
+    <dependency>
+      <groupId>com.squareup.okhttp3</groupId>
+      <artifactId>okhttp</artifactId>
+      <version>5.0.0-alpha.11</version>
```
timja, line 262: is there a reason you aren't using the built-in Java HttpClient? 
https://www.baeldung.com/java-9-http-client

this avoids adding another dependency, either way probably shouldn't be using an alpha one

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -0,0 +1,94 @@
...
+        RequestBody body = RequestBody.create(mediaType, "{ \"name\": \"" + repoName + "\" }");
+        Request request = new Request.Builder().url("http://" + getIpAddress()+"/api/v4/projects").post(body).addHeader("Content-Type", "application/json").addHeader("PRIVATE-TOKEN", token).build();
+        return sendRequest(request);
+    }
+
+    private Response sendRequest(Request request){
+        Response response = null;
+        try {
+            response = client.newCall(request).execute();
+        } catch (IOException e) {
```
raul-arabaolaza, line 69: Why this mapping?

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -0,0 +1,94 @@
...
+            throw new RuntimeException(e);
+        }
+        return response;
+    }
+
+    public void deleteRepo(String token) throws IOException {
```
raul-arabaolaza, line 75:
```suggestion
    public void deleteRepo(String repo) throws IOException {
```

You do not need the token as there is a getter for it, but you should not assume there is only one repo...

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -0,0 +1,94 @@
...
+        // get the project id and delete the project
+        GitLabApi gitlabapi = new GitLabApi("http://" + getIpAddress(), token);
+        ProjectApi projApi = new ProjectApi(gitlabapi);
+
+        try {
+            Project project = projApi.getProjects().get(0);
```
raul-arabaolaza, line 81: What if more than one test is running in parallel? There will be more than one repo

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -0,0 +1,94 @@
...
+            projApi.deleteProject(project.getId());
+            assertTrue(projApi.getProjects().size()==0);
+        } catch (GitLabApiException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public String createUserToken() throws IOException, InterruptedException {
```
raul-arabaolaza, line 89:
```suggestion
    public String createUser(String username) throws IOException, InterruptedException {
```

raul-arabaolaza: This should allow you to create any user you want and return the token for further auth if needed

src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer/create_user.rb
```diff
@@ -0,0 +1,13 @@
+user = User.create();
+user.name = 'test';
```
raul-arabaolaza, line 2: Hardcoded, what if this is called by tests in parallel? Is gonna fail. Maybe you can try what is described here https://www.codecademy.com/article/ruby-command-line-argv

src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer/create_user.rb
```diff
@@ -0,0 +1,13 @@
...
+user.username = 'test';
+user.password = 'oparolaoarecare12!';
```
raul-arabaolaza, line 4: Same as before, this should be an argument comming from the java call

src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer/create_user.rb
```diff
@@ -0,0 +1,13 @@
...
+user.confirmed_at = '01/01/1990';
+user.admin = true;
+user.email = '$test@example.com';
```
raul-arabaolaza, line 7: Ditto

raul-arabaolaza (commented):
Leave several comments

.github/renovate.json
```diff
@@ -22,9 +22,10 @@
       "description": "We focus on Guice 6 until core adapts 7"
     },
     {
-      "matchPackagePatterns": [
-        "selenium"
-      ],
+      "matchPackageNames": ["gitlab/gitlab-ce"]
```
timja, line 25: this will be hidden by `ignorePaths` on line 9, dependabot appears to be used for docker updates:
https://github.com/jenkinsci/acceptance-test-harness/blob/master/.github/dependabot.yml

it would probably be better to remove all the dependabot docker config though and just switch it to renovate as then we wouldn't need to explicitly list it

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
julieheard: What do you mean by this?

pom.xml
```diff
@@ -55,6 +55,16 @@
   </pluginRepositories>
 
   <dependencies>
+    <dependency>
+      <groupId>com.squareup.okhttp3</groupId>
+      <artifactId>okhttp</artifactId>
+      <version>5.0.0-alpha.11</version>
+    </dependency>
```
julieheard, line 62: Ignore that this is in here twice - we have Anda and I working on the same PR so just caught in the crossfire. We are removing as moving to http client shortly so ignore for now 🙂

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
julieheard: Comment resolved due to switching over to java http client

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    public void deleteRepo(String token) throws IOException {
+        // get the project id and delete the project
+        GitLabApi gitlabapi = new GitLabApi("http://" + getIpAddress(), token);
+        ProjectApi projApi = new ProjectApi(gitlabapi);
+        try {
+            Project project = projApi.getProjects().get(0);
+            projApi.deleteProject(project.getId());
+        } catch (GitLabApiException e) {
```
raul-arabaolaza, line 88: No need to wrap inside a `RuntimeException` just throw it and the test will fail if thrown

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -0,0 +1,116 @@
...
+            throw new RuntimeException(e);
+        }
+    }
+
+    public void waitForReady(CapybaraPortingLayer p) {
+        p.waitFor().withMessage("Waiting for GitLab to come up")
+                .withTimeout(Duration.ofSeconds(200)) // GitLab starts in about 2 minutes
```
raul-arabaolaza, line 95: Use `ElasticTime`

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -0,0 +1,116 @@
...
+                .until( () ->  {
+                    try {
+                          HttpRequest request = HttpRequest.newBuilder()
+                                .uri(new URI("http://" + getIpAddress()))
+                                .GET()
+                                .build();
+                        HttpResponse<String> response = client.send(request, HttpResponse.BodyHandlers.ofString());
+                        return response.body().contains("GitLab Community Edition");
+                    } catch (SocketException e) {
+                        return null;
```
raul-arabaolaza, line 105:
```suggestion
                        return false;
```

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -0,0 +1,78 @@
...
+    public void init() {
+        container = gitLabServer.get();
+        repoUrl = container.getRepoUrl();
+        host = container.host();
+        port = container.port();
+        container.waitForReady(this);
+        
+        try {
+            privateToken = container.createUserToken();
+        } catch (InterruptedException e) {
```
raul-arabaolaza, line 47: Do not convert to `RuntimeExceptions, just throw it and the test will fail if needed

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -0,0 +1,115 @@
...
+     * @param alias an alias for this container’s {@link #getCid} passed to {@code --link}
+     */
+    public String getRepoUrlInsideDocker(String alias) throws IOException {
+        return "ssh://git@" + alias + REPO_DIR;
+    }
+
+    public HttpResponse<String> createRepo(String repoName, String token) throws IOException {
+        try{
+            HttpRequest request = HttpRequest.newBuilder()
+                                             .uri(new URI("http://" + getIpAddress() + "/api/v4/projects"))
```
jtnord, line 70: I think this will not work in all scenarious - the web port has not been configured as exposed (`@DockerFixture(id = "gitlab-plugin", ports = 22)`) and as such you may not be able to communicate witht the web interface of GitLab depending on your docker setup.  IIUC you should add a new mapping for the web port.

Additionally you have `getURL()` that you should call to then addon the endpoint rather than repeating yourself.

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -0,0 +1,115 @@
...
+
+    private static final ElasticTime time = new ElasticTime();
+
+    public String host() {
+        return ipBound(22);
+    }
+
+    public int port() {
+        return port(22);
+    }
```
jtnord, line 42: there should be multiple ports - the SSH port and the web port.

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -0,0 +1,115 @@
...
+
+    public URL getURL() throws IOException {
+        // return new URL("http://" + host() + ":" + port());
+        return new URL("http://" + getIpAddress());
```
jtnord, line 46: you need to use the ipaddress and port that is mapped here.

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -0,0 +1,115 @@
...
+            HttpResponse<String> response = client.send(request, HttpResponse.BodyHandlers.ofString());
+            return response;
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public void deleteRepo(String token, String repoName) throws IOException, GitLabApiException {
+        // get the project and delete the project
+        GitLabApi gitlabapi = new GitLabApi("http://" + getIpAddress(), token);
```
jtnord, line 84: use `getURL()` to encapsulate the URL for GitLab

src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer/create_user.rb
jtnord: > what if this is called by tests in parallel

then each test should have its own instance of the container so it doesn't matter (unless we are running multiple tests in the same JVM at once which we are not (threading as opposed to forking))

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    public void waitForReady(CapybaraPortingLayer p) {
+        int timeout = (int) time.milliseconds(200000);
+        p.waitFor().withMessage("Waiting for GitLab to come up")
+                .withTimeout(Duration.ofSeconds(timeout)) // GitLab starts in about 2 minutes
+                .until( () ->  {
+                    try {
+                          HttpRequest request = HttpRequest.newBuilder()
+                                .uri(new URI("http://" + getIpAddress()))
```
jtnord, line 98: use `getURL()`

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -0,0 +1,115 @@
...
+        ProjectApi projApi = new ProjectApi(gitlabapi);
+
+        Project project = projApi.getProjects().stream().filter((proj -> repoName.equals(proj.getName()))).findAny().orElse(null);
+        projApi.deleteProject(project);
+    }
+
+    public void waitForReady(CapybaraPortingLayer p) {
+        int timeout = (int) time.milliseconds(200000);
+        p.waitFor().withMessage("Waiting for GitLab to come up")
+                .withTimeout(Duration.ofSeconds(timeout)) // GitLab starts in about 2 minutes
```
jtnord, line 94: would be so much more readable with something like

```suggestion
                .withTimeout(Duration.ofSeconds(60 * 2)) // GitLab starts in about 2 minutes
```
or 
suggestion
                .withTimeout(Duration.ofMinutes(2)) // GitLab starts in about 2 minutes
```

**Review by jtnord — COMMENTED** [review]
Should this be in draft until there is a test that exercises some functionality in Jenkins provided by the GitLab plugin?

File: .github/renovate.json (hunk 10 above)
Comment (julieheard) [reply .github/renovate.json:25]: @ampuscas This renovate bit can be reverted now - the plan is to use dependabot for this PR and then make another PR to deal with switching from Dependabot to Renovate 🙂

File: src/main/java/org/jenkinsci/test/acceptance/plugins/gitlab_plugin/GitLabServerConfig.java [hunk 22]
```diff
@@ -0,0 +1,31 @@
...
+    public String configureServer(String url, String token) {
+        control("name").set("servername");
+        control("url").set(url);
+        control(By.cssSelector("button.credentials-add-menu")).click();
+        control(By.cssSelector("li.yuimenuitem")).click();
+        find(by.option("GitLab API token")).click();
+        find(by.path("/credentials/apiToken")).sendKeys(token);
+
+        find(by.id("credentials-add-submit-button")).click();
+        sleep(2000);
```
rsandell, line 24: Sleeping is normally prone to flakyness, instead you should try to wait for some condition on the page instead.

rsandell: For example https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/test/java/plugins/ArtifactoryPluginTest.java#L56

src/main/java/org/jenkinsci/test/acceptance/plugins/gitlab_plugin/GitLabServerConfig.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class GitLabServerConfig extends PageAreaImpl {
+
+    @Inject
+    public GitLabServerConfig(Jenkins jenkins) {
+        super(jenkins, "/com-dabsquared-gitlabjenkins-connection-GitLabConnectionConfig/connections");
+    }
+
+    public String configureServer(String url, String token) {
+        control("name").set("servername");
```
raul-arabaolaza, line 16: Use `by.input` or `by.path` if possible see https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/main/java/org/jenkinsci/test/acceptance/ByFactory.java they are usually more resistant to changes.

src/main/java/org/jenkinsci/test/acceptance/plugins/gitlab_plugin/GitLabServerConfig.java
```diff
@@ -0,0 +1,31 @@
...
+        control("url").set(url);
+        control(By.cssSelector("button.credentials-add-menu")).click();
```
raul-arabaolaza, line 18: Using cssSelector is usually fragile... are there no other options like `by.id` or `by.path` available?

Also, please use `by` wrapper instead of `By` native. The wrapper contains extra logic tailored to jenkins UI

src/main/java/org/jenkinsci/test/acceptance/plugins/gitlab_plugin/GitLabServerConfig.java
```diff
@@ -0,0 +1,26 @@
...
+
+    @Inject
+    public GitLabServerConfig(Jenkins jenkins) {
+        super(jenkins, "/io-jenkins-plugins-gitlabserverconfig-servers-GitLabServers");
+    }
+
+    public void configureServer(String url) {
+        find(by.path("/io-jenkins-plugins-gitlabserverconfig-servers-GitLabServers/servers/name")).sendKeys("servername");
+        find(by.path("/io-jenkins-plugins-gitlabserverconfig-servers-GitLabServers/servers/serverUrl")).clear();
+        find(by.path("/io-jenkins-plugins-gitlabserverconfig-servers-GitLabServers/servers/serverUrl")).sendKeys(url);
```
jtnord, line 18: you shouldn't need to be using the full form here - things should be relative to the context (which is passed in the contructor) so  
`find(by.path("servers/name")).sendKeys("servername");`

but also - these should likely be controlles.  - is this a repeatable option - can it occur more than once?

src/main/java/org/jenkinsci/test/acceptance/plugins/gitlab_plugin/GitLabServerConfig.java
```diff
@@ -0,0 +1,26 @@
...
+
+        find(by.option("GitLab Personal Access Token")).click();
```
jtnord, line 20: this presumably causes some dynamic javascript - so you should probably wait to ensure that has completed before you move on?   
do you not need to do anything here - like specifiy the credential?

src/main/java/org/jenkinsci/test/acceptance/plugins/gitlab_plugin/GitLabServerConfig.java
```diff
@@ -0,0 +1,26 @@
...
+
+        find(by.path("/io-jenkins-plugins-gitlabserverconfig-servers-GitLabServers/servers/validate-button")).click();
+
+        waitFor(driver, hasContent("Credentials verified for user"), 2);
```
jtnord, line 24: this should come in a validation area?  in which case - checking the validation area for the form over the content of HTML would be preferred.  see [`FormValidation`](https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/main/java/org/jenkinsci/test/acceptance/po/FormValidation.java) and if you use control objects then you can get this from the control.

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -0,0 +1,133 @@
...
+    DockerContainerHolder<GitLabContainer> gitLabServer;
+
+    private GitLabContainer container;
+    private String repoUrl;
+    private String host;
+    private int port;
+
+    private String privateToken;
+
+    private String password = "arandompassword12#"; // I guess the password can be the same for all users
```
jtnord, line 40: ideally you would want it to be different to ensure that things do work as expected.

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -0,0 +1,133 @@
...
+    @Before
+    public void init() throws InterruptedException, IOException {
+        container = gitLabServer.get();
+        repoUrl = container.getRepoUrl();
+        host = container.host();
+        port = container.port();
+        container.waitForReady(this);
+
+        // create an admin user
+        privateToken = container.createUserToken(adminUserName, password, "testadmin@gmail.com", "true");
```
jtnord, line 61: this email may actually exist and I would hate for us to accidently at any point send emails to them - can you use `@example.com` or `@ath.jenkins.test` here as [they are reserved](https://datatracker.ietf.org/doc/html/rfc2606) and would never be valid?

src/main/java/org/jenkinsci/test/acceptance/plugins/gitlab_plugin/GitLabServerConfig.java
ampuscas: I put the whole path because it doesn’t find the element if I use `find(by.path("servers/name")).sendKeys("servername");`
I just tried again and it’s the same error I had before
`Method threw 'org.openqa.selenium.NoSuchElementException' exception.`

src/main/java/org/jenkinsci/test/acceptance/plugins/gitlab_plugin/GitLabServerConfig.java
ampuscas: modification coming soon for waiting 
>do you not need to do anything here - like specifiy the credential?

No, there is only one type (GitLab API token) that can be put

src/main/java/org/jenkinsci/test/acceptance/plugins/gitlab_plugin/GitLabServerConfig.java
ampuscas: The FormValidation is for the entire Config page, I didn't find anything particular for server configuration

src/main/java/org/jenkinsci/test/acceptance/plugins/gitlab_plugin/GitLabServerConfig.java
jtnord: because you are using find and not a control (the later will be relative to this PageArea)

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -0,0 +1,172 @@
...
+
+        // create an admin user
+        privateTokenAdmin = container.createUserToken(adminUserName, "arandompassword12#", "testadmin@example.com", "true");
+
+        // create another user
+        privateTokenUser = container.createUserToken(userName, "passwordforsimpleuser12#", "testsimple@example.com", "false");
+    }
+
+    @Test
+    public void dummy_test() {
```
car-roll, line 74: nit: Probably don't need this test anymore? Or I guess it can be kept as a smoke test

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -0,0 +1,217 @@
...
+    public int webPort() {
+        return port(443);
+    }
+
+    public String webHost() {
+        return ipBound(443);
+    }
+
+    public URL getURL() throws IOException {
+        return new URL("http://" + getIpAddress() + port());
```
jtnord, line 53: ?  
`port()` is `port(22)` so this is the SSH port - and as such would not be useable by SSH?

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -0,0 +1,217 @@
...
+                                                        .connectTimeout(Duration.ofMillis(200))
+                                                       .build();
+
+    private static final ElasticTime time = new ElasticTime();
+
+    public String host() {
+        return ipBound(22);
+    }
+
+    public int port() {
```
jtnord, line 39: `port` is an ambiguous given you have 3 ports exposed.  consider renaming to `sshPort()` or at least adding javadoc so it is obvious what this is!

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -0,0 +1,217 @@
...
+        return port(22);
+    }
+
+
+    public int webPort() {
+        return port(443);
+    }
+
+    public String webHost() {
+        return ipBound(443);
```
jtnord, line 49: is this ever going to be different to `host()` ?

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -0,0 +1,217 @@
...
+        String url = "http://" + ipBound(80) + ':' + port(80);
+        return new URL(url);
+    }
+
+    /** URL visible from the host. */
+    public String getRepoUrl() {
+        return "ssh://git@" + host() + ":" + port() + REPO_DIR;
+    }
+
+    @Deprecated
```
jtnord, line 66: this is new code - why deprecated something that is brand new - just remove the code and fix all callers?

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -0,0 +1,217 @@
...
+        return "ssh://git@" + getIpAddress() + REPO_DIR;
+    }
+
+    /**
+     * URL visible from other Docker containers.
+     * @param alias an alias for this container’s {@link #getCid} passed to {@code --link}
+     */
+    public String getRepoUrlInsideDocker(String alias) throws IOException {
+        return "ssh://git@" + alias + REPO_DIR;
+    }
```
jtnord, line 77: is this used - a search found no uses of this or the deprecated `getRepoUrlInsideDocker()`

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -0,0 +1,217 @@
...
+                                             .uri(new URI(getHttpUrl() + "/api/v4/projects"))
+                                             .header("Content-Type", "application/json")
+                                             .header("PRIVATE-TOKEN", token)
+                                             .POST(HttpRequest.BodyPublishers.ofString("{ \"name\": \"" + repoName + "\" }"))
+                                             .build();
+            HttpResponse<String> response = client.send(request, HttpResponse.BodyHandlers.ofString());
+            return response;
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
```
jtnord, line 113: the method declares IOException yet you catch everything and throw a `RuntimeException`?

just propagate the original exception - let it buble up and out of the test cases?

src/main/java/org/jenkinsci/test/acceptance/plugins/gitlab_plugin/GitLabOrganizationFolder.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Override
+    public URL getConfigUrl() {
+        return null;
+    }
+
+    public String getCheckLog() {
+        try {
+            return IOUtils.toString(url("computation/console").openStream(), StandardCharsets.UTF_8);
```
jtnord, line 31: this should ideally use webdriver so that you can see what is happening when it happens.  Additionally, if you have enabled security this would fail as it is not passing any authentication which would be in the session for the WebDriver

pom.xml
```diff
@@ -55,6 +55,11 @@
   </pluginRepositories>
 
   <dependencies>
+    <dependency>
```
raul-arabaolaza, line 58: NIT: I wonder if it would be better to depend on the gitlab-api plugin and get the gitlab4j library transitively so you are using (and in a limited way also testing) the library the gitlab-branch-source plugin is using internally.

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
ampuscas: deleted it

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
ampuscas: deleted

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    public void createBranch(String token, String repoName) throws IOException, GitLabApiException {
+        GitLabApi gitlabapi = new GitLabApi("http://" + getIpAddress(), token);
+        ProjectApi projApi = new ProjectApi(gitlabapi);
+        Project project = projApi.getProjects().stream().filter((proj -> repoName.equals(proj.getName()))).findAny().orElse(null);
+
+        RepositoryFile file = new RepositoryFile();
```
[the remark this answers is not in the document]
ampuscas: It is specific to the test, but it is created inside the container where GitLab server runs. Do you think I should move it some place else? For me it made sense to put it here as it's related to the container.

src/test/java/plugins/GitLabPluginTest.java
ampuscas: I was also thinking to delete this test, just didn't get to it, I am deleting it now

src/main/java/org/jenkinsci/test/acceptance/plugins/gitlab_plugin/GitLabPersonalAccessTokenCredential.java
```diff
@@ -0,0 +1,29 @@
...
+    public GitLabPersonalAccessTokenCredential(PageObject context, String path) {
+        super(context, path);
+    }
+
+    public GitLabPersonalAccessTokenCredential(PageAreaImpl area, String relativePath) {
+        super(area, relativePath);
+    }
+
+    public void setToken(String token) {
+        control(by.path("/credentials/token")).set(token);
```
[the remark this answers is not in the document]
ampuscas: I don't quite understand this, I used a lot `control` should I refactor everything? :( 
And then, the next comment suggests I shoud use `control` instead of `find`? https://github.com/jenkinsci/acceptance-test-harness/pull/1335/files#r1314662604

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -0,0 +1,217 @@
...
+        return ipBound(22);
+    }
+
+    public int port() {
+        return port(22);
+    }
+
+
+    public int webPort() {
+        return port(443);
```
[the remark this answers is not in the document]
ampuscas: modified

src/main/java/org/jenkinsci/test/acceptance/plugins/gitlab_plugin/GitLabPersonalAccessTokenCredential.java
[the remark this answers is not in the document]
jtnord: > I don't quite understand this, I

```
class ... {
  private Control token = control(by.path("/credentials/token"));
  
  //...,
  public void setToken(String gitLabToken) {
     token.set(gitLabToken);
   }
   
}

File: pom.xml (hunk 38 above)
Comment (jtnord) [reply pom.xml:58]: I would say not

the library may be updated to work with newer GitLab servers (which we update here) where it is not updated in the plugin and as such we would want to know the plugin is broken when GL is updated rather than the test errors for some issue that is more obscure.

There is no requirement to use the same version of the library or even the same library when interacting with any system.

Comment (ampuscas) [reply pom.xml:58]: If I remove the `gitlab4j` dependency and add `gitlab-api` there are several issues. Is it worth updating these dependencies?
```[ERROR] Rule 4: org.apache.maven.enforcer.rules.dependency.RequireUpperBoundDeps failed with message:
[ERROR] Failed while enforcing RequireUpperBoundDeps. The error(s) are [
[ERROR] Require upper bound dependencies error for org.javassist:javassist:3.27.0-GA paths to dependency are:
[ERROR] +-org.jenkins-ci:acceptance-test-harness:999999-SNAPSHOT
[ERROR]   +-com.browserup:browserup-proxy-core:2.1.2
[ERROR]     +-org.javassist:javassist:3.27.0-GA
[ERROR] and
[ERROR] +-org.jenkins-ci:acceptance-test-harness:999999-SNAPSHOT
[ERROR]   +-io.jenkins.plugins:gitlab-api:5.3.0-91.v1f9a_fda_d654f
[ERROR]     +-io.jenkins.plugins:jersey2-api:2.39.1-1
[ERROR]       +-org.glassfish.jersey.inject:jersey-hk2:2.39.1
[ERROR]         +-org.javassist:javassist:3.29.0-GA

Comment (jtnord) [reply pom.xml:58]: just don't depend on the plugin and keep it as is.

Comment (ampuscas) [reply pom.xml:58]: ah, sorry, I didn't refresh the page and I didn't see your previous comment before posting mine.

File: src/main/java/org/jenkinsci/test/acceptance/plugins/gitlab_plugin/GitLabServerConfig.java (hunk 25 above)
Comment (ampuscas) [reply src/main/java/org/jenkinsci/test/acceptance/plugins/gitlab_plugin/GitLabServerConfig.java:18]: It still doesn't work if I don't put the whole path, it doesn't find the element. I modified to use `Control` but with the full path...

File: src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java (hunk 39 above)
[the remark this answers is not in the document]
Comment (jtnord) [reply src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java:108]: To me it is part of the setup of the test - not setup of the GitLab server so would be with the test.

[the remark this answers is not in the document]
Comment (jtnord) [reply src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java:108]: so if this was more generic like `deleteRepo()` I would say it should live here - but this is too tightly coupled to the test as is.

in some ways - it does more that is is expected to as well - it is not creating a branch (and the name of the branch should be an option provided by the called).  it is creating a default (`main` branch with some specific content) as well as 2 other branches).

And then it is creating a merge request as well.

Could you possibly split this between the test that requires the specific setup, and the fixture that abstracts away the GitLab parts?  (e.g.  `createBranch(String reponame, String branchName, String sourceBranch)`    `addFile(String reponame, String branchName, String filePath, String content)`  `createMergeRequest(String project, String sourceBranch, String targetBranch, String mrTitle)` )

File: src/main/java/org/jenkinsci/test/acceptance/plugins/gitlab_plugin/GitLabServerConfig.java [hunk 42]
```diff
@@ -0,0 +1,30 @@
...
+import org.jenkinsci.test.acceptance.po.Control;
+import org.jenkinsci.test.acceptance.po.Jenkins;
+import org.jenkinsci.test.acceptance.po.PageAreaImpl;
+
+import static org.jenkinsci.test.acceptance.Matchers.hasContent;
+
+public class GitLabServerConfig extends PageAreaImpl {
+
+    private Control serverName = control(by.path("/io-jenkins-plugins-gitlabserverconfig-servers-GitLabServers/servers/name"));
+    private Control serverUrl = control(by.path("/io-jenkins-plugins-gitlabserverconfig-servers-GitLabServers/servers/serverUrl"));
```
jtnord, line 13: use the pagearea to obtain the relative path?

```suggestion
    private Control serverName = control("servers", "name"));
    private Control serverUrl = control("servers", "serverUrl"));
```

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+    @Before
+    public void init() throws InterruptedException, IOException {
+        container = gitLabServer.get();
+        host = container.host();
+        port = container.sshPort();
+        container.waitForReady(this);
+
+        // create an admin user
+        privateTokenAdmin = container.createUserToken(adminUserName, "arandompassword12#", "testadmin@example.com", "true");
```
jtnord, line 69: is there a risk that this email address can gets any emails or anything?

possibly better to use `testadmin@invalid.test`

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+        // create another user
+        privateTokenUser = container.createUserToken(userName, "passwordforsimpleuser12#", "testsimple@example.com", "false");
```
jtnord, line 72: see above

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -0,0 +1,204 @@
...
+        return new URL(url);
+    }
+
+    /** URL visible from the host. */
+    public String getRepoUrl() {
+        return "ssh://git@" + host() + ":" + sshPort() + REPO_DIR;
+    }
+
+    public void waitForReady(CapybaraPortingLayer p) {
+        long timeout =  time.seconds(200); // GitLab starts in about 2 minutes add some headway
```
jtnord, line 67: this is not actually correct when fed into `waitFor`  as waitFor() creates a `Wait` with an `ElasticTime` as its clock.
so if the factor was 1.5 and you waited 1 second you would end up waiting 2.25 seconds not 1.5 as expected.

https://github.com/jenkinsci/acceptance-test-harness/blob/3bef84db279305367c329dbb06624ac60a6ef3bb/src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java#L117-L121

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -0,0 +1,204 @@
...
+        p.waitFor().withMessage("Waiting for GitLab to come up")
+                .withTimeout(Duration.ofMillis(timeout))
```
jtnord, line 69: see above - just use `.withTimeout(Duration.ofSeconds(200))`

src/main/java/org/jenkinsci/test/acceptance/plugins/gitlab_plugin/GitLabOrganizationFolder.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public String getCheckLog() {
+        return driver.getPageSource();
+
+    }
+
+    public GitLabOrganizationFolder waitForCheckFinished(final int timeout) {
+        waitFor()
+                .withTimeout(Duration.ofMillis(super.time.seconds(timeout)))
```
jtnord, line 36: (not sure of the timeout unit here - should be documented as this is public...)

```suggestion
                .withTimeout(Duration.ofSeconds(timeout))
```

jtnord (commented):
some incorect use of Elastic time with wait.

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    private void configureJobWithGitLabBranchSource(final WorkflowMultiBranchJob job, String adminUserName, String project) {
+        final GitLabBranchSource ghBranchSource = job.addBranchSource(GitLabBranchSource.class);
+        ghBranchSource.setOwner(adminUserName);
+        ghBranchSource.setProject(adminUserName, project);
+    }
+
+    private void assertExistAndResult(final WorkflowJob job, final boolean withSuccess) {
+        final Build.Result expectedResult = (withSuccess) ? Build.Result.SUCCESS : Build.Result.FAILURE;
```
jtnord, line 146: I think you need to wait for the build to have finished here don't you?  it may be in progress or in the queue having not yet started?

car-roll (approved):

src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer/create_user.rb
```diff
@@ -0,0 +1,15 @@
...
+user.name = input_array[0];
+user.username = input_array[0];
+user.password = input_array[1];
+user.confirmed_at = '01/01/1990';
+user.admin = input_array[3];
+user.email = input_array[2];
+user.save!;
+
+token = user.personal_access_tokens.create(scopes: [:api], name: 'MyToken');
+token.expires_at='01/01/2024';
```
rsandell, line 13: This will be problematic in 4 months.

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -0,0 +1,323 @@
...
+        final WorkflowJob successJob3 = multibranchJob.getJob("MR-1");
+        final WorkflowJob failureJob = multibranchJob.getJob(failedJob);
+
+        this.assertExistAndResult(successJob1, true);
+        this.assertExistAndResult(successJob2, true);
+        this.assertExistAndResult(successJob3, true);
+        this.assertExistAndResult(failureJob, false);
+
+        // delete the repo when finished
+        container.deleteRepo(getPrivateTokenAdmin(), repoName);
```
rsandell, line 158: This should be done in a `@After` method or in a `try { ... } finally {...}` block because if any of the asserts above fails then the repo won't be deleted otherwise.

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -0,0 +1,323 @@
...
+
+    private String anotherRepoName = "anotherproject";
+
+    private String adminUserName = "testadmin";
+
+    private String userName = "testsimple";
+
+    private String groupName = "firstgroup";
+
+    private String JenkinsfileContent = "pipeline {\n" +
```
rsandell, line 58: This pipeline is using declarative syntax which should mean you need to declare the `pipeline-model-definition` plugin in the `@WithPlugins` annotation.

rsandell (approved):

raul-arabaolaza (approved):

raul-arabaolaza (approved):

raul-arabaolaza:
Can we get this merged, please? No blockers apparently and the builder is happy

jtnord:
@rsandell [master is broken](https://ci.jenkins.io/job/Core/job/acceptance-test-harness/job/master/949/testReport/) - looks like this PR is broken on java21?

timja:
Can this please be looked at quickly?

It's breaking all PRs.

car-roll:
This is from my first look at these failures: https://github.com/jenkinsci/acceptance-test-harness/pull/1364#issuecomment-1728101918

car-roll:
Possible fix #1365

PR #1352, opened 2023-09-01
chore(jenkinsfile): use spot instance for the first try or falls back to on-demand

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3673#issuecomment-1660606208

this PR select a spot instance on first try or falls back to on-demand.

spot instance ensure that most of the builds are cheap but we don't want builds to fail due to infra issues.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -110,8 +110,15 @@ for (int i = 0; i < splits.size(); i++) {
     def name = "${jenkinsVersion}-${platform}-jdk${jdk}-${browser}-split${index}"
     branches[name] = {
       stage(name) {
+        int retryCounts = 1
         retry(count: 2, conditions: [agent(), nonresumable()]) {
-          node('docker-highmem') {
+          String nodeLabel = 'docker-highmem-nonspot'
+          if (retryCounts == 1) {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -110,8 +110,15 @@ for (int i = 0; i < splits.size(); i++) {
     def name = "${jenkinsVersion}-${platform}-jdk${jdk}-${browser}-split${index}"
     branches[name] = {
       stage(name) {
+        int retryCounts = 1
         retry(count: 2, conditions: [agent(), nonresumable()]) {
-          node('docker-highmem') {
+          String nodeLabel = 'docker-highmem-nonspot'
+          if (retryCounts == 1) {
+            // use a spot instance for the first try
```
dduportal, line 117:
```suggestion
            // Use a spot instance for the first try
```

dduportal:
Replaying the build as admin to ensure the change is run in the PR check

dduportal:
We replayed the build a second time and kille the high meme agents while running to check the outcome.

- Spot instances are used for the first try during the 2 builds as expected
- The retry switches to an on-demand instance to completion

LGTM!

PR #1394, opened 2023-10-20
fix folders tests

mawinter69:
after switching folders to use a confirmation dialog the tests no longer work for lts where still the browser native dialogs are used.

### Testing done

```[tasklist]
### Submitter checklist
```


jglick (approved):
Not sure if ATH is expected to handle the older version of the plugin as well.

NotMyFault (approved):

timja:
> Not sure if ATH is expected to handle the older version of the plugin as well.

We wouldn't generally do that, unless the newer change isn't available on the active LTS line then we would need to as we test LTS and weekly

PR #1492, opened 2024-02-23
Add a page object for the snippet generator

uhafner:
I use this PO in the ATH tests of the warnings plugin:
https://github.com/jenkinsci/warnings-ng-plugin/blob/5effdb04eaa5dc1025085ac41257e5a6ade4de80/ui-tests/src/test/java/io/jenkins/plugins/analysis/warnings/SnippetGeneratorUiTest.java

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/SnippetGenerator.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/SnippetGenerator.java
@@ -31,10 +31,12 @@ public class SnippetGenerator extends PageObject {
      * @return the generated script
      */
     public String generateScript() {
-        WebElement button = find(By.id("generatePipelineScript-button"));
-        button.click();
-        elasticSleep(500);
-        WebElement textarea = find(By.id("prototypeText"));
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/SnippetGenerator.java
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * Generates the sample pipeline script.
+     *
+     * @return the generated script
+     */
+    public String generateScript() {
+        WebElement button = find(By.id("generatePipelineScript-button"));
+        button.click();
+        elasticSleep(500);
```
jtnord, line 36: please avoid introducing new POs with elastic sleep where they can be avoided.

if the callers are never going to call generate script more than once you can wait `wait.until` the `textarea` is not empty.  (and add this fact to the javadoc)
if it is going to call it more than once with different options then it should be possible to get a before and wait until it is not the same.

timja (approved):

PR #1493, opened 2024-02-23
compatibility for new progressbar rendering

mawinter69:
previously it was `table`, now it is `a` (theoretically it can also be `div` but not for the build history).
See https://github.com/jenkinsci/jenkins/pull/8821

### Testing done

```[tasklist]
### Submitter checklist
```


jtnord (approved):

raul-arabaolaza (approved):

raul-arabaolaza:
None of the test failures here are new and were already falling in previous runs of master. So I believe this will need to be merged manually.

I am checking the test failures just in case I can find an easy fix as the root cause looks limited

raul-arabaolaza:
I found no relationship between this change and the test failures, this can be merged IMO.

PR #1501, opened 2024-03-10
Validate Gradle wrapper

NotMyFault:
I noticed https://github.com/jenkinsci/acceptance-test-harness/commit/865cecc01a41076b69cbc0069a953f78e00f941d and in case someone bumps the wrapper in the future, if needed, the action added validates whether the .jar is legit or not.

### DIFF
--- a/.github/workflows/verify-gradle-wrapper.yaml
+++ b/.github/workflows/verify-gradle-wrapper.yaml
@@ -1,0 +1,11 @@
+name: Validate Gradle wrapper
+on: [pull_request, push]
+jobs:
+  build_pr:
+    if: github.repository_owner == 'jenkinsci'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/verify-gradle-wrapper.yaml
```diff
@@ -0,0 +1,11 @@
+name: Validate Gradle wrapper
+on: [pull_request, push]
+jobs:
+  build_pr:
+    if: github.repository_owner == 'jenkinsci'
```
timja, line 5: Does this work on forks with this if?

NotMyFault: Yes: https://github.com/jenkinsci/acceptance-test-harness/actions/workflows/verify-gradle-wrapper.yaml

The if-clause context refers to the owner of the repository where the workflow is running, not where the PR is coming from.
The job would only be skipped if you file a PR _on_ your fork _towards_ your fork.

jtnord (approved):

PR #1510, opened 2024-03-18
Adapt some tests to make it feasible to run it under IPv6 environments

imonteroperez:
### Highlights
- Follow up from https://github.com/jenkinsci/docker-fixtures/pull/73
- Updates some tests to use the mentioned IPv6 docker-fixtures support as well as some required additional refactoring (i.e. URL#<init> methods IPv6 compliant, etc.)

### Testing done
- Tested under dedicated IPv6 environment and no issues.
- CI errors obtained on the https://ci.jenkins.io/job/Core/job/acceptance-test-harness/job/PR-1510/ seems to be the same than in master branch 

```[tasklist]
### Submitter checklist
```


src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/LdapContainer.java
```diff
@@ -38,7 +38,10 @@ public String getManagerPassword() {
      * @return default ldap connection details from current running docker LdapContainer.
      */
     public LdapDetails createDefault() {
-        return new LdapDetails(getHost(), getPort(), getManagerDn(), getManagerPassword(), getRootDn());
+        return new LdapDetails(addBracketsIfNeeded(getHost()), getPort(), getManagerDn(), getManagerPassword(), getRootDn());
```
jtnord, line 41: this is incorrect.  LDAPDetails is expecting a host `[22:33::12]` is not a host but an encapsulated ip6 host.  

if this is needed them something is using LDAPDetails incorrectly - possibly via `getHostWithPort` which is likely to be incorrect.

src/test/java/plugins/GitPluginTest.java
```diff
@@ -127,10 +127,10 @@ public void name_remote_repo() {
         job.addShellStep("test -f foo && git remote -v");
         job.save();
 
-        job.startBuild().shouldSucceed().shouldContainsConsoleOutput("custom_origin\\s+" + repoUrl);
+        job.startBuild().shouldSucceed().shouldContainsConsoleOutput("custom_origin\\s+" + 
+                repoUrl.replace("[", "\\[").replace("]", "\\]"));
```
jtnord, line 131: use [`Pattern.quote`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/regex/Pattern.html#quote(java.lang.String))  host could contain other characters that are special (ie a `.` should be treated as a fullstop not any character)

src/test/java/plugins/LdapPluginTest.java
```diff
@@ -63,7 +63,7 @@ private LdapDetails createDefaults(LdapContainer ldapContainer) {
      * @return default ldap connection details without the manager credentials
      */
     private LdapDetails createDefaultsWithoutManagerCred(LdapContainer ldapContainer) {
-        return new LdapDetails(ldapContainer.getHost(), ldapContainer.getPort(), "", "", ldapContainer.getRootDn());
+        return new LdapDetails(LdapContainer.addBracketsIfNeeded(ldapContainer.getHost()), ldapContainer.getPort(), "", "", ldapContainer.getRootDn());
```
jtnord, line 66: this appears incorrect.  LDAPDetails is expecting a host, `[22:33::12]` is not a host but an encapsulated ip6 host?

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitContainer.java
```diff
@@ -53,5 +53,8 @@ public void addSSHCertificate(String pubKey) throws IOException, InterruptedExce
                 .popen()
                 .verifyOrDieWith("Unable to add SSH public key to authorized keys");
     }
-
+    
+    public static String addBracketsIfNeeded(String ipAddress) {
+        return ipv6Enabled() && !ipAddress.contains("[") ? String.format("[%s]", ipAddress) : ipAddress;
+    }
```
jtnord, line 59: `return new URL("http", ipAddress, 666, "").getHost()`?

you do not really want to care if ipv6 is enabled or not - or if you have a DNS name or an IPAddress..

src/test/java/plugins/WorkflowPluginTest.java
```diff
@@ -231,7 +231,7 @@ private String createGitRepositoryInDockerContainer() throws IOException {
         GitContainer gitContainer = gitServer.get();
         try (GitRepo repo = new GitRepo()) {
             repo.changeAndCommitFoo("Initial commit");
-            repo.transferToDockerContainer(gitContainer.host(), gitContainer.port());
+            repo.transferToDockerContainer(GitContainer.addBracketsIfNeeded(gitContainer.host()), gitContainer.port());
```
jtnord, line 234: fix `GitRepo`?  after all you are passing in a host not a URL?

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/LdapContainer.java
imonteroperez: I get your point, but AFAIK this `host` is to be used by docker fixtures to connect via `tcp` so finally it is an address and requires to be specified using brackets

src/test/java/plugins/LdapPluginTest.java
imonteroperez: Ditto https://github.com/jenkinsci/acceptance-test-harness/pull/1510#discussion_r1537489501

src/test/java/plugins/WorkflowPluginTest.java
imonteroperez: This `transferToDockerContainer` method is using jsch which will create a session to connect to the host as an address, so requires brackets AFAIK

jtnord: > This `transferToDockerContainer` method is using jsch which will create a session to connect to the host as an address, so requires brackets AFAIK

to be clear host addresses do not contain square brackets...  Some usage of them in some formats (e.g. URLs) may require them to be escaped, but that is up to the implementation to do not a caller.
Today this is using `jsch` but tomorrow it could be using MINA or a different library which does not have this bug.

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitContainer.java
jtnord: > Not following your proposal here, this method is adding brackets to an IPAddress if IPv6 is enabled.

it is adding square brackets **only* if ipv6Enabled() returns true, which is only if you are running dual stack and have asked for this to be enabled.

if you are running on an IPV6 only machine (not dual stack) then you will get an ipv6 address here - and then you are not escaping it at all.

what I am saying is you should not care about the enabling of ipv6 or not - if you use a URL to do the parsing then you will get an escaped host if needed in all situations (dual stack and preferring ipv6, single ipv6 stack etc...).
it will also not escape something if a hostname was returned rather than an IPAddress.

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/LdapContainer.java
jtnord: I do not follow your response,  IP uses bytes and it does not use `[]` in the address portion of the packets, so any `tcp` or `udp` connection is irrelevant.

src/test/java/plugins/WorkflowPluginTest.java
imonteroperez: I know, but today is today :-) and today based on the current `jsch` implementation this is not working under a dual-stack environment giving the following exception
```
Caused by: com.jcraft.jsch.JSchException: java.net.ConnectException: Connection refused
	at com.jcraft.jsch.Util.createSocket(Util.java:349)
	at com.jcraft.jsch.Session.connect(Session.java:215)
	at com.jcraft.jsch.Session.connect(Session.java:183)
	at org.jenkinsci.test.acceptance.plugins.git.GitRepo.transferToDockerContainer(GitRepo.java:289)
	... 42 more
```
So, this PR is trying to make it work there at the same time works under IPv4. Is there room for improvement? Sure thing, but at least this is better than nothing.

src/test/java/plugins/LdapPluginTest.java
imonteroperez: Yes, but at some point this is used to create an URL which is not working without brackets on dual-stack mode

```
Caused: javax.naming.NamingException: Cannot parse url: ldap://2600:1900:4020:cd9b::6:389/dc=jenkins-ci,dc=org [Root exception is java.net.MalformedURLException: unsupported authority: 2600:1900:4020:cd9b::6:389]
	at java.naming/com.sun.jndi.ldap.LdapURL.<init>(LdapURL.java:115)
	at java.naming/com.sun.jndi.ldap.DefaultLdapDnsProvider.lookupEndpoints(DefaultLdapDnsProvider.java:48)
	at java.naming/com.sun.jndi.ldap.LdapDnsProviderService.lookupEndpoints(LdapDnsProviderService.java:109)
	at java.naming/com.sun.jndi.ldap.LdapCtxFactory.getUsingURL(LdapCtxFactory.java:166)
	at java.naming/com.sun.jndi.ldap.LdapCtxFactory.getUsingURLs(LdapCtxFactory.java:247)
	at java.naming/com.sun.jndi.ldap.LdapCtxFactory.getLdapCtxInstance(LdapCtxFactory.java:154)
	at java.naming/com.sun.jndi.ldap.LdapCtxFactory.getInitialContext(LdapCtxFactory.java:84)
	at java.naming/javax.naming.spi.NamingManager.getInitialContext(NamingManager.java:732)
	at java.naming/javax.naming.InitialContext.getDefaultInitCtx(InitialContext.java:305)
	at java.naming/javax.naming.InitialContext.init(InitialContext.java:236)
	at java.naming/javax.naming.ldap.InitialLdapContext.<init>(InitialLdapContext.java:154)
	at org.springframework.ldap.core.support.LdapContextSource.getDirContextInstance(LdapContextSource.java:42)
	at org.springframework.ldap.core.support.AbstractContextSource.createContext(AbstractContextSource.java:351)
```

This change makes it to work under IPv6 dual stack and IPv4

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitContainer.java
imonteroperez: Agree, but this is the first attempt to fix this on dual-stack mode, if not applied then it hits 
```
Cloning the remote Git repository
Cloning repository ssh:///git@2600:1900:4020:cd9b::6:22/home/git/gitRepo
 > git init /root/ath/target/jenkinshome11706254101610635272/workspace/busy_analyst # timeout=10
Fetching upstream changes from ssh:///git@2600:1900:4020:cd9b::6:22/home/git/gitRepo
 > git --version # timeout=10
 > git --version # 'git version 2.34.1'
 > git fetch --tags --force --progress -- ssh:///git@2600:1900:4020:cd9b::6:22/home/git/gitRepo +refs/heads/*:refs/remotes/origin/* # timeout=10
ERROR: Error cloning remote repo 'origin'
hudson.plugins.git.GitException: Command "git fetch --tags --force --progress -- ssh:///git@2600:1900:4020:cd9b::6:22/home/git/gitRepo +refs/heads/*:refs/remotes/origin/*" returned status code 128:
stdout: 
stderr: ssh: Could not resolve hostname : Name or service not known
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

 at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(CliGitAPIImpl.java:2842)
 at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandWithCredentials(CliGitAPIImpl.java:2185)
 at org.jenkinsci.plugins.gitclient.CliGitAPIImpl$1.execute(CliGitAPIImpl.java:635)
 at org.jenkinsci.plugins.gitclient.CliGitAPIImpl$2.execute(CliGitAPIImpl.java:871)
 at hudson.plugins.git.GitSCM.retrieveChanges(GitSCM.java:1222)
 at hudson.plugins.git.GitSCM.checkout(GitSCM.java:1305)
 at hudson.scm.SCM.checkout(SCM.java:540)
 at hudson.model.AbstractProject.checkout(AbstractProject.java:1248)
 at hudson.model.AbstractBuild$AbstractBuildExecution.defaultCheckout(AbstractBuild.java:649)
 at jenkins.scm.SCMCheckoutStrategy.checkout(SCMCheckoutStrategy.java:85)
 at hudson.model.AbstractBuild$AbstractBuildExecution.run(AbstractBuild.java:521)
 at hudson.model.Run.execute(Run.java:1841)
 at hudson.model.FreeStyleBuild.run(FreeStyleBuild.java:44)
 at hudson.model.ResourceController.execute(ResourceController.java:101)
 at hudson.model.Executor.run(Executor.java:442)
ERROR: Error cloning remote repo 'origin'
```

src/test/java/plugins/GitPluginTest.java
imonteroperez: Done in https://github.com/jenkinsci/acceptance-test-harness/pull/1510/commits/40e880730b2f0b251f11049455f3df2ffd30a370

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitContainer.java
jtnord: right - but if you use my suggestion instead it should work regardless of dual stack single stack ipv6/v4 and regardless of having to specify some arbitrary flag.

imonteroperez: But `GitContainer` is using `ipBound` to retrieve the host info, so at the end it is using the IP (not hostname), same for `LdapContainer`

jtnord: I think I am not getting my point across well.

`ipv6Enabled` is a misnomer - it is not that IPV6 is enabled or not - only that java was started with the system property to prefer ipv6 addresses.
this ONLY makes a difference on dual stac


If you run this code on pure ipv6 stack that preference AFAICT will be `false` and then this escaping will not occur, and this code will then fail.

The code should not care if it is preferring ipv6 or ipv4.  All it should care about is if `ipAddress` is an ipv6 address (not already enclosed in `[` ) then to escape it.

And helpfully there is some code that can do this for you, it is in `URL` and pointed out at the start of this thread.  If it is passed an ipv4 literal address it will get an opv4 litteral address, pass it an ipv6 litteral and you get back the litteral address enclosed in `[]`, pass it something already enclosed and you get that back, pass it anything else like a string and you get that back to.

imonteroperez: > ipv6Enabled is a misnomer - it is not that IPV6 is enabled or not - only that java was started with the system property to prefer ipv6 addresses. This ONLY makes a difference on dual stack

Then the issue is on https://github.com/jenkinsci/docker-fixtures/blob/master/src/main/java/org/jenkinsci/test/acceptance/docker/DockerContainer.java#L239-L241 not in this PR.

> If you run this code on pure ipv6 stack that preference AFAICT will be false and then this escaping will not occur, and this code will then fail.

Agree, the same if you run it as it is on dual stack before this PR. With this PR it will work (on dual stack)

jtnord: > Agree, the same if you run it as it is on dual stack before this PR. With this PR it will work (on dual stack)

but with the suggestion it would work everywhere! and be simpler code!

raul-arabaolaza: My understanding from James comment is that instead of 

`ssh://git@" + addBracketsIfNeeded(host()) + ":" + port() + REPO_DIR` 

You build an URL object which can take a bracketed or non bracketed ipv6 ip as `host` and then use the `toExternalForm()` method.

This way the ATH do not care at all about the stack and will work everywhere

jtnord: FTR I am not opposed to a function that uses `new URL(....).getHost()`  rather than inlining it (to make the code more understandable)

jtnord: close but `toExternalForm` will give you a port and scheme, which is not what you need and then ned some parsing.  calling `getHost()` is enough to get the escaped (if required) host after construction.

src/test/java/plugins/WorkflowPluginTest.java
raul-arabaolaza: I agree with James here, if jsch is not able to generate proper urls then the point to fix the issue is in the code that calls it and not before. A host is a host and a url is another thing, so your parameters should not be an improper host representation to hack how the url representation is generated by the jsch library...

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitContainer.java
imonteroperez: Updated in https://github.com/jenkinsci/acceptance-test-harness/pull/1510/commits/5338f366038f3de2105851ac1ff39d83064583a8 to improve readability

jtnord (approved):

src/test/java/plugins/WorkflowPluginTest.java
imonteroperez: Addressed in https://github.com/jenkinsci/acceptance-test-harness/pull/1510/commits/5338f366038f3de2105851ac1ff39d83064583a8

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/LdapContainer.java
jtnord: issue here is that LDAP plugin field is not wanting a hostname but a portion of hosename:port and optionally scheme (ldap://hostname)  as such the plugin itself is requireing this and the host is just missnamed.
thus not an issue in the test but production code.

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitContainer.java
raul-arabaolaza: > close but toExternalForm will give you a port and scheme, which is not what you need and would then need some parsing to get the (possibly escaped) host.

I mean for existing code like https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitContainer.java#L30-L37 which is basically constructing the external form of an url by using strings...

imonteroperez:
@jtnord all test failures are unrelated with the proposed changes, asking for a merge :-)

jtnord:
[failures](https://ci.jenkins.io/job/Core/job/acceptance-test-harness/job/PR-1510/8/testReport/) all seem to be related to https://github.com/jenkinsci/acceptance-test-harness/issues/1526

uhafner:
This breaks the API for downstream plugins! 

See https://github.com/jenkinsci/warnings-ng-plugin/pull/1774 and compile error:
See https://github.com/jenkinsci/warnings-ng-plugin/actions/runs/9809905482/job/27088840082?pr=1774

PR #1687, opened 2024-08-26
Unpin plugin versions

basil:
These pinned versions are so old that I doubt we are managing to install them successfully.


timja (approved):
`WithPlugins` just signifies the minimum version the test requires, generally not needed these days as we just test latest for plugins

PR #1701, opened 2024-09-02
Add oic auth e2e

fcojfernandez:
Adding a test for the `oic-auth` plugin. The test relies on test-containers to start a container for `keycloak` to provide the user, groups and roles.

1. Configures the keycloak container
    1. Create a test realm with a couple of roles
    2. Create a client for Jenkins
    3. Create users and groups, adding users to the groups and assigning roles to the groups
2. Configures the SecurityRealm to use keycloak. A new `SecurityRealm` class has been created for this purpose
3. Adds two tests so the login and logout is tested from Jenkins to Keycloak and from Keycloak to Jenkins.
4. As part of the use check, the users are checked with their roles, which comes from the groups proving the user details are provided by the OICD provider through the `oic-auth` plugin. To check easily the roles, the `whoAmI` endpoint is used

### Submitter checklist


src/test/java/plugins/OicAuthPluginTest.java
```diff
@@ -0,0 +1,254 @@
...
+
+        assertThrows("User has logged out from keycloak", NoSuchElementException.class,
+                     () -> keycloakUtils.getUser(keycloak.getAuthServerUrl(), REALM));
+    }
+
+    private void assertLoggedUser(KeycloakUtils.User expectedUser, String roleToCheck) {
+        assertThat("User has logged in Jenkins", jenkins.getCurrentUser().id(), is(expectedUser.getId()));
+        jenkins.visit("whoAmI");
+
+        // TODO if needed for more tests, consider to create a proper WhoAmI PageObject
```
jtnord, line 239: Use the `API` luke!
`$jenkinsUrl/whoAmI/api/xml?tree=authorities`  (or use `getJson` but that would require a `PageObject` (at least an anonymous inner class) 

Then any page redesign will not break the test!

jtnord:
tests are failing as there is no docker available... (which is kinda fun for the ATH running here!)

Tests should skip if docker is not available, but the ATH run in ci.jenkins.io surely should have docker!!

```
org.testcontainers.containers.ContainerFetchException: Can't get Docker image: RemoteDockerImage(imageName=quay.io/keycloak/keycloak:25.0, imagePullPolicy=DefaultPullPolicy(), imageNameSubstitutor=org.testcontainers.utility.ImageNameSubstitutor$LogWrappedImageNameSubstitutor@5c7bfdc1)
	at org.testcontainers.containers.GenericContainer.getDockerImageName(GenericContainer.java:1364)
	at org.testcontainers.containers.GenericContainer.logger(GenericContainer.java:671)
	at dasniko.testcontainers.keycloak.ExtendableKeycloakContainer.<init>(ExtendableKeycloakContainer.java:138)
	at dasniko.testcontainers.keycloak.ExtendableKeycloakContainer.<init>(ExtendableKeycloakContainer.java:126)
	at dasniko.testcontainers.keycloak.KeycloakContainer.<init>(KeycloakContainer.java:24)
	at plugins.OicAuthPluginTest.<init>(OicAuthPluginTest.java:46)
	at java.base/jdk.internal.reflect.DirectConstructorHandleAccessor.newInstance(DirectConstructorHandleAccessor.java:62)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:502)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:486)
	at org.junit.runners.BlockJUnit4ClassRunner.createTest(BlockJUnit4ClassRunner.java:250)
	at org.junit.runners.BlockJUnit4ClassRunner.createTest(BlockJUnit4ClassRunner.java:260)
	at org.junit.runners.BlockJUnit4ClassRunner$2.runReflectiveCall(BlockJUnit4ClassRunner.java:309)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.BlockJUnit4ClassRunner.methodBlock(BlockJUnit4ClassRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:316)
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeWithRerun(JUnit4Provider.java:240)
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:214)
	at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:155)
	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:385)
	at org.apache.maven.surefire.booter.ForkedBooter.execute(ForkedBooter.java:162)
	at org.apache.maven.surefire.booter.ForkedBooter.run(ForkedBooter.java:507)
	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:495)
Caused by: java.lang.IllegalStateException: Could not find a valid Docker environment. Please see logs and check configuration
	at org.testcontainers.dockerclient.DockerClientProviderStrategy.lambda$getFirstValidStrategy$7(DockerClientProviderStrategy.java:277)
	at java.base/java.util.Optional.orElseThrow(Optional.java:403)
	at org.testcontainers.dockerclient.DockerClientProviderStrategy.getFirstValidStrategy(DockerClientProviderStrategy.java:268)
	at org.testcontainers.DockerClientFactory.getOrInitializeStrategy(DockerClientFactory.java:152)
	at org.testcontainers.DockerClientFactory.client(DockerClientFactory.java:194)
	at org.testcontainers.images.LocalImagesCache.get(LocalImagesCache.java:31)
	at org.testcontainers.images.AbstractImagePullPolicy.shouldPull(AbstractImagePullPolicy.java:18)
	at org.testcontainers.images.RemoteDockerImage.resolve(RemoteDockerImage.java:75)
	at org.testcontainers.images.RemoteDockerImage.resolve(RemoteDockerImage.java:33)
	at org.testcontainers.utility.LazyFuture.getResolvedValue(LazyFuture.java:20)
	at org.testcontainers.utility.LazyFuture.get(LazyFuture.java:41)
	at org.testcontainers.containers.GenericContainer.getDockerImageName(GenericContainer.java:1362)
	... 32 more
```

src/test/java/plugins/OicAuthPluginTest.java
```diff
@@ -0,0 +1,254 @@
...
+
+        jenkins.clickLink("log in");
+        keycloakUtils.login(bob.getUserName());
+        assertLoggedUser(bob, "jenkinsAdmin");
+
+        jenkins.logout();
+        jenkins.open();
+        assertLoggedOut();
+
+        // logout from Jenkins does mean logout from keycloak
```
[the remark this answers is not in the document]
fcojfernandez: good catch. This comment was before setting the logout from the openId provider and I forgot to update it. It's outdated

src/test/java/plugins/OicAuthPluginTest.java
fcojfernandez: That was my first try and the api is not so simple to use here

**whoAmI/api/json** -> generated output differs from Chrome and Firefox
`jenkins.getPageSource()` will return in firefox

```
<html platform="linux" class="theme-dark" dir="ltr" style="scroll-behavior: initial;"><head><meta http-equiv="Content-Security-Policy" content="default-src 'none'; script-src resource:; img-src 'self';"><link rel="stylesheet" type="text/css" href="chrome://devtools-jsonview-styles/content/main.css"><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="viewer-config" src="resource://devtools/client/jsonview/viewer-config.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="json-viewer" src="resource://devtools/client/jsonview/json-viewer.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/vendor/react-dom" src="resource://devtools/client/shared/vendor/react-dom.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/react-utils" src="resource://devtools/client/shared/react-utils.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/jsonview/components/MainTabbedArea" src="resource://devtools/client/jsonview/components/MainTabbedArea.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/tree/TreeView" src="resource://devtools/client/shared/components/tree/TreeView.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/vendor/react" src="resource://devtools/client/shared/vendor/react.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/vendor/react-prop-types" src="resource://devtools/client/shared/vendor/react-prop-types.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/jsonview/components/JsonPanel" src="resource://devtools/client/jsonview/components/JsonPanel.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/jsonview/components/TextPanel" src="resource://devtools/client/jsonview/components/TextPanel.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/jsonview/components/HeadersPanel" src="resource://devtools/client/jsonview/components/HeadersPanel.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/tabs/Tabs" src="resource://devtools/client/shared/components/tabs/Tabs.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/vendor/react-dom-factories" src="resource://devtools/client/shared/vendor/react-dom-factories.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/jsonview/components/JsonToolbar" src="resource://devtools/client/jsonview/components/JsonToolbar.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/constants" src="resource://devtools/client/shared/components/reps/reps/constants.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/rep" src="resource://devtools/client/shared/components/reps/reps/rep.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/jsonview/components/TextToolbar" src="resource://devtools/client/jsonview/components/TextToolbar.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/jsonview/components/LiveText" src="resource://devtools/client/jsonview/components/LiveText.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/jsonview/components/Headers" src="resource://devtools/client/jsonview/components/Headers.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/jsonview/components/HeadersToolbar" src="resource://devtools/client/jsonview/components/HeadersToolbar.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/tree/ObjectProvider" src="resource://devtools/client/shared/components/tree/ObjectProvider.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/tree/TreeRow" src="resource://devtools/client/shared/components/tree/TreeRow.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/tree/TreeHeader" src="resource://devtools/client/shared/components/tree/TreeHeader.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/scroll" src="resource://devtools/client/shared/scroll.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/jsonview/components/SearchBox" src="resource://devtools/client/jsonview/components/SearchBox.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/jsonview/components/reps/Toolbar" src="resource://devtools/client/jsonview/components/reps/Toolbar.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/undefined" src="resource://devtools/client/shared/components/reps/reps/undefined.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/null" src="resource://devtools/client/shared/components/reps/reps/null.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/string" src="resource://devtools/client/shared/components/reps/reps/string.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/number" src="resource://devtools/client/shared/components/reps/reps/number.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/array" src="resource://devtools/client/shared/components/reps/reps/array.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/object" src="resource://devtools/client/shared/components/reps/reps/object.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/symbol" src="resource://devtools/client/shared/components/reps/reps/symbol.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/infinity" src="resource://devtools/client/shared/components/reps/reps/infinity.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/nan" src="resource://devtools/client/shared/components/reps/reps/nan.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/accessor" src="resource://devtools/client/shared/components/reps/reps/accessor.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/accessible" src="resource://devtools/client/shared/components/reps/reps/accessible.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/attribute" src="resource://devtools/client/shared/components/reps/reps/attribute.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/big-int" src="resource://devtools/client/shared/components/reps/reps/big-int.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/date-time" src="resource://devtools/client/shared/components/reps/reps/date-time.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/document" src="resource://devtools/client/shared/components/reps/reps/document.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/document-type" src="resource://devtools/client/shared/components/reps/reps/document-type.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/event" src="resource://devtools/client/shared/components/reps/reps/event.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/function" src="resource://devtools/client/shared/components/reps/reps/function.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/promise" src="resource://devtools/client/shared/components/reps/reps/promise.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/regexp" src="resource://devtools/client/shared/components/reps/reps/regexp.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/stylesheet" src="resource://devtools/client/shared/components/reps/reps/stylesheet.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/comment-node" src="resource://devtools/client/shared/components/reps/reps/comment-node.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/element-node" src="resource://devtools/client/shared/components/reps/reps/element-node.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/text-node" src="resource://devtools/client/shared/components/reps/reps/text-node.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/error" src="resource://devtools/client/shared/components/reps/reps/error.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/window" src="resource://devtools/client/shared/components/reps/reps/window.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/object-with-text" src="resource://devtools/client/shared/components/reps/reps/object-with-text.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/object-with-url" src="resource://devtools/client/shared/components/reps/reps/object-with-url.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/grip-array" src="resource://devtools/client/shared/components/reps/reps/grip-array.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/grip-entry" src="resource://devtools/client/shared/components/reps/reps/grip-entry.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/grip-map" src="resource://devtools/client/shared/components/reps/reps/grip-map.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/grip" src="resource://devtools/client/shared/components/reps/reps/grip.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/tree/TreeCell" src="resource://devtools/client/shared/components/tree/TreeCell.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/tree/LabelCell" src="resource://devtools/client/shared/components/tree/LabelCell.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/focus" src="resource://devtools/client/shared/focus.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/rep-utils" src="resource://devtools/client/shared/components/reps/reps/rep-utils.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/shared/dom-node-constants" src="resource://devtools/client/shared/components/reps/shared/dom-node-constants.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/reps/prop-rep" src="resource://devtools/client/shared/components/reps/reps/prop-rep.js"></script><script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="devtools/client/shared/components/reps/shared/grip-length-bubble" src="resource://devtools/client/shared/components/reps/shared/grip-length-bubble.js"></script></head><body><div id="content"><div class="tabs tabs-tall "><nav class="tabs-navigation"><ul class="tabs-menu" role="tablist"><li class="tabs-menu-item json is-active" role="presentation"><span class="devtools-tab-line"></span><a id="json-tab" tabindex="0" title="JSON" aria-controls="json-panel" aria-selected="true" role="tab" data-tab-index="0">JSON</a></li><li class="tabs-menu-item rawdata " role="presentation"><span class="devtools-tab-line"></span><a id="rawdata-tab" tabindex="-1" title="Raw Data" aria-controls="rawdata-panel" aria-selected="false" role="tab" data-tab-index="1">Raw Data</a></li><li class="tabs-menu-item headers " role="presentation"><span class="devtools-tab-line"></span><a id="headers-tab" tabindex="-1" title="Headers" aria-controls="headers-panel" aria-selected="false" role="tab" data-tab-index="2">Headers</a></li></ul></nav><div class="panels"><div id="json-panel" style="visibility: visible; height: 100%;" class="tab-panel-box" role="tabpanel" aria-labelledby="json-tab"><div class="tab-panel json"><div class="jsonPanelBox tab-panel-inner"><div class="toolbar"><button class="btn save">Save</button><button class="btn copy">Copy</button><button class="btn collapse">Collapse All</button><button class="btn expand">Expand All</button><div class="devtools-separator"></div><div class="devtools-searchbox"><input class="searchBox devtools-filterinput" placeholder="Filter JSON" value=""><button class="devtools-searchinput-clear" hidden=""></button></div></div><div class="panelContent" id="json-scrolling-panel" tabindex="0"><table class="treeTable" role="tree" tabindex="0" aria-label="" cellpadding="0" cellspacing="0"><thead role="presentation"><tr class="" role="presentation"><td class="" role="presentation" id="default"></td><td class="" style="width: 100%;" role="presentation" id="value"></td></tr></thead><tbody role="presentation" tabindex="-1"><tr id="/_class" role="treeitem" aria-level="1" aria-selected="false" class="treeRow stringRow opened"><td class="treeLabelCell" style="--tree-label-cell-indent: 0px;" role="presentation"><span class="treeIcon open" role="presentation"></span><span class="treeLabel stringLabel" data-level="0">_class</span></td><td class="treeValueCell stringCell" role="presentation"><span><span class="objectBox objectBox-string">"hudson.security.WhoAmI"</span></span></td></tr><tr id="/anonymous" role="treeitem" aria-level="1" aria-selected="false" class="treeRow booleanRow opened"><td class="treeLabelCell" style="--tree-label-cell-indent: 0px;" role="presentation"><span class="treeIcon open" role="presentation"></span><span class="treeLabel booleanLabel" data-level="0">anonymous</span></td><td class="treeValueCell booleanCell" role="presentation"><span><span class="objectBox objectBox-number">false</span></span></td></tr><tr id="/authenticated" role="treeitem" aria-level="1" aria-selected="false" class="treeRow booleanRow opened"><td class="treeLabelCell" style="--tree-label-cell-indent: 0px;" role="presentation"><span class="treeIcon open" role="presentation"></span><span class="treeLabel booleanLabel" data-level="0">authenticated</span></td><td class="treeValueCell booleanCell" role="presentation"><span><span class="objectBox objectBox-number">true</span></span></td></tr><tr id="/authorities" role="treeitem" aria-level="1" aria-selected="false" aria-expanded="true" class="treeRow objectRow hasChildren opened"><td class="treeLabelCell" style="--tree-label-cell-indent: 0px;" role="presentation"><span class="treeIcon theme-twisty open" role="presentation"></span><span class="treeLabel objectLabel" data-level="0">authorities</span></td><td class="treeValueCell objectCell" role="presentation"><span></span></td></tr><tr id="/authorities/0" role="treeitem" aria-level="2" aria-selected="false" class="treeRow stringRow opened"><td class="treeLabelCell" style="--tree-label-cell-indent: 16px;" role="presentation"><span class="treeIcon open" role="presentation"></span><span class="treeLabel stringLabel" data-level="1">0</span></td><td class="treeValueCell stringCell" role="presentation"><span><span class="objectBox objectBox-string">"authenticated"</span></span></td></tr><tr id="/authorities/1" role="treeitem" aria-level="2" aria-selected="false" class="treeRow stringRow opened"><td class="treeLabelCell" style="--tree-label-cell-indent: 16px;" role="presentation"><span class="treeIcon open" role="presentation"></span><span class="treeLabel stringLabel" data-level="1">1</span></td><td class="treeValueCell stringCell" role="presentation"><span><span class="objectBox objectBox-string">"jenkinsAdmin"</span></span></td></tr><tr id="/authorities/2" role="treeitem" aria-level="2" aria-selected="false" class="treeRow stringRow opened"><td class="treeLabelCell" style="--tree-label-cell-indent: 16px;" role="presentation"><span class="treeIcon open" role="presentation"></span><span class="treeLabel stringLabel" data-level="1">2</span></td><td class="treeValueCell stringCell" role="presentation"><span><span class="objectBox objectBox-string">"default-roles-test-realm"</span></span></td></tr><tr id="/authorities/3" role="treeitem" aria-level="2" aria-selected="false" class="treeRow stringRow opened"><td class="treeLabelCell" style="--tree-label-cell-indent: 16px;" role="presentation"><span class="treeIcon open" role="presentation"></span><span class="treeLabel stringLabel" data-level="1">3</span></td><td class="treeValueCell stringCell" role="presentation"><span><span class="objectBox objectBox-string">"offline_access"</span></span></td></tr><tr id="/authorities/4" role="treeitem" aria-level="2" aria-selected="false" class="treeRow stringRow opened"><td class="treeLabelCell" style="--tree-label-cell-indent: 16px;" role="presentation"><span class="treeIcon open" role="presentation"></span><span class="treeLabel stringLabel" data-level="1">4</span></td><td class="treeValueCell stringCell" role="presentation"><span><span class="objectBox objectBox-string">"uma_authorization"</span></span></td></tr><tr id="/name" role="treeitem" aria-level="1" aria-selected="false" class="treeRow stringRow opened"><td class="treeLabelCell" style="--tree-label-cell-indent: 0px;" role="presentation"><span class="treeIcon open" role="presentation"></span><span class="treeLabel stringLabel" data-level="0">name</span></td><td class="treeValueCell stringCell" role="presentation"><span><span class="objectBox objectBox-string">"11d9bbf1-ded6-4023-a7ca-3a756c2850a2"</span></span></td></tr></tbody></table></div></div></div></div><div id="rawdata-panel" style="visibility: hidden; height: 0px;" class="tab-panel-box hidden" role="tabpanel" aria-labelledby="rawdata-tab"></div><div id="headers-panel" style="visibility: hidden; height: 0px;" class="tab-panel-box hidden" role="tabpanel" aria-labelledby="headers-tab"></div></div></div></div><script src="resource://devtools/client/jsonview/lib/require.js" data-main="resource://devtools/client/jsonview/viewer-config.js"></script></body></html>
```
while in Chrome
```
<html style="scroll-behavior: initial;"><head><meta name="color-scheme" content="light dark"><meta charset="utf-8"></head><body><pre>{"_class":"hudson.security.WhoAmI","anonymous":false,"authenticated":true,"authorities":["authenticated","jenkinsAdmin","default-roles-test-realm","offline_access","uma_authorization"],"name":"e860be16-d8b9-4fc2-9ca6-f1dcb65169c4"}</pre><div class="json-formatter-container"></div></body></html>
```

and `CapybaraPortingLayerImpl.getPageContent(this.driver) `will generate in firefox
```
JSON
Raw Data
Headers
Save
Copy
Collapse All
Expand All
_class "hudson.security.WhoAmI"
anonymous false
authenticated true
authorities
0 "authenticated"
1 "jenkinsAdmin"
2 "default-roles-test-realm"
3 "offline_access"
4 "uma_authorization"
name "11d9bbf1-ded6-4023-a7ca-3a756c2850a2"
```
as a result of 
image

while in Chrome
```
{"_class":"hudson.security.WhoAmI","anonymous":false,"authenticated":true,"authorities":["authenticated","jenkinsAdmin","default-roles-test-realm","offline_access","uma_authorization"],"name":"e860be16-d8b9-4fc2-9ca6-f1dcb65169c4"}
```
as a result of
image

I thought check the browser and process the output accordingly, but it seemed too dirty to me

fcojfernandez: Then I tried the `api/xml`, but both `jenkins.getPageSource()` and `CapybaraPortingLayerImpl.getPageContent(this.driver)` fail as they are expecting to read HTML and this is not HTML but XML

* https://github.com/jenkinsci/acceptance-test-harness/blob/fda9804fe3e68cf63946ed747bcddd7966af4f2f/src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java#L531-L539
* https://github.com/jenkinsci/acceptance-test-harness/blob/fda9804fe3e68cf63946ed747bcddd7966af4f2f/src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java#L541-L546

fcojfernandez: So I opted for the UI, as it seems to me the cleanest solution

src/main/java/org/jenkinsci/test/acceptance/utils/keycloack/KeycloakUtils.java
```diff
@@ -0,0 +1,95 @@
+package org.jenkinsci.test.acceptance.utils.keycloack;
+
+import java.net.URL;
+
+import org.jenkinsci.test.acceptance.po.CapybaraPortingLayerImpl;
+import org.jenkinsci.test.acceptance.utils.ElasticTime;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+import jakarta.inject.Inject;
```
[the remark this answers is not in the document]
jtnord: 🤷 I recalled right, but you can leave it as is.

src/test/java/plugins/OicAuthPluginTest.java
```diff
@@ -0,0 +1,254 @@
...
+    @Before
+    public void setUpKeycloak() throws Exception {
+        configureOIDCProvider();
+        configureRealm();
+    }
+
+    private void configureOIDCProvider() throws Exception {
+        try (Keycloak keycloakAdmin = keycloak.getKeycloakAdminClient()) {
+            // Create Realm
+            RealmRepresentation testRealm = new RealmRepresentation();
```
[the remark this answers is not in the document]
jtnord: I have no strong opinion, hence it was a `NIT`

src/test/java/plugins/OicAuthPluginTest.java
```diff
@@ -0,0 +1,254 @@
...
+        securityRealm.setAutomaticConfiguration(String.format("%s/realms/%s/.well-known/openid-configuration", keycloakUrl, REALM));
+        securityRealm.logoutFromOpenidProvider(true);
+        securityRealm.setPostLogoutUrl(jenkins.url("OicLogout").toExternalForm());
+        securityRealm.setUserFields(null, null, null, "groups");
+        sc.useAuthorizationStrategy(LoggedInAuthorizationStrategy.class);
+        sc.save();
+    }
+
+    @Test
+    public void fromJenkinsToKeycloak() {
```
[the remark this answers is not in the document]
jtnord: javadoc (or non javadoc) comments can always be much more helpful that a name :)

src/test/java/plugins/OicAuthPluginTest.java
jtnord: for self education you can check out [`getJson`](https://github.com/jenkinsci/acceptance-test-harness/blob/fda9804fe3e68cf63946ed747bcddd7966af4f2f/src/main/java/org/jenkinsci/test/acceptance/po/ContainerPageObject.java#L56-L79) to see how it solves this :) 

to get the page, you can interact with webdriver directly (`driver.getPageSource()`).  probably need to turn off some pretty printing, or to use the `view-source:URL` style.

Anyway, you can leave this as is, if/when it breaks we can improve it

fcojfernandez: let me check `getJson`. I didn't find that method and if it works, I prefer to use the JSON object

`driver.getPageSource()` returns the same output as `jenkins.getPageSource()`, I checked that as well

fcojfernandez: ah! I didn't find that method because I looked for something in the `CapybaraPortingLayerImpl` class

olamy:
@fcojfernandez the Jenkinsfile is not standard compared to plugins. In plugins using testcontainers needs `useContainerAgent: false`. So I'm not quite sure about what needs to be done in this scripted Jenkinsfile. 
Looking here it looks to be a different label to use https://github.com/jenkins-infra/pipeline-library/blob/afa988b6beb40944005f02b841492d34c026e268/vars/buildPlugin.groovy#L50
Maybe @dduportal can confirm?

dduportal:
Hello folks, thanks for the pings!

- All the "split tests" stages are running in a Linux VM agents with Docker available
  - https://github.com/jenkinsci/acceptance-test-harness/blob/fda9804fe3e68cf63946ed747bcddd7966af4f2f/Jenkinsfile#L117-L123 defines the labels `docker-highmem<...>`proving this
  - If you reach the step (which the builds 1,2 and 3 of this PR did) https://github.com/jenkinsci/acceptance-test-harness/blob/fda9804fe3e68cf63946ed747bcddd7966af4f2f/Jenkinsfile#L128C1-L128C132, then it means Docker CE is there and ready.
- Container agents are used for other stages (such as the Launchable one) with labels `maven-21`. This is what @olamy (thanks!) described for the usual plugin builds with the pipeline library `buildPlugin()`. I confirm this function is NOT used in the ATH here.


I'm not sure about the root problem: the only log I can get from `testcontainer` is:

```
WARNING: DOCKER_HOST unix:///var/run/docker.sock is not listening
```

I have 2 workings hypothesis for now:

- Might be an issue due to Docker CE 27.2: we upgraded from 27.1.x to 27.2.0 in the VM agents templates yesterday: https://github.com/jenkins-infra/jenkins-infra/pull/3635. But this version seems buggy (we saw failure in Docker on other jobs of ci.jenkins.io). We reverted to Docker CE 27.1 a few minutes ago (https://github.com/jenkins-infra/jenkins-infra/pull/3638) and I've triggered a new build (number **4**) to see the result. 
  - Note: I did NOT spend much time analyzing for now. The infra team will spend the required time if this revert does not fix your build though!
- Second hypothesis: the failure is due to `testcontainer` in the context of "DonD" (Docker on Docker). It's not nested containerization but a container with its own Docker clients (CLI, Jenkins pipeline native Docker, testcontainer native Docker client, etc.) and the Unix socket mounted.
  - You can see the "Unix socket" mount here: https://github.com/jenkinsci/acceptance-test-harness/blob/fda9804fe3e68cf63946ed747bcddd7966af4f2f/Jenkinsfile#L128
  - I do NOT know **why** DonD is used instead of running natively. My best guess is "it's historical" because of the default JDK version which used to be JDK8 or JDK11 months ago. Not the case anymore though (use Jenkins tool for this).
  - Let's cover this hypothesis if the first one is evicted. There would be some facts to check: which user `testcontainers` uses to reach Docker? does this user has permissions to read and write the Unix socket? etc.

dduportal:
> Hello folks, thanks for the pings!
> 
>     * All the "split tests" stages are running in a Linux VM agents with Docker available
>       
>       * https://github.com/jenkinsci/acceptance-test-harness/blob/fda9804fe3e68cf63946ed747bcddd7966af4f2f/Jenkinsfile#L117-L123
>          defines the labels `docker-highmem<...>`proving this
>       * If you reach the step (which the builds 1,2 and 3 of this PR did) https://github.com/jenkinsci/acceptance-test-harness/blob/fda9804fe3e68cf63946ed747bcddd7966af4f2f/Jenkinsfile#L128C1-L128C132, then it means Docker CE is there and ready.
> 
>     * Container agents are used for other stages (such as the Launchable one) with labels `maven-21`. This is what @olamy (thanks!) described for the usual plugin builds with the pipeline library `buildPlugin()`. I confirm this function is NOT used in the ATH here.
> 
> 
> I'm not sure about the root problem: the only log I can get from `testcontainer` is:
> 
> ```
> WARNING: DOCKER_HOST unix:///var/run/docker.sock is not listening
> ```
> 
> I have 2 workings hypothesis for now:
> 
>     * Might be an issue due to Docker CE 27.2: we upgraded from 27.1.x to 27.2.0 in the VM agents templates yesterday: [Bump Packer Agent Templates (All-In-One) Version to 1.92.0 jenkins-infra/jenkins-infra#3635](https://github.com/jenkins-infra/jenkins-infra/pull/3635). But this version seems buggy (we saw failure in Docker on other jobs of ci.jenkins.io). We reverted to Docker CE 27.1 a few minutes ago ([Bump Packer Agent Templates (All-In-One) Version to 1.94.0 jenkins-infra/jenkins-infra#3638](https://github.com/jenkins-infra/jenkins-infra/pull/3638)) and I've triggered a new build (number **4**) to see the result.
>       
>       * Note: I did NOT spend much time analyzing for now. The infra team will spend the required time if this revert does not fix your build though!
> 
>     * Second hypothesis: the failure is due to `testcontainer` in the context of "DonD" (Docker on Docker). It's not nested containerization but a container with its own Docker clients (CLI, Jenkins pipeline native Docker, testcontainer native Docker client, etc.) and the Unix socket mounted.
>       
>       * You can see the "Unix socket" mount here: https://github.com/jenkinsci/acceptance-test-harness/blob/fda9804fe3e68cf63946ed747bcddd7966af4f2f/Jenkinsfile#L128
>       * I do NOT know **why** DonD is used instead of running natively. My best guess is "it's historical" because of the default JDK version which used to be JDK8 or JDK11 months ago. Not the case anymore though (use Jenkins tool for this).
>       * Let's cover this hypothesis if the first one is evicted. There would be some facts to check: which user `testcontainers` uses to reach Docker? does this user has permissions to read and write the Unix socket? etc.

OK so we can remove the first hypothesis: it's NOT the Docker CE version.
=> It's clearly related to `testcontainer` in the ATH context.

As per https://java.testcontainers.org/supported_docker_environment/continuous_integration/dind_patterns/#docker-wormhole-pattern-sibling-docker-containers, the pattern of mounting the `docker.sock` socket is expected to work.

We *might* need to fine tune `testcontainer` for the host's hostname (see https://java.testcontainers.org/features/configuration/#customizing-docker-host-detection) but it will be **after** solving the problem here.

@fcojfernandez can you try running the same command with the ATH 's image (and it's DinD) on your laptop to see if you can reproduce the same issue?

olamy:
Usually with `dind` image using testcontainer may need the following env var:

      DOCKER_HOST: tcp://localhost:2375
      DOCKER_TLS_VERIFY: 0

not sure exactly how the pods are configured though.

dduportal:
> Usually with `dind` image using testcontainer may need the following env var:
> 
> ```
>   DOCKER_HOST: tcp://localhost:2375
>   DOCKER_TLS_VERIFY: 0
> ```
> 
> not sure exactly how the pods are configured though.

Be careful, in the case here (e.g. ATH in ci.jenkins.io).

- It is "DonD", aka Docker on Docker (or "Docker wormhole" as per testcontainer). It's a different pattern than DinD (Docker in Docker). DinD is about nested container engines while DonD reuses the host's Docker engine. Both are really unsecured and bad practises honestly.
- We do not use Pod for ATH but VM with a fully fledged Docker CE Engine

=> Check the pipeline, it is the source of truth (while my memory can betrays me :D) .


The PR here introduces a new pattern (testcontainer test harnesses next to existing Dockerfixture test harnesses) so it's not expected to work at first step. My understanding of testcontainer is that we should run the Maven command directly on the host (and not in a container based on `jenkins/ath`) but that requires the PR to change elements on the pipeline.

dduportal:
I'm currently trying to understand which Maven command to run to reproduce the test (yeah Java is hard for me)

fcojfernandez:
> can you try running the same command with the ATH 's image (and it's DinD) on your laptop to see if you can reproduce the same issue?

@dduportal it's reproducible using the ATH's image :/ I'll try to debug and see how the docker command is executed

dduportal:
For documentation sake, here are the steps for local reproduction (on a Silicon Mac laptop with Docker Desktop, but also an Intel Windows 11 with Docker Desktop in WSL Linux containers mode):

```
docker pull --platform=linux/amd64 jenkins/ath
# From repo root
docker run -ti --platform=linux/amd64 \
  -v /var/run/docker.sock:/var/run/docker.sock \
  -v $(pwd)/target/ath-reports:/reports:rw \
  -v $(pwd):$(pwd) \
  -w $(pwd) \
  --shm-size 2g \
  jenkins/ath
> mvn -V -e -Plts -Dmaven.test.failure.ignore=true -DforkCount=1 -B -Dtest=OicAuthPluginTest
```

jtnord:
> ```
> WARNING: DOCKER_HOST unix:///var/run/docker.sock is not listening
> ```


is misleading, filed https://github.com/testcontainers/testcontainers-java/issues/9195

src/main/resources/ath-container/Dockerfile
```diff
@@ -37,6 +37,13 @@ RUN install -m 0755 -d /etc/apt/keyrings \
   && apt-get clean \
   && rm -rf /var/lib/apt/lists/*
 
+# Despite the docker SUID hack below, test-containers accesses /var/run/docker.sock directly and so we can not rely on the SUID hack.
+# Rather take the docker user group as an arg and make the ath-user a member of that group
+# we retain the suid workaround as this method requires a local build of the container
+# we need to do this before we install docker so that any files have the correct permission
+ARG dockergid=1002
```
basil, line 44: I think the default group ID should match the one used in practice by our production setup (131, I think). Using a default group ID that doesn't match the Ubuntu default seems like it will be a maintenance burden.

src/main/resources/ath-container/Dockerfile
```diff
@@ -37,6 +37,13 @@ RUN install -m 0755 -d /etc/apt/keyrings \
...
+RUN groupadd docker -g $dockergid
```
basil, line 45: Replaying the build with these `Jenkinsfile` changes, it fails:

```
12:49:01   > [ 4/18] RUN groupadd docker -g 999:
12:49:01  0.172 groupadd: GID '999' already exists
```

basil (changes requested):

src/main/resources/ath-container/Dockerfile
jtnord: The default in my Ubuntu setup is 1001...  
There is no default as it depends on the order that packages may be installed.

src/main/resources/ath-container/Dockerfile
jtnord: Well that sucks...

src/main/resources/ath-container/Dockerfile
basil: Interesting. On my Ubuntu 22.04 desktop it's 131. In a fresh `noble` Docker image (installing the `docker.io` package) it's 103. I think you are right that there is no default.

src/main/resources/ath-container/Dockerfile
dduportal: For debugging, we have the following in the ci.jio. environment:

```shell
[2024-09-04T19:45:51.737Z] ++ id -u
[2024-09-04T19:45:51.738Z] ++ id -g
[2024-09-04T19:45:51.738Z] ++ getent group docker
[2024-09-04T19:45:51.739Z] ++ cut -d: -f3
[2024-09-04T19:45:51.739Z] + docker build -t jenkins/ath --build-arg uid=1001 --build-arg gid=1001 --build-arg dockergid=999 ./src/main/resources/ath-container/
```

For the record, we are using the Docker CE package from <https://docs.docker.com/engine/install/ubuntu/> as per https://github.com/jenkins-infra/packer-images/blob/1.94.0/provisioning/ubuntu-provision.sh#L187-L205 to install it. I believe we depends on their setup.
But it could be worth defining a value for the UID/GID for the `docker` user/group and set it up in the install: as far as I know ATH is the only "weird" usage of Docker so I don't mind setting up such a contract with the infra image template. It would be easy to check as we have an [automated test harness](https://github.com/jenkins-infra/packer-images/blob/1.94.0/tests/goss-linux.yaml) to describe such kind of requirement and avoid loosing it in hte future.

src/main/resources/ath-container/Dockerfile
dduportal: OH: interesting. The `999` user is created by one of our early `Dockerfile` steps:

```shell
$ docker run --rm -ti --entrypoint=grep --platform=linux/amd64 ubuntu:noble 999 /etc/group
$ docker run --rm -ti --entrypoint=grep --platform=linux/amd64 jenkins/ath 999 /etc/group
systemd-journal:x:999:
```

That means removing it should not be a problem

dduportal: Confirmed: the user `systemd-journal` is created when installing the `openjdk-*-jdk` packages. I believe they have a dependency to a  `systemd-journal*` package.

- I'm adding a suggestion with a short term fix (deleting the group if existing as we don't really care about this one)
- WDYT about using the temurin JDKs instead of the `openjdk` packages in this context?
  - Faster image build as we could use multi stage builds (parallel)
  - Closer to the rest of the environment

dduportal: See https://github.com/jenkinsci/acceptance-test-harness/pull/1701/files#r1744863012

jtnord: removing system-d (or other groups) does not fill me with confidence.

but all groups created by apt should be dynimcally allocated an id (should not be hard coded) so if I move the group creation first it should be able to do the right thing in all cases

dduportal: SystemD is not expected to be run inside a container unless you do terrible hacks. 
since we tracked this group as a consequence of installing the ubuntu openjdk packages, which can be replaced by temurin standalone jdks, you can do it safely.

however you are correct : if you create the group before the apt installs, it should be good 👍🏼

jtnord: https://github.com/jenkinsci/acceptance-test-harness/pull/1705/commits/f336732785122935d85a2b5f847c6cf699138c0a (different PR as the Jenkinsfile change needs to be from a trusted fork)

basil:
Updated to consume the incremental build of https://github.com/jenkinsci/jenkins/pull/9696 (and converted to draft until that is released)

pom.xml
```diff
@@ -314,6 +315,28 @@
...
+      <groupId>jakarta.annotation</groupId>
+      <artifactId>jakarta.annotation-api</artifactId>
+      <version>2.1.1</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>jakarta.xml.bind</groupId>
+      <artifactId>jakarta.xml.bind-api</artifactId>
+      <version>3.0.1</version>
+    </dependency>
```
basil, line 339: Why the need for these dependencies? I don't see us consuming them directly anywhere in this PR. If this is to resolve some upper bounds error, should this be placed in `<dependencyManagement>` instead?

fcojfernandez: those are dependency of the keycloak testcontainer, more precisely the keycloak admin client. However they have scope provided and they weren't found in the local execution. After your comment I've double checked to post here the stacktrace ... but it's working now in local ( :confused: ). I've removed then and let's check with the CI if they are really needed or not

basil: Thanks, I agree these are still needed. But it wasn't obvious as to why, and I (or someone else) might forget about this conversation and be tempted to remove them in the future, so I think it would be good to add a comment stating what you just said above (in particular, that these are provided dependencies of the Keyclock admin client). That way, their purpose remains clear to future readers.

pom.xml
```diff
@@ -29,7 +29,8 @@
   <properties>
     <changelist>999999-SNAPSHOT</changelist>
     <spotbugs.skip>true</spotbugs.skip>
-    <jenkins.version>2.475</jenkins.version>
+    <!-- TODO https://github.com/jenkinsci/jenkins/pull/9696 -->
+    <jenkins.version>2.476-rc35330.1b_8c1984dc3b_</jenkins.version>
```
jtnord, line 33: Release is available

```suggestion
    <jenkins.version>2.476</jenkins.version>
```

src/test/java/plugins/OicAuthPluginTest.java
```diff
@@ -0,0 +1,307 @@
...
+import org.openqa.selenium.NoSuchElementException;
+
+@WithPlugins("oic-auth")
+public class OicAuthPluginTest extends AbstractJUnitTest {
+
+    private static final String REALM = "test-realm";
+    private static final String CLIENT = "jenkins";
+
+    @Rule
+    public KeycloakContainer keycloak = new KeycloakContainer();
```
jtnord, line 46: this uses the `latest` version of the container at time of running, which could lead to unexpected failures in PRs when a new version of keycloak is published.

Probably better to use a specific version here and get renovate to update it.

jtnord (changes requested):
Doesn't work correctly with a remote-webdriver

the keycloak URL is set to 
`http://localhost:54940/realms/test-realm/`  but this is not accessible, unclear why, most likely this is not the mapped port or the port it not correctly exposed.

Additionally the image should be tied an updated by dependabot.  

(I am working on both issues)

src/test/java/plugins/OicAuthPluginTest.java
jtnord: https://github.com/jenkinsci/acceptance-test-harness/pull/1701/commits/d9fa1de575fa825f5b9c6fbef08873403c69472a

docs/BROWSER.md
```diff
@@ -69,6 +69,8 @@ If the host running maven is different to the host running Selenium (e.g. `remot
 If this is the case you can specify the address to use using:
     `SELENIUM_PROXY_HOSTNAME=ip.address.of.host mvn install`
 **Important**: this could exposed the proxy wider beyond your machine and expose other internal services, so this should only be used on private or internal networks to prevent any information leak.
+The same issue will also impact any other containers started that the tests that the Browser (rather than Jenkins) needs to access.
+For TestContainers you can additionally set `TESTCONTAINERS_HOST_OVERRIDE=ip.address.of.host`
```
jtnord, line 73:
```suggestion
For [Testcontainers](https://testcontainers.com/) you can additionally set `TESTCONTAINERS_HOST_OVERRIDE=ip.address.of.host`
```

docs/DOCKER.md
```diff
@@ -71,3 +71,10 @@ It's best explained with an example:
 # Using a custom docker network
 
 If you are using a custom network for the container that executes the testing you may instruct the docker-fixtures to use the same one by setting the env variable `DOCKER_FIXTURES_NETWORK`to specify the network you want your fixtures to connect to. 
+
+# Using TestContainers with a remote webdriver
```
jtnord, line 75:
```suggestion
# Using Testcontainers with a remote webdriver
```

jtnord (approved):

src/test/java/plugins/OicAuthPluginTest.java
```diff
@@ -0,0 +1,309 @@
...
+import org.keycloak.representations.idm.UserRepresentation;
+import org.openqa.selenium.NoSuchElementException;
+
+@WithPlugins("oic-auth")
+public class OicAuthPluginTest extends AbstractJUnitTest {
+
+    private static final String REALM = "test-realm";
+    private static final String CLIENT = "jenkins";
+
+    private static final String KEYCLOAK_IMAGE="keycloak/keycloak:25.0.4@sha256:bf788a3b7fd737143f98d4cb514cb9599c896acee01a26b2117a10bd99e23e11";
```
jtnord, line 45: this is specifically not the latest to double check renovate confit.
was checked locally with `LOG_LEVEL=debug RENOVATE_CONFIG_FILE=/usr/src/app/.github/renovate.json renovate --platform=local` which did output the new versions, however I still want to see propose an update in CI.

.github/renovate.json
```diff
@@ -81,7 +81,14 @@
       "matchStrings": ["ARG DOCKER_VERSION=(?<currentValue>.*?)\n"],
       "depNameTemplate": "docker",
       "datasourceTemplate": "docker"
+    },
+    {
+      "fileMatch": ["src/test/java/plugins/OicAuthPluginTest.java"],
+      "matchStrings": [".* KEYCLOAK_IMAGE =\n\\s*\"(?<repo>.*?):(?<currentValue>.*?)@(?<currentDigest>sha256:.*?)\";\n"],
+      "depNameTemplate": "{{{repo}}}",
+      "datasourceTemplate": "docker"
```
jtnord, line 89: `LOG_LEVEL=debug RENOVATE_CONFIG_FILE=/usr/src/app/.github/renovate.json renovate --platform=local` shows this will match:

```
             "deps": [
               {
                 "depName": "keycloak/keycloak",
                 "currentValue": "25.0.4",
                 "currentDigest": "sha256:bf788a3b7fd737143f98d4cb514cb9599c896acee01a26b2117a10bd99e23e11",
                 "datasource": "docker",
                 "replaceString": "    private static final String KEYCLOAK_IMAGE =\n            \"keycloak/keycloak:25.0.4@sha256:bf788a3b7fd737143f98d4cb514cb9599c896acee01a26b2117a10bd99e23e11\";\n",
                 "updates": [
                   {
                     "bucket": "non-major",
                     "newVersion": "25.0.5",
                     "newValue": "25.0.5",
                     "newDigest": "sha256:410fce4b9b40e1f4e7f90b44acedbaa6d935bacea432a6884769067b253d46fb",
                     "releaseTimestamp": "2024-09-10T05:21:31.638Z",
                     "newMajor": 25,
                     "newMinor": 0,
                     "newPatch": 5,
                     "updateType": "patch",
                     "branchName": "renovate/keycloak-keycloak-25.x"
                   }
                 ],
                 "packageName": "keycloak/keycloak",
                 "versioning": "docker",
                 "warnings": [],
                 "sourceUrl": "https://github.com/keycloak-rel/keycloak-rel",
                 "registryUrl": "https://index.docker.io",
                 "currentVersion": "25.0.4",
                 "currentVersionTimestamp": "2024-08-19T09:21:17.507Z",
                 "isSingleVersion": true,
                 "fixedVersion": "25.0.4"
               }
```

PR #1706, opened 2024-09-05
Cleanup the docker permission issues.

jtnord:
The permissions on the mounted `docker.sock` where incorrect for the current user, which lead to workarounds setting the docker binary SUID.

However this was a bit hacky and if programmatic access to docker was needed (e.g. TestContainers, or anything else that used the socket and not the binary) then access would fail.

Rather than set the binary SUID which only works for some of the docker use cases, we add the `ath-user` to the docker group that has access to the socket on the host at run time.

extracted from #705 

### Testing done

Testing in this PR and #705 for the test containers case.
https://ci.jenkins.io/job/Core/job/acceptance-test-harness/job/PR-1705/4/testReport/plugins/OicAuthPluginTest/ (shows the test-container tests run and do not fail because of docker)
[LdapPluginTest](https://github.com/jenkinsci/acceptance-test-harness/blob/93f5d22015268caea286a15d7c6288aec1d2c43c/src/test/java/plugins/LdapPluginTest.java) which requires docker is [passing](https://ci.jenkins.io/job/Core/job/acceptance-test-harness/job/PR-1706/1/testReport/plugins/LdapPluginTest/)

Shell script has not been tested, nor has any testing been performed on a macbook with docker-desktop

### Submitter checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -125,7 +125,8 @@ for (int i = 0; i < splits.size(); i++) {
             def image = skipImageBuild ? docker.image('jenkins/ath') : docker.build('jenkins/ath', '--build-arg uid="$(id -u)" --build-arg gid="$(id -g)" ./src/main/resources/ath-container/')
             sh 'mkdir -p target/ath-reports && chmod a+rwx target/ath-reports'
             def cwd = pwd()
-            image.inside("-v /var/run/docker.sock:/var/run/docker.sock -v '${cwd}/target/ath-reports:/reports:rw' --shm-size 2g") {
+            def dockergid = sh label: 'get docker group', returnStdout: true, script: 'getent group docker | cut -d: -f3'
+            image.inside("--group-add ${dockergid} -v /var/run/docker.sock:/var/run/docker.sock -v '${cwd}/target/ath-reports:/reports:rw' --shm-size 2g") {
               def exclusions = splits.get(index).join('\n')
               writeFile file: 'excludes.txt', text: exclusions
               infra.withArtifactCachingProxy {
--- a/ath-container.sh
+++ b/ath-container.sh
@@ -27,7 +27,7 @@ docker build \
 	-t "$tag"
 
 # obtain the groupId to grant to access the docker socket
-dockergid=$(docker run --rm -v /var/run/docker.sock:/var/run/docker.sock alpine stat -c %g /var/run/docker.sock)
+dockergid=$(docker run --rm -v /var/run/docker.sock:/var/run/docker.sock ubuntu:noble stat -c %g /var/run/docker.sock)
 
 docker run \
 	--interactive \
--- a/ath-container.sh
+++ b/ath-container.sh
@@ -34,6 +34,7 @@ docker run \
 	--user ath-user \
 	--workdir /home/ath-user/sources \
 	--shm-size 2g \
+	--group-add $(getent group docker | cut -d: -f3) \
 	-v /var/run/docker.sock:/var/run/docker.sock \
 	-v "$(pwd):/home/ath-user/sources" \
 	-v "${HOME}/.m2/repository:/home/ath-user/.m2/repository" \
--- a/ath-container.sh
+++ b/ath-container.sh
@@ -26,6 +26,9 @@ docker build \
 	"$DIR/src/main/resources/ath-container" \
 	-t "$tag"
 
+# obtain the groupId to grant to access the docker socket
+dockergid=$(docker run --rm -v /var/run/docker.sock:/var/run/docker.sock alpine stat -c %g /var/run/docker.sock)
+
 docker run \
 	--interactive \
 	--tty \
@@ -34,7 +37,7 @@ docker run \
 	--user ath-user \
 	--workdir /home/ath-user/sources \
 	--shm-size 2g \
-	--group-add $(getent group docker | cut -d: -f3) \
+	--group-add ${dockergid} \
 	-v /var/run/docker.sock:/var/run/docker.sock \
 	-v "$(pwd):/home/ath-user/sources" \
 	-v "${HOME}/.m2/repository:/home/ath-user/.m2/repository" \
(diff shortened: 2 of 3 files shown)


ath-container.sh
```diff
@@ -34,6 +34,7 @@ docker run \
 	--user ath-user \
 	--workdir /home/ath-user/sources \
 	--shm-size 2g \
+	--group-add $(getent group docker | cut -d: -f3) \
```
jtnord, line 37: not sure if this will work on mac - does docker-desktop create a group with this name like it does on linux?

src/main/resources/ath-container/Dockerfile
```diff
@@ -86,11 +86,6 @@ RUN deluser --remove-home ubuntu \
   && groupadd ath-user -g $gid \
   && useradd ath-user -l -c 'ATH User' -u $uid -g $gid -m -d /home/ath-user -s /bin/bash
 
-# Set SUID and SGID for docker binary so it can communicate with mapped socket its uid:gid we can not control. Alternative
-# approach used for this is adding ath-user to the group of /var/run/docker.sock but that require root permission we do not
-# have in ENTRYPOINT as the container is started as ath-user.
-RUN chmod ug+s /usr/bin/docker*
-
```
jtnord, line 93: removed as we now have the correct groups when we run (due to `--group-add`).

Jenkinsfile
```diff
@@ -125,7 +125,8 @@ for (int i = 0; i < splits.size(); i++) {
             def image = skipImageBuild ? docker.image('jenkins/ath') : docker.build('jenkins/ath', '--build-arg uid="$(id -u)" --build-arg gid="$(id -g)" ./src/main/resources/ath-container/')
             sh 'mkdir -p target/ath-reports && chmod a+rwx target/ath-reports'
             def cwd = pwd()
-            image.inside("-v /var/run/docker.sock:/var/run/docker.sock -v '${cwd}/target/ath-reports:/reports:rw' --shm-size 2g") {
+            def dockergid = sh label: 'get docker group', returnStdout: true, script: 'getent group docker | cut -d: -f3'
+            image.inside("--group-add ${dockergid} -v /var/run/docker.sock:/var/run/docker.sock -v '${cwd}/target/ath-reports:/reports:rw' --shm-size 2g") {
```
jtnord, line 129: use the supported https://docs.docker.com/engine/containers/run/#additional-groups

ath-container.sh
dduportal: When dealing with Docker Desktop, we want to retrieve the group on server side, eg. on the VM, not on the client side. You can see that as if your client reaches the remote Docker Engine through SSH.

I have the following command in my archives which I've used for years for this trick (e.g. "to collect the GID of the group owner of the socket in the context of DonD"):

```
docker run --rm -v /var/run/docker.sock:/var/run/docker.sock alpine stat -c %g /var/run/docker.sock
```

=> works on macOS (Docker Desktop), Linux (both Docker CE and Docker Desktop), Windows (Docker Desktop in Linux container mode) but also any SSH remote Docker

```
# On Ubuntu with Docker CE 27.2.0 - Docker package
$ docker run --rm -v /var/run/docker.sock:/var/run/docker.sock ubuntu:noble stat -c %g /var/run/docker.sock      
999
```

```
# On macOS with Docker Desktop latest
$ docker run --rm -v /var/run/docker.sock:/var/run/docker.sock ubuntu:noble stat -c %g /var/run/docker.sock      
0
# because it is a special socket shared with the VM  through macOS isolation
```

Jenkinsfile
dduportal:
```suggestion
            image.inside("--group-add $(getent group docker | cut -d: -f3) -v /var/run/docker.sock:/var/run/docker.sock -v '${cwd}/target/ath-reports:/reports:rw' --shm-size 2g") {
```

Would it avoid to execute the shell step before (to avoid a back and forth) ? I remember that `$(id -u)` was used on other cases

jtnord:
build failure due to upstream issue

```
15:28:27   > [ 3/16] RUN install -m 0755 -d /etc/apt/keyrings   && curl -fsSL https://packages.mozilla.org/apt/repo-signing-key.gpg -o /etc/apt/keyrings/packages.mozilla.org.asc   && printf 'deb [signed-by=/etc/apt/keyrings/packages.mozilla.org.asc] https://packages.mozilla.org/apt mozilla main\n' > /etc/apt/sources.list.d/mozilla.list   && printf 'Package: *\nPin: origin packages.mozilla.org\nPin-Priority: 1000\n' > /etc/apt/preferences.d/mozilla   && apt-get update   && DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC apt-get install --no-install-recommends -y     firefox="129.0.2*"   && apt-get clean   && rm -rf /var/lib/apt/lists/*:
15:28:27  0.430 curl: (22) The requested URL returned error: 502
```

As we have a successfull build of the image and we have `LdapPluginTest` passing I am intending to merge with these failings, once approved (to save getting another round of builds run on the infra)

timja (approved):

fcojfernandez (approved):

Jenkinsfile
jtnord: the would appear to me to be very brittle and relies on the image.inside implementation that would currently go via a shell to interpret this.   (which is why I moved it outside).

ath-container.sh
jtnord: c29029d3dd91d8d06d3487488414a5610c3dccf3

basil:
@jtnord Can you also please update https://github.com/jenkinsci/jenkins/blob/db5121061c90fee5c0387e7bd2cf49fe74b31bc7/ath.sh as well? Even though we don't currently run Docker-based tests as part of core PRs, we might in the future, and it seems like good hygiene in general to keep all consumers of this image up-to-date with recent calling conventions.

jtnord:
> @jtnord Can you also please update https://github.com/jenkinsci/jenkins/blob/db5121061c90fee5c0387e7bd2cf49fe74b31bc7/ath.sh as well? Even though we don't currently run Docker-based tests as part of core PRs, we might in the future, and it seems like good hygiene in general to keep all consumers of this image up-to-date with recent calling conventions.

Thanks, I had forgotten about that, filed https://github.com/jenkinsci/jenkins/pull/9701

PR #1733, opened 2024-09-24
Migrate from deprecated `org.eclipse.aether.impl.guice.AetherModule` to Sisu

### LINKED ISSUE #1730: Migrate from deprecated `org.eclipse.aether.impl.guice.AetherModule` to Sisu
Noticed in #1729: We currently rely on `org.eclipse.aether.impl.guice.AetherModule`, which was deprecated in Maven 3.9.5 (which uses Maven Resolver 1.9.16) in https://github.com/apache/maven-resolver/pull/328 and removed in Maven 4.0.0 Alpha 9 (which uses Maven Resolver 2.0.0 Alpha 1) in https://github.com/apache/maven-resolver/pull/340. The Javadoc states:

> This class is about to be dropped in 2.0.0 release. Use [Sisu](https://projects.eclipse.org/projects/technology.sisu) or use [Maven Resolver Supplier](https://maven.apache.org/resolver/third-party-integrations.html) to get `Resolver` instances.

cstamas:
Migrate to Sisu and drop use of deprecated AetherModule (is removed in Resolver 2.x).

This PR uses "fully fledged" Sisu to create `RepositorySystem` instance, and will discover (using sisu index) dynamically any JSR330 component that is present on classpath at the moment `AetherModule` is being constructed.

Also, Guice and Sisu are changed (by exclusion from transitive dep and explicit addition as direct dependency) to "no asm" versions (these have no shaded in ASM), and since ASM was already on path (was 9.2) updated to latest 9.7. Now Guice and Sisu and that something (that pulled ASM in orginally) all use same ASM version 9.7.

### Testing done

CI build

### Submitter checklist

Fixes #1730

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,8 +33,12 @@
     <selenium.version>4.25.0</selenium.version>
     <!-- aligned with selenium -->
     <guava.version>33.3.0-jre</guava.version>
+
+    <!-- Keep them aligned to Maven version -->
     <maven.version>3.9.6</maven.version>
+    <maven-sisu.version>0.9.0.M3</maven-sisu.version>
     <maven-resolver.version>1.9.18</maven-resolver.version>
+
     <groovy.version>3.0.22</groovy.version>
     <jffi.version>1.3.13</jffi.version>
     <spotless.check.skip>false</spotless.check.skip>
@@ -61,6 +65,13 @@
         <groupId>org.slf4j</groupId>
         <artifactId>slf4j-bom</artifactId>
         <version>2.0.16</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+      <dependency>
+        <groupId>org.ow2.asm</groupId>
+        <artifactId>asm-bom</artifactId>
+        <version>9.7</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
@@ -199,6 +210,7 @@ and
       <groupId>com.google.inject</groupId>
       <artifactId>guice</artifactId>
       <version>6.0.0</version>
+      <classifier>classes</classifier>
       <exclusions>
         <exclusion>
           <groupId>com.google.guava</groupId>
@@ -245,11 +257,32 @@ and
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-resolver-provider</artifactId>
       <version>${maven.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>org.eclipse.sisu</groupId>
+          <artifactId>org.eclipse.sisu.inject</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-settings-builder</artifactId>
       <version>${maven.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven.resolver</groupId>
+      <artifactId>maven-resolver-api</artifactId>
+      <version>${maven-resolver.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven.resolver</groupId>
+      <artifactId>maven-resolver-util</artifactId>
+      <version>${maven-resolver.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven.resolver</groupId>
+      <artifactId>maven-resolver-impl</artifactId>
+      <version>${maven-resolver.version}</version>
     </dependency>
     <dependency>
       <groupId>org.apache.maven.resolver</groupId>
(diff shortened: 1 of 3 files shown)


cstamas:
This uses sisu w/ index. Basically whatever you "throw" onto classpath (transport, new checksum, yada) it is discovered and is added to Resolver system. Currently only file and http transports are added.

cstamas:
Also, using asm-less guice and sisu, and provided latest asm (as asm was already present as transitive).

pom.xml
```diff
@@ -33,8 +33,12 @@
     <selenium.version>4.25.0</selenium.version>
     <!-- aligned with selenium -->
     <guava.version>33.3.0-jre</guava.version>
+
+    <!-- Keep them aligned to Maven version -->
     <maven.version>3.9.6</maven.version>
+    <maven-sisu.version>0.9.0.M3</maven-sisu.version>
```
basil, line 38: According to https://github.com/apache/maven/blob/maven-3.9.6/pom.xml#L130 Maven 3.9.6 uses 0.9.0.M2, so I think this should be 0.9.0.M2 in order to follow the comment two lines earlier ("Keep them aligned to Maven version"):

```suggestion
    <maven-sisu.version>0.9.0.M2</maven-sisu.version>
```

Please feel free to ignore this suggestion if I have misunderstood the intention of the comment above.

pom.xml
```diff
@@ -199,6 +210,7 @@ and
       <groupId>com.google.inject</groupId>
       <artifactId>guice</artifactId>
       <version>6.0.0</version>
+      <classifier>classes</classifier>
```
basil, line 212: I can't quite understand what this is for (I have never even used `<classifier>classes</classifier>` before), and tests seemed to work fine when I removed it. Is it better to keep it for some particular reason?

pom.xml
```diff
@@ -281,6 +314,12 @@ and
       <artifactId>groovy-console</artifactId>
       <version>${groovy.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.eclipse.sisu</groupId>
+      <artifactId>org.eclipse.sisu.inject</artifactId>
+      <version>${maven-sisu.version}</version>
+      <classifier>no_asm</classifier>
```
basil, line 320: Since ASM is a transitive dependency of this repository anyway, it seems consistent to use it throughout rather than using it in some places but not others. Plus this makes the code (very slightly) easier to maintain by avoiding the need for a custom classifier:

```suggestion
```

I tested this locally and tests seemed to work fine.

timja (approved):
Thanks! I had the same questions as Basil on the classifiers, rest looks good if it works 😄

pom.xml
basil: I think I understand this now: this seems to be avoiding a shaded copy of ASM. That is better for us, since we already have ASM in our tree and have no need to shade it. If my understanding is correct, I think we should keep the `no_asm` classifier and resolve this conversation.

basil (approved):
Thanks!

cstamas:
Yes, **both** Guice and Sisu contains shaded in ASM. Since Guice 4 and Sisu 0.9.0.M2 (? unsure) both projects offer "no asm shaded" artifacts as well (w/ classifier, Guice classes and Sisu no_asm). Basically you had 3 ASMs on classpath (2 shaded and 3rd was some transitive). Now you have one ASM that is latest.

pom.xml
cstamas: This is "guice w/o asm shaded in"

pom.xml
cstamas: Yup, let be consistent and have them **truly aligned**. Also, checked, and M2 does provide no_asm as well:
https://repo.maven.apache.org/maven2/org/eclipse/sisu/org.eclipse.sisu.inject/0.9.0.M2/

cstamas: In reality does not stir much (as Sisu versions were more important due shaded in Asm, but now you use own ASM), also I don't want to kill current CI jobs, so please merge this  suggestion _before_ merging the PR

cstamas: Oh, 0.9.0.M2 does not have "strict" mode, so if downgrading, this will need change:
remove "false" after INDEX here https://github.com/jenkinsci/acceptance-test-harness/pull/1733/files#diff-14b9e71c9ede5549e786ac9e299d0b2da2969ff999f9c0a9ad55a3849d8abca2R40

cstamas:
One more thing post merge @basil and @timja I just realized:
While debugging, I kinda saw that `AetherModule` is constructed **twice** (I may be wrong here) per test.

But in any case, I am pretty much sure that `RepositorySystem` is not being extended dynamically (as in added new transport, removed transport), so to me it seems likely possible you need _one cached_ instance of `RepositorySystem`. There is no reason to build and then tear down it, as it is _same_. `RepositorySystem` is stateless, while `RepositorySystemSession` is not (!).

So I think your lengthy tests could be mildly improved, if you split the two: keep `RepositorySystem` across whole run (as it has huge object graph, but is stateless and you do not change it), but create `RepositorySystemSession` per-case basis.

PR #1762, opened 2024-10-07
Adjust to jenkinsci-matrix-project-plugin#213

mawinter69:
Adjust to https://github.com/jenkinsci/matrix-project-plugin/pull/213


basil:
Can we make this PR forward-compatible so that it can be merged in advance of https://github.com/jenkinsci/matrix-project-plugin/pull/213, or is that prohibitively difficult?

basil (approved):
Thanks for the PR!

PR #1839, opened 2024-11-20
Adapt to new CSP design

timja:
see https://github.com/jenkinsci/csp-plugin/pull/16

### Testing done

Tested it works if no CSP violations, haven't managed to test with something with violations, artifactory plugin test won't run on my machine, artifactory just won't start.

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/csp/ContentSecurityPolicyReport.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/csp/ContentSecurityPolicyReport.java
@@ -1,6 +1,7 @@
 package org.jenkinsci.test.acceptance.plugins.csp;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import org.jenkinsci.test.acceptance.po.Jenkins;
 import org.jenkinsci.test.acceptance.po.PageObject;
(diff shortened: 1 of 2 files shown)


daniel-beck:
> Tested it works if no CSP violations, haven't managed to test with something with violations

Looks like `LogParserTest` is taking care of this for you?

src/main/java/org/jenkinsci/test/acceptance/plugins/csp/ContentSecurityPolicyReport.java
```diff
@@ -14,7 +15,11 @@ public ContentSecurityPolicyReport(Jenkins context) {
 
     public List<String> getReport() {
         List<String> lines = new ArrayList<>();
-        WebElement table = find(By.className("bigtable"));
+        WebElement table = getElement(By.className("jenkins-table"));
+        if (table == null) {
```
basil, line 19: I've seen cases where we weren't able to successfully browse to the CSP Report page (for example, because the test ended while logged in as a user without admin privileges), so I'm worried this would give a false negative. Can we instead look for the text "No reports" and only return an empty list in that case? Otherwise I think we would be hitting the case where we lack admin privileges and should throw an exception in that case.

daniel-beck:
Haven't looked at `LogParser` in detail what exactly is happening there, but there is a legitimate regression in the CSP plugin here: DBS via Resource Root URL is deliberately intended to not have CSP, but now does. WIP fix in https://github.com/jenkinsci/csp-plugin/pull/19

PR #1982, opened 2025-04-30
Do not use logo but link to home instead

fcojfernandez:
https://github.com/jenkinsci/acceptance-test-harness/pull/1917 changed the criteria to find the link to "/" after the changes in the header. Because there exists https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/jenkins/views/Header.java in Jenkins it is possible to have a customised header that might not have the "jenkins-head-icon" element, so maybe it's better to use the link instead of the id. Otherwise the ATH will only work with the Jenkins OSS header.

### Submitter checklist

### DIFF
--- a/src/test/java/core/FormValidationTest.java
+++ b/src/test/java/core/FormValidationTest.java
@@ -76,7 +76,7 @@ public class FormValidationTest extends AbstractJUnitTest {
 
     private void navigateAway() {
         jenkins.runThenConfirmAlert(
-                () -> driver.findElement(By.id("jenkins-head-icon")).click());
+                () -> driver.findElement(By.xpath("//a[@href='/']")).click());
         sleep(1000); // Needed for some reason
     }
 }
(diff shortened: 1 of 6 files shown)


src/test/java/core/FormValidationTest.java
```diff
@@ -76,7 +76,7 @@ private void jsValidation() {
 
     private void navigateAway() {
         jenkins.runThenConfirmAlert(
-                () -> driver.findElement(By.id("jenkins-head-icon")).click());
+                () -> driver.findElement(By.xpath("//a[@href='/']")).click());
```
jtnord, line 79: Jenkins may not be running on the root context though.

fcojfernandez: In all these tests it is

fcojfernandez: I wasn't aware of https://github.com/jenkinsci/acceptance-test-harness/blob/master/docs/CONTROLLER.md#existing-jenkins-controller-typeexisting ...

basil:
https://github.com/jenkinsci/acceptance-test-harness/pull/1984 is slightly cleaner than this PR (for the reasons explained there) and ought to address the immediate problem, although it is not a full cleanup and more could likely be done.

timja:
No longer needed AFAICT

PR #1657, opened 2024-07-31
Add support for Firefox 129 or newer

basil:
Without dropping support for Firefox 123 and earlier, add support for Firefox 129 or newer. Note that Firefox 124 through 128 are known to not work, as explained in https://github.com/mozilla/geckodriver/issues/2182#issuecomment-2180048632.

### Testing done

- CI build of this PR
- CI build with Firefox 129 in #1648

Fixes #1639

PR #1984, opened 2025-04-30
Clean up `beforeunload` user prompt handling

basil:
#1657 introduced a fragile workaround in `FormValidationTest` (moving from `jenkins.open()` to clicking on a breadcrumb link), and that fragile workaround was copied and pasted in more places in #1749. That code was then adjusted even further in #1917, which is still causing some problems for OEMs that have not adopted the Jenkins header change. This PR does some code cleanup:

- Restores `FormValidationTest` to the less fragile way it was before #1657, effectively reimplementing #1657 to use BiDi to dismiss the alert. This removes the fragility associated with `#jenkins-head-icon`.
- Adds this new BiDi implementation to `CspRule`, avoiding the need for copypasta in a bunch of tests and saving test execution time when `CspRule` is disabled.

The net result is a strict improvement over the status quo: less copypasta and less fragility because the codebase now has no references to `#jenkins-head-icon`.

There is likely still more cleanup that can be done; for example, https://github.com/jenkinsci/acceptance-test-harness/blob/40f6a5807ddebb81591aedcbb9d8992b7d88b1a2/src/main/java/org/jenkinsci/test/acceptance/FallbackConfig.java#L364-L372 still looks suspicious, and the sleep statement in `FormValidationTest` (which has been there for many, many years) is probably suspicious as well. I am explicitly not tackling these additional cleanups in this PR.

### Testing done

PR build (which has `CspRule` enabled)

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/FallbackConfig.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/FallbackConfig.java
@@ -87,7 +87,6 @@ public class FallbackConfig extends AbstractModule {
 
     public static final String DOM_MAX_SCRIPT_RUN_TIME = "dom.max_script_run_time";
     public static final String DOM_MAX_CHROME_SCRIPT_RUN_TIME = "dom.max_chrome_script_run_time";
-    public static final String DOM_DISABLE_BEFOREUNLOAD = "dom.disable_beforeunload";
     public static final int PAGE_LOAD_TIMEOUT = 30;
     public static final int IMPLICIT_WAIT_TIMEOUT = 1;
 
@@ -184,8 +183,6 @@ public class FallbackConfig extends AbstractModule {
                 DOM_MAX_SCRIPT_RUN_TIME, (int) getElasticTime().seconds(600));
         firefoxOptions.addPreference(
                 DOM_MAX_CHROME_SCRIPT_RUN_TIME, (int) getElasticTime().seconds(600));
-        firefoxOptions.addPreference(
-                DOM_DISABLE_BEFOREUNLOAD, false); // TODO remove when we require Firefox 129 or newer
         firefoxOptions.enableBiDi();
         if (HarRecorder.isCaptureHarEnabled()) {
             firefoxOptions.setProxy(createSeleniumProxy(testName.get()));
--- a/src/main/java/org/jenkinsci/test/acceptance/junit/CspRule.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/junit/CspRule.java
@@ -41,7 +41,7 @@ public final class CspRule implements TestRule {
                             && !isSkipped()
                             && !d.getTestClass().getName().equals("plugins.ArtifactoryPluginTest")) {
                         ContentSecurityPolicyReport csp = new ContentSecurityPolicyReport(jenkins);
-                        csp.open();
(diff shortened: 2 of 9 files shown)


src/main/java/org/jenkinsci/test/acceptance/FallbackConfig.java
```diff
@@ -184,8 +183,6 @@ private FirefoxOptions buildFirefoxOptions(TestName testName) throws IOException
                 DOM_MAX_SCRIPT_RUN_TIME, (int) getElasticTime().seconds(600));
         firefoxOptions.addPreference(
                 DOM_MAX_CHROME_SCRIPT_RUN_TIME, (int) getElasticTime().seconds(600));
-        firefoxOptions.addPreference(
-                DOM_DISABLE_BEFOREUNLOAD, false); // TODO remove when we require Firefox 129 or newer
```
basil, line 188: https://www.jenkins.io/doc/book/platform-information/support-policy-web-browsers/ version 129 is from last August, so seems safe to drop support for it now. We now fully rely on the officially supported BiDi technology (enabled on the next line) rather than this undocumented preference.

src/main/java/org/jenkinsci/test/acceptance/junit/CspRule.java
```diff
@@ -41,7 +41,7 @@ public void evaluate() throws Throwable {
                             && !isSkipped()
                             && !d.getTestClass().getName().equals("plugins.ArtifactoryPluginTest")) {
                         ContentSecurityPolicyReport csp = new ContentSecurityPolicyReport(jenkins);
-                        csp.open();
+                        jenkins.runThenHandleUserPrompt(() -> csp.open());
```
basil, line 44: Avoids the need for copypasta by dismissing the user prompt inside of `CspRule`. There are likely other ways of doing this; perhaps as of https://github.com/mozilla/geckodriver/issues/1151 `driver.quit()` could be used instead, although that means that the driver has to be reopened and it wasn't immediately obvious to me how to implement that. Since this existing PR is working, I didn't see a need to spend more time thinking about a potential alternative implementation with no clear benefit.

src/test/java/core/FormValidationTest.java
```diff
@@ -75,8 +74,7 @@ private void jsValidation() {
     }
 
     private void navigateAway() {
-        jenkins.runThenConfirmAlert(
-                () -> driver.findElement(By.id("jenkins-head-icon")).click());
+        jenkins.runThenHandleUserPrompt(() -> jenkins.open());
```
basil, line 77: This sleep below remains suspicious, but that code has been here for many, many years.

src/test/java/core/PublisherOrderTest.java
```diff
@@ -60,13 +59,5 @@ public void testUnordered() {
...
         JUnitPublisher junit = upstream.addPublisher(JUnitPublisher.class);
         fingerprint.targets.set("yetanother");
-
-        /*
-         * Navigate back to the dashboard first to dismiss the alert so that CspRule can check for violations (see
-         * FormValidationTest).
-         */
-        jenkins.runThenConfirmAlert(
-                () -> driver.findElement(By.id("jenkins-head-icon")).click());
-        sleep(1000);
```
basil, line 70: Here and below, removing the copypasta added when `CspRule` was enabled.

timja (approved):

PR #2001, opened 2025-05-12
Propagate cleanup failures

basil:
Recently I spent quite a bit of time debugging a resource leak that would have been far more obvious if the cleanup handler propagated exceptions rather than silently swallowing them.

### Testing done

Tested a few scenarios:

- No exception
- Skipped test
- Exception in cleanup handler but not in test
- Exception in test but not in cleanup handler
- Exception in both test and cleanup handler (the latter was added as a suppressed exception)

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/guice/Cleaner.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/guice/Cleaner.java
@@ -2,7 +2,9 @@ package org.jenkinsci.test.acceptance.guice;
 
 import java.io.Closeable;
 import java.util.ArrayDeque;
+import java.util.ArrayList;
 import java.util.Deque;
+import java.util.List;
 import java.util.concurrent.Callable;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -66,15 +68,18 @@ public class Cleaner {
         });
     }
 
-    public void performCleanUp() {
+    public List<Throwable> performCleanUp() {
         LOGGER.info("Performing cleanup tasks in order: " + tasks);
+        List<Throwable> errors = new ArrayList<>();
         for (Statement task : tasks) {
             try {
                 task.evaluate();
             } catch (Throwable t) {
                 LOGGER.log(Level.SEVERE, task + " failed", t);
+                errors.add(t);
             }
         }
         tasks.clear();
+        return errors;
     }
 }
--- a/src/main/java/org/jenkinsci/test/acceptance/guice/World.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/guice/World.java
@@ -8,8 +8,10 @@ import com.google.inject.Injector;
 import com.google.inject.Module;
 import com.google.inject.util.Modules;
 import jakarta.inject.Singleton;
+import java.util.List;
 import org.jenkinsci.test.acceptance.Config;
 import org.jenkinsci.test.acceptance.FallbackConfig;
+import org.junit.runners.model.MultipleFailureException;
 
 /**
  * Holder of the Guice world for running tests. Singleton.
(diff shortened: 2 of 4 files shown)


src/main/java/org/jenkinsci/test/acceptance/guice/Cleaner.java
```diff
@@ -66,15 +68,18 @@ public String toString() {
...
 
-    public void performCleanUp() {
+    public List<Throwable> performCleanUp() {
         LOGGER.info("Performing cleanup tasks in order: " + tasks);
+        List<Throwable> errors = new ArrayList<>();
         for (Statement task : tasks) {
             try {
                 task.evaluate();
             } catch (Throwable t) {
                 LOGGER.log(Level.SEVERE, task + " failed", t);
```
basil, line 78: Normally it's bad practice to both log and throw, but in this case the method is also called from the world cleaner thread and connection handling thread, both of which have no way to propagate errors, so logging is still useful here.

src/main/java/org/jenkinsci/test/acceptance/junit/JenkinsAcceptanceTestRule.java
```diff
@@ -53,14 +53,28 @@ public void evaluate() throws Throwable {
 
                 injector.injectMembers(this);
 
+                Throwable throwable = null;
                 try {
                     decorateWithRules(base).evaluate();
                 } catch (AssumptionViolatedException e) {
                     System.out.printf("Skipping %s%n", description.getDisplayName());
                     e.printStackTrace();
```
basil, line 61: Normally it's bad practice to both log and throw, but in this case the method is also called from the world cleaner thread and connection handling thread, both of which have no way to propagate errors, so logging is still useful here.

src/main/java/org/jenkinsci/test/acceptance/guice/World.java
```diff
@@ -62,14 +64,16 @@ private void startTestScope(Injector i, String testName) {
...
+        List<Throwable> errors = injector.getInstance(TestCleaner.class).performCleanUp();
         injector.getInstance(TestLifecycle.class).endTestScope();
 
         for (SubWorld sw : subworlds.list(injector)) {
-            sw.injector.getInstance(TestCleaner.class).performCleanUp();
+            errors.addAll(sw.injector.getInstance(TestCleaner.class).performCleanUp());
             sw.injector.getInstance(TestLifecycle.class).endTestScope();
         }
+
+        MultipleFailureException.assertEmpty(errors);
```
basil, line 76: This is one of three callers of `performCleanup`. Here we propagate the errors. The other two callers are the world cleaner thread and connection handling thread, both of which have no way of propagating the errors and as such ignore the return value of `performCleanup()`, unlike what we are doing here.

src/main/java/org/jenkinsci/test/acceptance/junit/JenkinsAcceptanceTestRule.java
```diff
@@ -53,14 +53,28 @@ public void evaluate() throws Throwable {
...
+                    throwable = t;
                 } finally {
-                    world.endTestScope();
+                    try {
+                        world.endTestScope();
+                    } catch (Throwable t) {
+                        if (throwable == null) {
+                            throwable = t;
+                        } else {
+                            throwable.addSuppressed(t);
```
basil, line 72: Using `.addSuppresed` here to ensure that the visual display of the error focuses on the testcase error rather than the cleanup error.

scherler (approved):

jtnord (commented):
In principal all looks good, but I would like to run this through windows first, so we do not end up with a huge list of failures caused by failure to delete some temp files as they are still locked (without at least knowing the scope of the issue if indeed it is an issue).
I will run that locally overnight tonight.

jtnord:
> In principal all looks good, but I would like to run this through windows first, so we do not end up with a huge list of failures caused by failure to delete some temp files as they are still locked (without at least knowing the scope of the issue if indeed it is an issue). I will run that locally overnight tonight.

Lots of failures but nothing that is related to this code, so this is good to go.

timja (approved):

PR #2010, opened 2025-05-20
Remove Firefox from image; run with Docker Selenium image and Docker Compose

### LINKED ISSUE #1322: ATH container is broken on AArch64

basil:
Reasons to do this:

- Slim down size of image
- Always use a Firefox version that is known to be compatible with Selenium
- One fewer version to update (Selenium version, not Selenium & Firefox)
- Matches CloudBees configuration for more unified testing lifecycle

This will require Jenkins core changes to adapt. I will prepare those once this is finished.

### Testing done

Ran a local build and observed the results with `vncviewer`.

Fixes #1322


basil:
Marking this as ready for review since it is close to done, even though there are still a few failing tests.

@jtnord Any idea these might be failing?

```
plugins.AuditTrailPluginTest#trail_should_contain_logged_events
plugins.LogParserTest#testErrorReporting
plugins.LogParserTest#testLinksAndColor
plugins.LogParserTest#trendVisible
plugins.SshSlavesPluginTest#customJavaPath
```

Would appreciate help if anyone wants to work on finishing this PR together.

basil:
Looks like the recordings of failed tests are missing too...

src/main/java/org/jenkinsci/test/acceptance/po/Control.java
```diff
@@ -155,14 +155,7 @@ public String get() {
...
     public void set(@Nullable String text) {
-        // if the text is longer than 255 characters, use the high throughput variant
-        if (text != null && text.length() > 255) {
-            setAtOnce(text);
-        } else {
-            WebElement e = resolve();
-            e.clear();
-            e.sendKeys(StringUtils.defaultString(text));
-        }
+        setAtOnce(text);
```
basil, line 158: Somehow `sendKeys()` is busted and sends something other than the text, apparently a bogus path in `/tmp`. This wasn't exposed before this PR because the path was above 255 characters, triggering the "high throughput variant" that avoids the bug.

jtnord:
> ```
> plugins.SshSlavesPluginTest#customJavaPath
> ```


`javaPath='/tmp/session8049123842935935202ac721f97-6f93-472f-bec1-e30dff8a5d99/upload2347917722228506155file/java'` 

sounds like it is detecting an entered path as a local file and trying to convert it suitable for the remote end. and copying the file to the selenium container.

Check https://github.com/jenkinsci/acceptance-test-harness/blob/master/src/main/java/org/jenkinsci/test/acceptance/selenium/UselessFileDetectorReplacement.java

basil:
Thanks @jtnord; that worked, and I filed #2016 to extract those fixes into a smaller commit.

basil:
I fixed the recordings by using `DISPLAY=firefox:99:0`. In the future we can probably rip out Monte completely and use Selenium's video recording feature, but this preserves the status quo for now.

At this point I can't see any remaining issues, so would be great to get some eyes on this from ATH stakeholders/users @jtnord @timja @uhafner

CC @dduportal since this is a major operational change, but I don't think there should be any direct Ops team impact

timja (approved):
Looks fine to me as long as it works, I very rarely use this setup.

(normally just use chrome running tests directly from my IDE)

Jenkinsfile
```diff
@@ -140,24 +133,14 @@ for (int i = 0; i < splits.size(); i++) {
...
-                          set-java.sh ${jdk}
-                          # Ensure that Jenkins node setup does not influence the container java setup
-                          unset JAVA_HOME
-                          eval \$(vnc.sh)
-                          java -version
-                          mvn -v
-                          run.sh ${browser} ${jenkinsVersion} -Dmaven.repo.local=${WORKSPACE_TMP}/m2repo -Dmaven.test.failure.ignore=true -Dcsp.rule -DforkCount=1 -B
-                          cp --verbose target/surefire-reports/TEST-*.xml /reports
-                          """
+                      sh "./ci.sh ${jdk} ${browser} ${jenkinsVersion}"
```
timja, line 136: It would be easier to read the build logs if the building of the container was separated out from the running of it.

There's 2300 lines from building of the image.
I only care about that when it fails generally and if it was in another step it would be cleaner imo

basil: Sure, but that problem also exists before this PR.

timja: Not from what I can tell, the image was either built separately or reusing and existing built image.

This PR changes it to use `ath-container.sh` that wasn't used before in CI.

basil: I don't understand what "built separately or reused and existing built image" means.

When the `build-image` label was used before this PR, an image was built, also resulting in many lines of log output.

timja: Before this change:
https://ci.jenkins.io/job/Core/job/acceptance-test-harness/job/PR-2019/1/pipeline-overview/?selected-node=108

image

image

After this change:
image

-----

Basically there's a lot more log output to go through, that is introduced in this PR

basil: > Basically there's a lot more log output to go through in a single step rather than before it was in a separate docker build step, that is introduced in this PR

Please read my previous comment. That has been addressed in the latest revision.

timja: Looks good thanks

timja (approved):

docker-compose.yml
```diff
@@ -0,0 +1,63 @@
...
+      - ${DOCKER_GID}
+    image: jenkins/ath:latest
+    networks:
+      - ath-network
+    ports:
+      - 5005:5005  # Maven debug port
+    shm_size: 2g
+    user: ath-user
+    volumes:
+      - ${PWD}:/home/ath-user/sources
```
jtnord, line 50: I think this causes a conflict with the diagnostic files, as the path to them would be `/home/ath-user/soruces/.....` so they would not be found by Jenkins when recording unit tests (unless the path in the agent ($PWD) is also /home/ath-user/soruces/? (or I'm missing something that edits the files after the fact?) 

if we infact are running in `/home/ath-user/soruces/` that would seem to be brittle.

As all tests are currently passing it's not easy to see if this is an issue or not.

jtnord (commented):
not sure but I think CI may have an issue recording attachments.

jtnord (approved):

basil:
@jtnord I updated the latest revision to fix this properly, using absolute paths. Please take a look again. It is being tested in #2018

basil:
> I updated the latest revision to fix this properly, using absolute paths. Please take a look again. It is being tested in #2018

Changes are to `ci.sh`:

```
if [[ -d target/surefire-reports ]]; then
	find target/surefire-reports -type f -name 'TEST-*.xml' -print0 |
		xargs -0 sed -i 's!\[\[ATTACHMENT|/home/ath-user/sources/target\(/[^]]*\)\]\]![[ATTACHMENT|'"$PWD"'/target\1]]!g'
else
	echo "No test results to be saved"k
fi
```

basil:
The code looks fine now, but the Java 17 / LTS runs seem to be hanging, and I can't see why without access to the box.

basil:
The 30-second timeout just postponed the hang to closing the Selenium session. Looks like somehow the combination of Java 17, Monte, and a remote X11 display wedges Selenium sometimes, so I am just turning off the Monte recorder on Java 17.

PR #2018, opened 2025-05-21
Test with Chrome

basil:
Checking to see if the test suite works with Chrome or not.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -42,7 +42,7 @@ def axes = [
   jenkinsVersions: ['lts', 'latest'],
   platforms: ['linux'],
   jdks: [17, 21],
-  browsers: ['firefox'],
+  browsers: ['chrome', 'firefox'],
 ]
 
 stage('Record builds and sessions') {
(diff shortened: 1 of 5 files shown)


basil:
Looks like some 30-odd failures when running in Chrome

@timja You mentioned you run with Chrome locally. Do you see the same problems there or are they unique to this CI build?

timja:
> @timja You mentioned you run with Chrome locally. Do you see the same problems there or are they unique to this CI build?

I tried the test:

`plugins.OicAuthPluginTest.loginAndLogoutInKeycloakIsReflectedInJenkins failed`

It doesn't fail with exactly the same error but yes it does fail with a similar one:

```
java.lang.AssertionError: User has logged in Jenkins
Expected: is "1dbcf113-4f8b-4908-b84e-992b348a1bbe"
     but: was null
Expected :is "1dbcf113-4f8b-4908-b84e-992b348a1bbe"
Actual   :null
<Click to see difference>
```

Pausing in a debugger shows its on the login page when its expected to be logged in:
image

basil:
Tracked in #2025

PR #2022, opened 2025-05-23
Record videos with Selenium instead of Monte

### LINKED ISSUE #1041: Screen recording can't be opened natively on Mac
### Jenkins and plugins versions report

-

### What Operating System are you using (both controller, and any agents involved in the problem)?

-

### Reproduction steps

Open a ui-recording.mov from a test result
If you're on mac it will default to quicktime player, it fails to open

### Expected Results

It opens

### Actual Results

It doesn't

### Anything else?

I'm using VLC to open them currently which works but it's the only thing I need VLC for these days...

Code is around here:
https://github.com/jenkinsci/acceptance-test-harness/blob/200aafd9410b487aa184c5d92c639c3baacdc09d/src/main/java/org/jenkinsci/test/acceptance/recorder/TestRecorderRule.java#L72-L87

Found a blog post that's newer than our code here which may have some updated format:
https://www.ontestautomation.com/creating-a-video-capture-of-your-selenium-tests-using-monte-screen-recorder/

basil:
Fixes #1041

In #2010 we observed that Monte was crashing Selenium on Java 17, so we turned off recording there. This PR removes that workaround and re-enables recording everywhere by moving from Monte to Selenium's native recording system. This also fixes #1041.

See the docs here: https://github.com/SeleniumHQ/docker-selenium?tab=readme-ov-file#video-recording

This PR does not remove Monte, since CloudBees is still using it. But when `VIDEO_FOLDER` is set we now use the native Selenium system. This triggers writing out a recording when the Selenium WebDriver is closed, which we then fish out and move to where Monte would have placed it. Once CloudBees switches to this new system, the Monte code can be deleted.

### Testing done

Ensured that the videos could be opened on a macOS system.
Tested with failing tests in #2018

### Submitter checklist

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -11,15 +11,9 @@ if [[ -z ${DOCKER_GID:-} ]]; then
 	export DOCKER_GID
 fi
 
-RECORDER=failuresOnly
-if ((jdk == 17)); then
-	# Crashes Monte when used with a remote X11 display
-	RECORDER=off
-fi
-
 trap 'docker-compose kill && docker-compose down' EXIT
 
-docker-compose run -e "MAVEN_ARGS=${MAVEN_ARGS}" -e "RECORDER=${RECORDER}" --name mvn -T --rm -v "${MAVEN_SETTINGS}:${MAVEN_SETTINGS}" mvn bash -s <<-INSIDE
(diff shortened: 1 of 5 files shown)


timja (approved):
I tested viewing a video from the Chrome PR and I was able to view it directly in the browser, this was not possible previously.

Thanks!!

PR #2140, opened 2025-08-19
Simplify scrolling implementation

basil:
As observed by @jtnord, the scrolling implementation is needlessly complicated and could be some 28 lines shorter with less fragility. However, I don't think this is the cause of our recent Firefox upgrade issues.

### Testing done

CI build

### Submitter checklist


src/main/java/org/jenkinsci/test/acceptance/selenium/Scroller.java
```diff
@@ -141,16 +137,11 @@ public void scrollIntoView(WebElement e) {
...
-        final int eXCoord = element.getLocation().getX();
-        final String id = element.getAttribute("id");
         final JavascriptExecutor executor = (JavascriptExecutor) driver;
         // Wait until web element is successfully scrolled.
         try {
-            new Wait<>(Boolean.TRUE)
-                    .withTimeout(Duration.ofSeconds(5)) // Wall-clock time
-                    .until(() -> (Boolean) executor.executeScript(scrollJs, eYCoord, eXCoord, id));
-        } catch (TimeoutException ex) {
+            executor.executeScript("arguments[0].scrollIntoView({block:'center', inline:'nearest'});", element);
```
basil: Aligns the element vertically at the center of the scrollable container, so we don't need to worry about headers or footers or the like and fragile logic to do coordinate arithmetic. This just lets the browser deal with it.

src/main/java/org/jenkinsci/test/acceptance/selenium/Scroller.java
```diff
@@ -141,16 +137,11 @@ public void scrollIntoView(WebElement e) {
...
+        } catch (JavascriptException ex) {
```
basil: Tested by changing the above line to nonsense that this exception was indeed thrown. So if something funky does happen in the call to `scrollIntoView`, we would enter the `catch` block and print the stack trace.

src/main/java/org/jenkinsci/test/acceptance/selenium/Scroller.java
jtnord: so the issue here and the JS in general is that IIUC scrolling is async - esp with smooth scrolling and that this can return before the scrolling is actually completed.
This is where I *expect* (but have not validated) that the selenium API to do the right thing.

e.g. old but possibly relevant -> https://stackoverflow.com/questions/46795955/how-to-know-scroll-to-element-is-done-in-javascript

jtnord: e.g. the following script on a largish page https://w3c.github.io/webdriver-bidi/

```javascript
if (window.onscrollend !== undefined) {
  window.addEventListener(
    "scrollend",
    (evt) => console.log("Scroll Event Complete"),
    { once: false }
  );
} else {
  console.warn("Your browser doesn't support the scrollend event");
}

function doScollingTest() {
  we = document.getElementById("issues-index");
  console.log("About to Scroll into View");
  we.scrollIntoView({block:'center', inline:'nearest'})
  console.log("Call to scrollIntoView returned");
}

doScrollingTest()
```

produces the following:

```
doScollingTest()
VM545:13 About to Scroll into View
VM545:15 Call to scrollIntoView  returned
undefined
VM545:4 Scroll Event Complete
```

Thus the call to scroll does appear to returns before the scroll is complete.

[`scrollEnd`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollend_event#browser_compatibility) is not supported by safari, but is by chrome and firefox, so maybe we can use this if I can not get the native API to work?

Vlatombe: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView#behavior
Maybe adding `behavior: "instant"` could help with the timing issue (no guarantee though)

basil: While that may be a problem, I don't think this PR is making things worse than the status quo, and it is improving them over the status quo by simplifying the code. I still don't think this is the root cause of the Selenium upgrade issue. As observed there, when the tests were failing after the Selenium upgrade they weren't failing because the scrolling was happening too late; they were failing because the scrolling wasn't happening at all. That is consistent with an explanation of picking the wrong element rather than with an explanation of scrolling being async.

basil: ⇒ commit 1f1735a4d

PR #2137, opened 2025-08-19
Selenium bump

jtnord:
replacing scroller.js with native selenium code.  https://github.com/jenkinsci/acceptance-test-harness/pull/2136#discussion_r2284566716

fixes #2129  (maybe!)

### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

PR #2149, opened 2025-08-21
Correctly set timeouts

basil:
Extracted from #2137 by @jtnord


jtnord (approved):
TBH I'm not sure if this is needed or not, we set them in a call to driver.manage().timeouts() and I added this just incase to see.

Won't harm, but could add code where it's not actually needed

PR #2150, opened 2025-08-21
Remove `scroller.js` and let Selenium do its thing

basil:
Extracted from https://github.com/jenkinsci/acceptance-test-harness/pull/2137 by @jtnord

Selenium natively handles scrolling elements into view before performing actions like clicking or sending text, as specified in the WebDriver standard.

[WebDriver Specification](https://www.w3.org/TR/webdriver2/#element-click):

> The [Element Click](https://www.w3.org/TR/webdriver2/#dfn-element-click) [command](https://www.w3.org/TR/webdriver2/#dfn-commands) [scrolls into view](https://www.w3.org/TR/webdriver2/#dfn-scrolls-into-view) the [element](https://dom.spec.whatwg.org/#concept-element) if it is not already [pointer-interactable](https://www.w3.org/TR/webdriver2/#dfn-pointer-interactable), and clicks its [in-view center point](https://www.w3.org/TR/webdriver2/#dfn-center-point).


jtnord (approved):

basil:
@jtnord This seems like a prerequisite for #2137. But this PR won't pass as-is; there is still a problem with `GitPluginTest`. Somehow we are clicking on the Save button and that causes form validation to run on the field we just modified, which is interfering with the save operation. I "fixed" this in commit dec09ef4964684dc78d1c7e7097960d83aad4527 but sending a tab key to do the form validation sooner, but not sure if that is correct or not. The correct fix should probably be tested in this PR and then merged separately as a prerequisite to this PR.

basil:
That "fix" broke a bunch of other stuff, so clearly it wasn't the right fix.

PR #2151, opened 2025-08-21
Remove Groovy configuration injection support

basil:
https://github.com/jenkinsci/lib-groovy-guice-binder hasn't been maintained in 11 years and I don't think we're actively using this feature anywhere. Seems simpler in the long term to delete this and archive that repository rather than attempt to maintain it.

CC @jtnord @uhafner to make sure nobody is actually using this

### Testing done

Ran one test locally. The CI build should very that we're not using this anywhere in our official infrastructure.

### Submitter checklist


jtnord:
CloudBees did used to use this for iirc the operations center suite (or the pse suite) , and I cannot currently check if we still do.
@Vlatombe should be able to confirm

uhafner:
No, I do not use this feature.

timja (approved):

basil:
Actually it looks like CloudBees uses this to set `uploadPlugins=false` in one place, but an entire Groovy subsystem for that seems overkill. And actually it seems to have been deprecated in commit 2c9d6f0507cbe6766b230045da132fa86bff8a4d and is off by default? So actually it might be a good thing to drop support for this, since it means we'll stop consuming the Groovy that tries to set `uploadPlugins` to its (already default value of) false and thus we can then remove `uploadPlugins` after this PR. Am I misreading this?

PR #2166, opened 2025-09-04
wait for the notification to disappear

jtnord:
observed that scheduling a build did not wait for a notification to happen and disappear.  in rare cases the notifacation can obscure other things that we want to interact with, so wait for the notification to appear and then be hiddent

### Testing done

None

### Submitter checklist

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -51,6 +51,7 @@
       "customType": "regex",
       "managerFilePatterns": [
         "/src/main/java/org/jenkinsci/test/acceptance/FallbackConfig\\.java/",
+        "/vars.cmd/",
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/Job.java
```diff
@@ -348,7 +349,9 @@ public Build scheduleBuild(Map<String, ?> params) {
                 throw new Error(e);
             }
         }
-
+        // the notification bar can place itslef over other elements
```
jtnord: Needs to move to the non parameter branch as parameterized builds are a separate page which redirects back

PR #2178, opened 2025-09-16
[JENKINS-76114] Test "All" and "Suggested" buttons from the plugin configuration wizard

jeromepochat:
This tests that "All" and "Suggested" button from plugin setup wizard are clickable ("None" link is already tested).

image

### Testing done

Manual tests were performed as part of https://github.com/jenkinsci/jenkins/pull/11070.

This extends coverage of the existing automated test. I tested locally:

```sh
JENKINS_VERSION=2.526 mvn test -Dtest=InstallWizardTest#wizardInstallCustomPluginsTest
```

:warning: This is failing with 2.527 => https://github.com/jenkinsci/jenkins/pull/11070 fixes the issue.

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/WizardCustomizeJenkins.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/WizardCustomizeJenkins.java
@@ -99,7 +99,6 @@ public class WizardCustomizeJenkins extends PageObject {
      * @deprecated use {@link #selectNone()} instead.
      */
     @Deprecated
-
     public void deselectAll() {
         selectNone();
     }


src/main/java/org/jenkinsci/test/acceptance/po/WizardCustomizeJenkins.java
```diff
@@ -94,8 +94,26 @@ public void startInstall() {
         clickButton("Install");
     }
 
-    public void deselectAll() {
+    public void selectAll() {
+        clickLink("All");
+        elasticSleep(200);
```
jtnord: the use of elasticSleep is a sure sign that this test is flaky.
Whatever you waiting for to complete please wait for that to happen rather than just some random amount of time.

src/test/java/core/InstallWizardTest.java
```diff
@@ -90,7 +90,9 @@ public void wizardInstallCustomPluginsTest() throws IOException {
         // Customize Jenkins step
         WizardCustomizeJenkins wizardCustomize = new WizardCustomizeJenkins(jenkins);
         wizardCustomize.doSelectPluginsToInstall();
-        wizardCustomize.deselectAll();
+        wizardCustomize.selectSuggested();
+        wizardCustomize.selectAll();
+        wizardCustomize.selectNone();
```
jtnord, line 95: ?  this is clicking everything suggested (which is already checked IIRC?), then everything, then nothing, but the test only wants the nothing.  Why?

there is no point doing this as there are no assertions.  if you want to test the buttons work as expected please assert side effects.

if you want to test suggested plugins, none and all fully e2e then split this out into 3 tests and then assert that plugins are not installed, junit is installed with recommended, and some bundled but non suggested plugin is installed.

src/main/java/org/jenkinsci/test/acceptance/po/WizardCustomizeJenkins.java
```diff
@@ -94,8 +94,26 @@ public void startInstall() {
...
+    }
+
+    public void selectNone() {
         clickLink("None");
         elasticSleep(200);
```
jtnord: do not elastic sleep, wait for the number selected to be zero for example (or to change).

image

src/main/java/org/jenkinsci/test/acceptance/po/WizardCustomizeJenkins.java
```diff
@@ -94,8 +94,26 @@ public void startInstall() {
...
     }
+
+    public void selectSuggested() {
+        clickLink("Suggested");
+        elasticSleep(200);
+    }
```
jtnord, line 125: wait for the number of selected plugins to change?  (this assumes that a test would not click on this twice, but well that would be silly right!)?

jtnord (changes requested):
test would appear to be flaky and performs actions without asserting anything

src/test/java/core/InstallWizardTest.java
```diff
@@ -90,7 +90,9 @@ public void wizardInstallCustomPluginsTest() throws IOException {
...
         wizardCustomize.searchPlugin("JUnit");
         wizardCustomize.selectPlugin("junit");
         wizardCustomize.startInstall();
```
jtnord, line 98: FWIW this is a weak test, as junit is a suggested plugin - so the customisation may be completely broken.  (there is also no check that the plugin is installed later on!)

src/main/java/org/jenkinsci/test/acceptance/po/WizardCustomizeJenkins.java
jeromepochat: https://github.com/jenkinsci/acceptance-test-harness/pull/2178/commits/2763a52f06cfdd13143cd424bc1efd36153e743e

src/test/java/core/InstallWizardTest.java
jeromepochat: The test needs "none". I introduced clicks on "all" and "suggested" to check that links are displayed and clickable.

https://github.com/jenkinsci/acceptance-test-harness/pull/2178/commits/2763a52f06cfdd13143cd424bc1efd36153e743e adds some assertions about plugin list after clicks.

src/main/java/org/jenkinsci/test/acceptance/po/WizardCustomizeJenkins.java
```diff
@@ -87,15 +88,62 @@ public void searchPlugin(String searchSring) {
...
 
     public void startInstall() {
         clickButton("Install");
     }
 
+    /**
+     * @deprecated use {@link #selectNone()} instead.
+     */
+    @Deprecated
+
```
jtnord:
```suggestion
```

src/main/java/org/jenkinsci/test/acceptance/po/WizardCustomizeJenkins.java
```diff
@@ -87,15 +88,61 @@ public void searchPlugin(String searchSring) {
...
+    private boolean isSelected(String pluginKey) {
+        return driver
+                .findElements(by.xpath(
+                        "//div[contains(@class, 'plugins-for-category')]/div[contains(@class, 'plugin') and contains(@class, 'selected') and contains(@class, '%s')]",
+                        pluginKey))
+                .stream()
+                .findFirst()
+                .isPresent();
+    }
```
jtnord, line 135: @jeromepochat this fails for some plugins.  e.g.  `workflow-aggregator`

can be reproduced with

```diff
diff --git a/src/test/java/core/InstallWizardTest.java b/src/test/java/core/InstallWizardTest.java
index 2a8507cb..6d3da103 100644
--- a/src/test/java/core/InstallWizardTest.java
+++ b/src/test/java/core/InstallWizardTest.java
@@ -93,8 +93,8 @@ public class InstallWizardTest extends AbstractJUnitTest {
         wizardCustomize.selectSuggested();
         wizardCustomize.selectAll();
         wizardCustomize.selectNone();
-        wizardCustomize.searchPlugin("JUnit");
-        wizardCustomize.selectPlugin("junit");
+        wizardCustomize.searchPlugin("Pipeline");
+        wizardCustomize.selectPlugin("workflow-aggregator");
         wizardCustomize.startInstall();

         wizardCustomize.shouldFinishInstallSuccessfully();
```

PR #2199, opened 2025-10-09
Introduce TestResult PO and update junit (plugin) related code

jtnord:
* Introduce the ability to follow an `Action` by link text
* Intoduce a `TestResult` `PageObject` to interact with the `junit` plugin `testResults` page
* fixup tests following https://github.com/jenkinsci/junit-plugin/pull/727 which changed the link for the test results (and other things).

### Testing done

run locally with the addition of #2202

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/junit/TestReport.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/junit/TestReport.java
@@ -54,9 +54,8 @@ public class TestReport extends Action {
             // this is a javascript expand
             WebElement link = row.findElement(By.tagName("a"));
             link.click();
-            WebElement details = waitFor(row)
-                    .ignoring(NoSuchElementException.class)
-                    .until(r -> r.findElement(By.xpath("following-sibling::*[1]")));
+            WebElement details =
+                    waitFor(row).ignoring(NoSuchElementException.class).until(TestReport::getTestDetailsRow);
             found = details.getText().contains(content);
             link.click(); // hide the content
             waitFor(details).until(CapybaraPortingLayerImpl::isStale);
@@ -71,5 +70,13 @@ public class TestReport extends Action {
     @Override
     public boolean isApplicable(ContainerPageObject po) {
         return po instanceof Build;
+    }
+
+    private static WebElement getTestDetailsRow(WebElement testRow) {
+        WebElement sibling = testRow.findElement(By.xpath("following-sibling::*[1]"));
+        if ("foldout-row".equals(sibling.getDomAttribute("data-type"))) {
+            return sibling;
+        }
+        return null;
     }
 }


src/test/java/plugins/XUnitPluginTest.java
```diff
@@ -33,7 +33,7 @@ public void publish_xunit_results() {
         Build b = job.startBuild().shouldSucceed();
 
         b.open();
-        clickLink("Test Result");
+        clickLink("Tests");
         assertThat(driver, Matchers.hasContent("1 failures"));
```
jtnord: @janfaracik (@timja) this is now harder to test than it should be.  
the page content is merely
```
Jenkins Header....

Tests 1  1   Took 7 ms
All Failed Tests
Name | Age | Duration
testApp com.simple.project.AppTest | 2 | 7 ms

All Tests
Package | Failed |Skipped |Passed |Total |Duration
com.simple.project |1 |0 |0 |1 |7 ms

REST API|Jenkins 2.516.3
```

The pill for the failing is not distinguised in any machine suitable way - nor does the text say failing (relying purely on the colour of the pill and a `x`.  (I guess the same would apply for passed and skipped, and the pills are missing if the value is zero IIUC?

Whilst the ATH could be hacked I think that the junit plugin as this is a new UI should be written in a test friendly manner - either using specific `data-testid`s for the various parts of including the words "passed" "skipped"  and "failed" in the pills.   WDYT?

jtnord: The raw HTML is (and there is nothing for passing/skipped)
```
<div tooltip="1 failing" class="jp-pill jenkins-!-error-color" title="1 failing">
  <svg xmlns="http://www.w3.org/2000/svg" aria-hidden="true" viewBox="0 0 512 512">
    <title></title>
    <path d="M368 368L144 144M368 144L144 368" fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="32"></path>
  </svg>
  1
</div>
```

timja: There is information available in the title / tooltip that can be used to identify it.

image

image

If its too difficult then definitely the plugin can be modified to make it easier to test in ATH.


It does lead me to the page title and total test count could have some visually hidden text to give a bit more context on what they are:

image

jtnord: > There is information available in the title / tooltip that can be used to identify it.

sure in this specific example (a bad one) I can use `find(by.xpath("//div[@title='1 failing']"));` 
but this is only half the info - where is the info that there are **zero** passing tests or **zero** skipped tests?
Some tests are asserting this and I beleive this is valid and checking for the absence of some text is bad (I beleieve it is bad UX to hide the fact that something would have zero passing tests, but show if there is one or more passing)

timja: I don't think the UI should include 0's its unnecessary information. I would only expect a category to show if there's a value > 0.

i.e. if all are passing you don't need to show that 0 are failing and 0 are skipped 

I understand where you're coming from though.

src/main/java/org/jenkinsci/test/acceptance/po/ContainerPageObject.java
```diff
@@ -86,13 +86,23 @@ public JsonNode getJson(String queryString) {
...
+    /**
+     * Create action of this page object.
+     * Preffer to use {@link #action(Class)} where fixed paths and link text is known.
+     * @param <T> The concrete action class to create.
+     * @param type Action type to create
+     * @param path the relative path from this PageObject
+     * @param linkText optional text of the link in the sidebar.
+     * @return the newly contructed Action of type T
+     */
+    public <T extends Action> T action(Class<T> type, String path, String linkText) {
```
jtnord, line 103: not backwards compatable, but if this breaks anyone (cloudbees' tests are ok) then it is a simple fix (can just pass `null` if desired), at the same time, should be use `ActionPageObject` annotation which would mean no issues.

src/main/java/org/jenkinsci/test/acceptance/plugins/junit/TestReport.java
```diff
@@ -0,0 +1,75 @@
...
+@ActionPageObject(relativePath = "testReport", linkText = "Tests")
+public class TestReport extends Action {
+
+    public TestReport(Build build, String path, String linkText) {
+        super(build, path, linkText);
+    }
+
+    public int getTotalTestCount() {
+        // ugly hack as the junit plugin is not super friendly for testing.
+        WebElement tests = find(by.xpath("//span[class='jenkins-app-bar__subtitle']"));
```
jtnord, line 26: 🤮

src/main/java/org/jenkinsci/test/acceptance/plugins/junit/TestReport.java
```diff
@@ -0,0 +1,75 @@
...
+        return Integer.parseInt(tests.getText());
+    }
+
+    public int getSkippedTestCount() {
+        WebElement tests = getElement(by.xpath("//div[contains(@title, ' skipped')]"));
```
jtnord, line 31: 😢

src/main/java/org/jenkinsci/test/acceptance/plugins/junit/TestReport.java
```diff
@@ -0,0 +1,75 @@
...
+        boolean found = false;
+        for (WebElement row : rows) {
+            // this is a javascript expand
+            WebElement link = row.findElement(By.tagName("a"));
+            link.click();
+            WebElement details = waitFor(row)
+                    .ignoring(NoSuchElementException.class)
+                    .until(r -> r.findElement(By.xpath("following-sibling::*[1]")));
+            found = details.getText().contains(content);
+            link.click(); // hide the content
```
jtnord, line 60: we hide the section so that any later tests looking for failures will not close the expand and fail.
Not ideal in the case of a failure but good enough.

src/test/java/plugins/JUnitPluginTest.java
```diff
@@ -66,33 +64,14 @@ public void publish_parametrized_tests() {
...
 
-        assertMessage("TestNG.testScore", "expected:<42> but was:<0>");
-        assertMessage("TestNG.testScore", "expected:<42> but was:<1>");
-        assertMessage("TestNG.testScore", "expected:<42> but was:<2>");
-    }
+        testReport.assertFailureContent("JUnit.testScore[0]", "expected:<42> but was:<0>");
+        testReport.assertFailureContent("JUnit.testScore[1]", "expected:<42> but was:<1>");
+        testReport.assertFailureContent("JUnit.testScore[2]", "expected:<42> but was:<2>");
 
-    private void assertMessage(String test, String content) {
```
jtnord, line 79: migrated to the `TestRerport` PageObject

src/main/java/org/jenkinsci/test/acceptance/plugins/junit/TestReport.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public int getFailedTestCount() {
+        WebElement tests = getElement(by.xpath("//div[contains(@title, ' failing')]"));
+        if (tests == null) {
+            return 0;
+        }
+        return Integer.parseInt(tests.getText());
+    }
+
+    public void assertFailureContent(String test, String content) {
```
jtnord, line 46: migrated from `JUnitPluginTest` so it can be used from other tests.

Not the best as a PO should be for modelling the page and testing/assertions should be elsewhere but better than copy-paste and good enough for now.

src/main/java/org/jenkinsci/test/acceptance/plugins/junit/TestReport.java
```diff
@@ -0,0 +1,75 @@
...
+        // that at least one of the pages have the requested content
+
+        final List<WebElement> rows = all(by.xpath("//tr[@data='%s']", test));
+
+        boolean found = false;
+        for (WebElement row : rows) {
+            // this is a javascript expand
+            WebElement link = row.findElement(By.tagName("a"));
+            link.click();
+            WebElement details = waitFor(row)
```
jtnord: the page adds a new row below.

src/main/java/org/jenkinsci/test/acceptance/plugins/junit/TestReport.java
```diff
@@ -0,0 +1,75 @@
...
+                    .ignoring(NoSuchElementException.class)
+                    .until(r -> r.findElement(By.xpath("following-sibling::*[1]")));
```
jtnord: well this sucks.. if this is actually needed then we need to check that this is the correct `tr`.

PR #2202, opened 2025-10-09
Update copyRemoteBatch

jtnord:
The existing implementation had 2 flaws

1) the timestamp of files was whatever the source timestamp was, this
   causes errors for things like junit where the resources would be too
   old (the non windows variant does not have this issue)
2) it assumes that the Agent is running in the same host as the maven
   build, so using this for remote systems would never work.

Whilst this method works correctly for remote windows agents now, any code calling copyResource would still potentially be broken if the agent was windows yet the surefire host was on a non windows machine.  No attempt to update this has been performed, the changes where primarily to fix the first issue.

### Testing done

tested locally (windows) in the context of #2199 to show that the resources are copied and that they have an acceptable timestamp

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/envinject/EnvInjectAction.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/envinject/EnvInjectAction.java
@@ -28,11 +28,11 @@ import org.jenkinsci.test.acceptance.po.ActionPageObject;
 import org.jenkinsci.test.acceptance.po.Build;
 import org.jenkinsci.test.acceptance.po.ContainerPageObject;
 
-@ActionPageObject("injectedEnvVars")
+@ActionPageObject(relativePath = "injectedEnvVars", linkText = "Environment variables")
 public class EnvInjectAction extends Action {
 
-    public EnvInjectAction(Build build, String path) {
-        super(build, path);
+    public EnvInjectAction(Build build, String path, String linkText) {
+        super(build, path, linkText);
     }
 
     public EnvInjectAction shouldContain(String key, String value) {
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/html_publisher/HtmlReport.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/html_publisher/HtmlReport.java
@@ -31,10 +31,10 @@ import org.jenkinsci.test.acceptance.po.ActionPageObject;
 import org.jenkinsci.test.acceptance.po.ContainerPageObject;
 import org.jenkinsci.test.acceptance.po.Job;
 
-@ActionPageObject("HTML_Report")
+@ActionPageObject(relativePath = "HTML_Report", linkText = "HTML Report")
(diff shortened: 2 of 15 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/Job.java
```diff
@@ -253,13 +256,33 @@ protected String copyResourceShell(Resource resource, String fileName) {
...
-    protected String copyResourceBatch(Resource resource) {
-        String path = resource.url.getPath();
-        if (path.startsWith("/")) {
-            path = path.substring(1);
+    protected String copyResourceBatch(Resource resource, String fileName) {
+        try (InputStream in = resource.asInputStream()) {
+            ByteArrayOutputStream out = new ByteArrayOutputStream();
+
+            try (ZipOutputStream zip = new ZipOutputStream(out)) {
+                zip.putNextEntry(new ZipEntry(fileName));
```
jtnord, line 264: not sure why this strips the leading slash before, the non-windows variant did not allowing you to (badly) write into the root of the filesystem...  not sanitizing this to match the non-windows variant.

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
```diff
@@ -253,13 +256,33 @@ protected String copyResourceShell(Resource resource, String fileName) {
...
+            // additionally certutil can do not base64 decoding from stdin but needs a file.
+            String encoded = java.util.Base64.getMimeEncoder().encodeToString(out.toByteArray());
+            sb.append(encoded.lines()
+                            .map(s -> "  echo." + s)
+                            .collect(Collectors.joining("\r\n", "> ath.resource.tmp.zip.b64 (\r\n", "\r\n)")))
+                    .append("\r\n");
+            sb.append("certutil -decode ath.resource.tmp.zip.b64 ath.resource.tmp.zip")
+                    .append("\r\n");
+            // tar is the way to work natively with zip files on windows!
+            sb.append("tar -x -m -f ath.resource.tmp.zip").append("\r\n");
```
jtnord, line 279: windows ships with `BSD` tar which is different from GNU tar.  it natively handles zip files :)

`-m    Don't restore modification times` so that the files have the timestamp of the job (similar to Unix and so that junit resources will work correctly on both platforms.

The filename (including any path) is in the name of the file in the zip and this creates folders if needed too!

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
```diff
@@ -253,13 +256,33 @@ protected String copyResourceShell(Resource resource, String fileName) {
...
+            sb.append("del ath.resource.tmp.zip.b64").append("\r\n");
+            sb.append("del ath.resource.tmp.zip").append("\r\n");
```
jtnord, line 281: just cleaning up.
Whilst these files could be there as part of the build and needed - the names should be unusual and distinct enough that it does not matter for anything running tests.  Adding unique files would add an extra layer of complexity here that would be over the top.

PR #2211, opened 2025-10-22
Adapt to removal of built-in executor config from global system configuration

janfaracik:
Testing https://github.com/jenkinsci/jenkins/pull/11224

### Submitter checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
   <properties>
     <changelist>999999-SNAPSHOT</changelist>
     <spotbugs.skip>true</spotbugs.skip>
-    <jenkins.version>2.534-rc37639.dd09655b_f527</jenkins.version>
+    <jenkins.version>2.532</jenkins.version>
     <selenium.version>4.37.0</selenium.version>
     <!-- aligned with selenium -->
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/JenkinsConfig.java
```diff
@@ -11,9 +11,7 @@
 public class JenkinsConfig extends ConfigurablePageObject {
     public final Jenkins jenkins;
 
-    public final Control numExecutors = control("/jenkins-model-MasterBuildConfiguration/numExecutors");
```
jtnord, line 14: this is in use in a proprietary test suites and was removed without replacement - and it should still work on exising Jenkins versions.

If you just deprecate this and add a @link to javadoc then IDEs can automatically update allowing easier migration for consumers.  (and then could have been removed when the baseline LTS contains the upstream PR)

Allhough there is in this case no actual replacement AFAICT - `Jenkins.setExecutors(int)` does not work (Jenkins PR extends from Node).  to get to the Config page you need to force a `Slave` of name `(built-in)` which is a slightly questionable abuse.

PR #2219, opened 2025-10-27
disable implicit waiting

jtnord:
implicit waiting is a sure fire sign that the tests or page objects are flaky.  If something needs to wait for something async it should do it in a `waitFor` Doing it blindly leads to cases where it looks like the code is fine but is only working by chance on the current hardware and can fail elsewhere on different hardware with different timings.

See also https://www.selenium.dev/documentation/webdriver/waits/#implicit-waits and the warning about mixing implicit and explicit waits.

Flagged as breaking, but in reality it's only breaking something that was already broken in a flaky way.  
If you see breakages then you should add appropriate `waitFor()` around obtaining elements or otherwise wait for an asynchronous event to complete.

https://youtu.be/gyfUpOysIF8?t=2201&si=3Lug3BAGyuEwK6Yy 

### Testing done

Run the ATH here and fixed any failing tests.  (also fixedup CloudBees' internal test suite)

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/Config.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/Config.java
@@ -88,7 +88,8 @@ public class Config extends AbstractModule {
     public static final String DEVTOOLS_JSONVIEW_ENABLED = "devtools.jsonview.enabled";
     public static final String PASSWORD_MANAGER_LEAK_DETECTION = "profile.password_manager_leak_detection";
     public static final int PAGE_LOAD_TIMEOUT = 30;
-    public static final int IMPLICIT_WAIT_TIMEOUT = 0;
+    // Never attempt to configure a non zero value here (see PR for the reasons why).
+    private static final int IMPLICIT_WAIT_TIMEOUT = 0;
 
     @Override
     protected void configure() {


src/main/java/org/jenkinsci/test/acceptance/po/WizardCreateAdminUser.java
```diff
@@ -65,14 +65,10 @@ public WizardCreateAdminUser createAdminUser(String userName, String password, S
...
 
         driver.switchTo().defaultContent();
-        control(by.css(".btn-primary.save-first-user")).click();
+        control(by.css(".btn-primary.save-first-user")).clickAndWaitToBecomeStale(Duration.ofSeconds(time.seconds(5)));
         return this;
     }
 
-    public void shouldCreateTheUserSuccessfully() {
-        waitFor(hasContent("Jenkins is ready!"));
-    }
```
jtnord: 1. Jenkins does not show the "Jenkins is ready!" message after creation of the admin user, it shows the set URL dialog!
2. this worked because whilst you may think this is waiting for Jenkins to have that message, it is actually not waiting as there is no `until` and this is not using the `waitFor(Matcher, timeout)` method.  ie it is a complete no-op as far as checking anything is concerned.

src/main/java/org/jenkinsci/test/acceptance/po/FormValidation.java
```diff
@@ -88,14 +89,16 @@ public static FormValidation await(Control control, boolean silent) {
...
-            validationArea = control.waitFor().until(() -> {
-                // path to validation area is the parents sibling with class `validation-error-area`
-                String xpath = silent
-                        ? "../../div[contains(@class,'validation-error-area')]"
-                        : "../../div[contains(@class,'validation-error-area--visible')]";
-
-                return element.findElement(by.xpath(xpath));
-            });
+            validationArea = control.waitFor()
+                    .ignoring(NoSuchElementException.class)
```
jtnord, line 93: validation is certainly async so relying on the implicit wait (of 1 second) would always be potentially flaky.
Thankfully the CasC test reliably needs more than 1 second so this was an easy find :)

src/main/java/org/jenkinsci/test/acceptance/plugins/stageview/StageView.java
```diff
@@ -136,11 +138,11 @@ public List<StageViewHeadline> getStageViewHeadlines() {
...
 
     /**
-     * Returns the root eelement
+     * Returns the root element
      *
      * @return the root element
      */
     public WebElement getRootElementName() {
-        return super.find(By.id("pipeline-box"));
+        return super.find(By.id(ID_WRAPPER));
```
jtnord, line 146: there was a constant that was completely unused :-o

src/main/java/org/jenkinsci/test/acceptance/plugins/stageview/StageViewBuild.java
```diff
@@ -56,8 +56,8 @@ public List<StageViewStage> getAllStageViewItem() {
      *
      * @return specific item
      */
-    public StageViewStage getStageViewItem(int buildNumber) {
-        return this.stageViewStages.get(stageViewStages.size() - 1);
+    public StageViewStage getStageViewItem(int stageNumber) {
+        return this.stageViewStages.get(stageNumber);
```
jtnord, line 60: this has only ever worked by fluke...  there is a Test that is actually calling this and expecting a different response based on the stage, it would only pass if the last stage failed in under 1 second.

https://github.com/jenkinsci/acceptance-test-harness/blob/0f06fc2dab9247ab45ed40d7764044bc9d370c7b/src/test/java/plugins/StageViewTest.java#L146-L150

src/test/java/plugins/StageViewTest.java
```diff
@@ -143,11 +145,11 @@ public void stageViewContainsMultipleStagesWithFail() {
         job.open();
         job.getNavigationLinks();
         StageView stageView = new StageView(job, JOB_PATH);
-        String firstJob = stageView.getLatestBuild().getStageViewItem(0).toString();
```
jtnord: these are the stages not jobs (or builds).

src/test/java/plugins/StageViewTest.java
```diff
@@ -143,11 +145,11 @@ public void stageViewContainsMultipleStagesWithFail() {
...
-        String secondJob = stageView.getLatestBuild().getStageViewItem(1).toString();
+        String firstStage = stageView.getLatestBuild().getStageViewItem(0).toString();
+        String secondStage = stageView.getLatestBuild().getStageViewItem(1).toString();
         assertThat(stageView.getLatestBuild().getCssClasses(), containsString("FAILED"));
-        assertThat(firstJob, containsString("ms"));
-        assertThat(secondJob, containsString("failed"));
+        assertThat(firstStage, allOf(containsString("ms"), not(containsString("failed"))));
```
jtnord, line 151: this was previously passing erronously as the page object always returned the same stage (the failed one that is running `sh exit 1`).

On my windows machine invoking `sh` (which is the second stage) goes via either cygwin or WLS and as such it does not complete in < 1 second so this would fail (as it returned the stage with the sh not the stage with the echo).  This makes the test a little more robust by also checking that the stage is not marked as a failuire.

src/main/java/org/jenkinsci/test/acceptance/Config.java
```diff
@@ -88,7 +88,7 @@ public class Config extends AbstractModule {
     public static final String DEVTOOLS_JSONVIEW_ENABLED = "devtools.jsonview.enabled";
     public static final String PASSWORD_MANAGER_LEAK_DETECTION = "profile.password_manager_leak_detection";
     public static final int PAGE_LOAD_TIMEOUT = 30;
-    public static final int IMPLICIT_WAIT_TIMEOUT = 1;
+    public static final int IMPLICIT_WAIT_TIMEOUT = 0;
```
jglick: A comment about why we have a constant set to zero would be in order here. Or if it cannot be overridden at runtime, why not just inline?

jtnord: The default in Selenium is 0 if not configured, so we could completely remove it or I could leave it here but change it to


```suggestion
    // Never attempt to configure a non zero value here (see PR for the reasons why).
    private static final int IMPLICIT_WAIT_TIMEOUT = 0;
```

I'm in favour of the field with the comment so that no one attempts to re-add it in the future

PR #2220, opened 2025-10-28
correctly wait when setting the admin user

jtnord:
The wizard testing was flaky.  was using implicit wait of 1s (non elastic) which had been observed to randomly fail.

[Cloudbeees internal issue](https://cloudbees.atlassian.net/browse/BEE-15894)
### Testing done

extracted and tested in #2219 

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/WizardCreateAdminUser.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/WizardCreateAdminUser.java
@@ -65,12 +65,8 @@ public class WizardCreateAdminUser extends PageObject {
         }
 
         driver.switchTo().defaultContent();
-        control(by.css(".btn-primary.save-first-user")).click();
+        control(by.css(".btn-primary.save-first-user")).clickAndWaitToBecomeStale(Duration.ofSeconds(time.seconds(5)));
         return this;
-    }
-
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/WizardCreateAdminUser.java
```diff
@@ -65,14 +65,10 @@ public WizardCreateAdminUser createAdminUser(String userName, String password, S
...
         driver.switchTo().defaultContent();
-        control(by.css(".btn-primary.save-first-user")).click();
+        control(by.css(".btn-primary.save-first-user")).clickAndWaitToBecomeStale(Duration.ofSeconds(time.seconds(5)));
         return this;
     }
 
-    public void shouldCreateTheUserSuccessfully() {
-        waitFor(hasContent("Jenkins is ready!"));
-    }
-
```
jtnord, line 75: 1. Jenkins does not show the "Jenkins is ready!" message after creation of the admin user, it shows the set URL dialog!
2. this worked because whilst you may think this is waiting for Jenkins to have that message, it is actually not waiting as there is no `until` and this is not using the `waitFor(Matcher, timeout)` method.  ie it is a complete no-op as far as checking anything is concerned.

PR #2222, opened 2025-10-28
Fix and deflake StageView POs

jtnord:
StageView PO was broken and flaky

### Testing done

tested in and extracted from #2219 

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/stageview/StageView.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/stageview/StageView.java
@@ -58,6 +58,8 @@ public class StageView extends PageAreaImpl {
      * To generate and build up the whole stageview structure.
      */
     public void buildStructure() {
+        // stage view loads asynchronously
+        waitFor(by.id(ID_WRAPPER));
         List<WebElement> children = driver.findElements(By.xpath(XPATH_JOBS + "/tr"));
 
         for (WebElement e : children) {
@@ -136,11 +138,11 @@ public class StageView extends PageAreaImpl {
     }
 
     /**
-     * Returns the root eelement
+     * Returns the root element
      *
      * @return the root element
      */
     public WebElement getRootElementName() {
-        return super.find(By.id("pipeline-box"));
+        return super.find(By.id(ID_WRAPPER));
     }
 }
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/stageview/StageViewBuild.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/stageview/StageViewBuild.java
@@ -56,8 +56,8 @@ public class StageViewBuild {
      *
      * @return specific item
      */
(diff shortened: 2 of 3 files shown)


src/main/java/org/jenkinsci/test/acceptance/plugins/stageview/StageView.java
```diff
@@ -136,11 +138,11 @@ public List<StageViewHeadline> getStageViewHeadlines() {
...
 
     /**
-     * Returns the root eelement
+     * Returns the root element
      *
      * @return the root element
      */
     public WebElement getRootElementName() {
-        return super.find(By.id("pipeline-box"));
+        return super.find(By.id(ID_WRAPPER));
```
jtnord, line 146: there was a constant for this that was completely unused :-o

src/main/java/org/jenkinsci/test/acceptance/plugins/stageview/StageViewBuild.java
```diff
@@ -56,8 +56,8 @@ public List<StageViewStage> getAllStageViewItem() {
      *
      * @return specific item
      */
-    public StageViewStage getStageViewItem(int buildNumber) {
```
jtnord, line 59: this was just broken and working by fluke :)

this gets the column of a row and is not a build at all but a stage, so updated the parameter name whilst I was here.

src/test/java/plugins/StageViewTest.java
```diff
@@ -143,11 +145,11 @@ public void stageViewContainsMultipleStagesWithFail() {
         job.open();
         job.getNavigationLinks();
         StageView stageView = new StageView(job, JOB_PATH);
-        String firstJob = stageView.getLatestBuild().getStageViewItem(0).toString();
```
jtnord, line 146: its not jobs, its stages. updated so that test is clearer

src/test/java/plugins/StageViewTest.java
```diff
@@ -143,11 +145,11 @@ public void stageViewContainsMultipleStagesWithFail() {
...
-        String secondJob = stageView.getLatestBuild().getStageViewItem(1).toString();
+        String firstStage = stageView.getLatestBuild().getStageViewItem(0).toString();
+        String secondStage = stageView.getLatestBuild().getStageViewItem(1).toString();
         assertThat(stageView.getLatestBuild().getCssClasses(), containsString("FAILED"));
-        assertThat(firstJob, containsString("ms"));
-        assertThat(secondJob, containsString("failed"));
+        assertThat(firstStage, allOf(containsString("ms"), not(containsString("failed"))));
```
jtnord, line 151: this was previously passing erronously as the page object always returned the same stage (the failed one that is running `sh exit 1`).

On my windows machine invoking `sh` (which is the second stage) goes via either `cygwin` or `WSL` and as such it does not complete in < 1 second so this would fail (as it returned the stage with the `sh` not the stage with the echo). This makes the test a little more robust by also checking that the stage is not marked as a failure.

PR #2244, opened 2025-11-10
simplify Dialog handling

jtnord:
A dialog is not an alert and attempting to fallback will run the runnable twice which is a silly thing to do.

Update so that we wait for the dialog to be shown and then use the default button for progressing.
In the case of input find the input within the dialog to reduce the search scope.

### Testing done

Run with some proprietary test suites and `MatrixAuthPluginTest#helloWorld` observed no failures.

Any downstream user that was using this for handling `alert`s will need to update to use the specific alert method.

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
@@ -416,7 +416,7 @@ public class CapybaraPortingLayerImpl implements CapybaraPortingLayer {
      *
      * Executes the runnable, wait for dialog then confirm with a press on the default button
      * of the modal dialog.
-     * In case the runnable opens a classical browse alert then use #runThenHandleAlert
+     * In case the runnable opens a classical browse alert then use {@link #runThenHandleAlert}
      *
      * @param runnable the code to run that will trigger the display of a dialog
      */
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
```diff
@@ -411,51 +412,43 @@ public void check(String locator) {
...
-        }
+    public void runThenHandleInputDialog(Runnable runnable, String input) {
+        runnable.run();
+        WebElement dialog = waitFor(driver)
+                .pollingEvery(Duration.ofMillis(100))
+                .ignoring(NoSuchElementException.class)
+                .until(waitForElementAnimationToFinish(By.className("jenkins-dialog")));
+
+        dialog.findElement(by.css("dialog input")).sendKeys(input);
+        WebElement defaultBtn = dialog.findElement(by.css(".jenkins-button--primary"));
```
jglick, line 449: I hope weird CSS class names like this are stable…

src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
```diff
@@ -623,4 +616,37 @@ protected final Object find(String caption) {
...
+                    return element;
+                }
+                opacity = newOpacity;
+                rect = newRect;
+                return null;
+            }
+
+            @Override
+            public String toString() {
+                return "Annimation complete for selector: " + by;
```
jglick:
```suggestion
                return "Animation complete for selector: " + by;
```

jglick (approved):
OK?

src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java
jtnord: possibly, possibly not.  But when the upstream changes we should ensure it adds a `data-testid='default-dialog-action'` attribute so it becomes stable.

timja: `jenkins-` is part of the public API for CSS, yes it is stable`.

Issue #1461, opened 2024-01-02
plugins.GitLabPluginTest.testGitLabMultibranchPipeline is flaky


timja:
### Jenkins and plugins versions report

.

### What Operating System are you using (both controller, and any agents involved in the problem)?

.

### Reproduction steps

Noticed in https://github.com/jenkinsci/acceptance-test-harness/pull/1460/checks?check_run_id=20106547650 and many other runs

### Expected Results

passing

### Actual Results

failing

jtnord:
appears to have started in around https://ci.jenkins.io/job/Core/job/acceptance-test-harness/job/master/1033/  (for some reason this is not your normal flaky test but script-security failing to start?)



Build(deps): Bump jfrog/artifactory-oss from 7.71.8 to 7.71.9 in /src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/ArtifactoryContainer (#1456) ([details](https://ci.jenkins.io/job/Core/job/acceptance-test-harness/job/master/1033/changes#c53904812955300490c236df791c99bb22a80012) / [githubweb](https://github.com/jenkinsci/acceptance-test-harness/commit/c53904812955300490c236df791c99bb22a80012))
Build(deps): Bump gitlab/gitlab-ce from 16.6.2-ce.0 to 16.7.0-ce.0 in /src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer (#1457) ([details](https://ci.jenkins.io/job/Core/job/acceptance-test-harness/job/master/1033/changes#1fc6a8d433948a93541929289c721dd4aa354920) / [githubweb](https://github.com/jenkinsci/acceptance-test-harness/commit/1fc6a8d433948a93541929289c721dd4aa354920))
Update jenkins.version to v2.438 (#1455) ([details](https://ci.jenkins.io/job/Core/job/acceptance-test-harness/job/master/1033/changes#761b1e394788b2b5338f23e50ddbc45d65578970) / [githubweb](https://github.com/jenkinsci/acceptance-test-harness/commit/761b1e394788b2b5338f23e50ddbc45d65578970))

PR #1365, opened 2023-09-21
Fix failing gitlab test on master

car-roll:
`GitLabPluginTest.gitLabGroupFolderOrganization` consistently flaky on master CI

using `selectDropdownMenu` appears to solve this issue. See #1364 for discovery and experimentation.

### Testing done

```[tasklist]
### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue
```

PR #2285, opened 2025-12-08
Re-enable GitLabPluginTest

apuig:
Attempt on optimizing the GitLab Docker image, increased timeouts, and incorporated defensive retries. Wherever possible, use Git directly instead of GitLabApi.

Thanks to @raul-arabaolaza and others for their previous work in jenkinsci/acceptance-test-harness#1425

Aim to fix:
https://github.com/jenkinsci/acceptance-test-harness/issues/1461
https://github.com/jenkinsci/acceptance-test-harness/pull/1365

### Submitter checklist

### DIFF
--- a/src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer/Dockerfile
+++ b/src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer/Dockerfile
@@ -1,5 +1,5 @@
 # Use the official GitLab Community Edition image as the base
-FROM gitlab/gitlab-ce:18.8.1-ce.0
+FROM gitlab/gitlab-ce:18.8.3-ce.0
 
 COPY create_user.rb /usr/bin/
 COPY gitlab.rb /etc/gitlab/
--- a/pom.xml
+++ b/pom.xml
@@ -30,7 +30,7 @@
   <properties>
     <changelist>999999-SNAPSHOT</changelist>
     <spotbugs.skip>true</spotbugs.skip>
-    <jenkins.version>2.547</jenkins.version>
+    <jenkins.version>2.549</jenkins.version>
     <selenium.version>4.40.0</selenium.version>
     <!-- aligned with selenium -->
     <guava.version>33.5.0-jre</guava.version>
@@ -225,7 +225,7 @@ and
     <dependency>
       <groupId>commons-codec</groupId>
       <artifactId>commons-codec</artifactId>
-      <version>1.20.0</version>
+      <version>1.21.0</version>
     </dependency>
     <dependency>
       <groupId>commons-logging</groupId>
@@ -787,7 +787,7 @@ and
       <id>lts</id>
       <properties>
         <!--RENOVATE-LTS-->
-        <jenkins.version>2.528.3</jenkins.version>
+        <jenkins.version>2.541.1</jenkins.version>
       </properties>
     </profile>
   </profiles>
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/configuration_as_code/JcascManage.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/configuration_as_code/JcascManage.java
@@ -28,10 +28,13 @@ import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.not;
 
 import java.time.Duration;
+import org.jenkinsci.test.acceptance.po.Conditions;
 import org.jenkinsci.test.acceptance.po.Control;
 import org.jenkinsci.test.acceptance.po.FormValidation;
 import org.jenkinsci.test.acceptance.po.Jenkins;
 import org.jenkinsci.test.acceptance.po.PageObject;
+import org.openqa.selenium.By;
+import org.openqa.selenium.NoSuchElementException;
 
 public class JcascManage extends PageObject {
     public JcascManage(Jenkins jenkins) {
@@ -42,6 +45,14 @@ public class JcascManage extends PageObject {
      * Configure and apply the new source
      */
     public void configure(String path) {
+        visit("#new");
+
+        waitFor(driver)
+                .withTimeout(Duration.ofSeconds(10))
+                .pollingEvery(Duration.ofMillis(100))
+                .ignoring(NoSuchElementException.class)
+                .until(Conditions.waitForElementAnimationToFinish(By.cssSelector("dialog[open]")));
+
         Control control = control("/newSource");
         control.set(path);
         waitFor()
@@ -54,12 +65,12 @@ public class JcascManage extends PageObject {
                     return true;
                 });
 
-        clickButton("Apply new configuration");
+        clickButton("Apply configuration");
         verifySuccessfulApplication();
     }
 
     public void reload() {
-        clickButton("Reload existing configuration");
+        find(by.name("reload")).click();
         verifySuccessfulApplication();
     }
 
--- a/src/main/resources/ath-container/Dockerfile
+++ b/src/main/resources/ath-container/Dockerfile
@@ -25,8 +25,8 @@ RUN apt-get update \
   && rm -rf /var/lib/apt/lists/*
 
 # Docker installation according to https://docs.docker.com/engine/install/ubuntu/
-ARG DOCKER_BUILDX_VERSION=0.30.1
-ARG DOCKER_VERSION=29.2.0
+ARG DOCKER_BUILDX_VERSION=0.31.1
+ARG DOCKER_VERSION=29.2.1
 RUN install -m 0755 -d /etc/apt/keyrings \
   && curl -fsSL https://download.docker.com/linux/ubuntu/gpg -o /etc/apt/keyrings/docker.asc \
   && printf 'deb [arch=%s signed-by=/etc/apt/keyrings/docker.asc] https://download.docker.com/linux/ubuntu %s stable\n' "$(dpkg --print-architecture)" "$(lsb_release -cs)" > /etc/apt/sources.list.d/docker.list \
--- a/src/test/java/plugins/OicAuthPluginTest.java
+++ b/src/test/java/plugins/OicAuthPluginTest.java
@@ -45,7 +45,7 @@ public class OicAuthPluginTest extends AbstractJUnitTest {
     private static final String CLIENT = "jenkins";
 
     private static final String KEYCLOAK_IMAGE =
-            "keycloak/keycloak:26.4.7@sha256:9409c59bdfb65dbffa20b11e6f18b8abb9281d480c7ca402f51ed3d5977e6007";
+            "keycloak/keycloak:26.5.2@sha256:fb31a59deb46f746f2aaa25adc5da39ceccac4fd22d36a519562b0bf02e8df20";
 
     private KeycloakContainer keycloak;
 


src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +58,559 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
-        // create an admin user
-        privateTokenAdmin =
-                container.createUserToken(adminUserName, "arandompassword12#", "testadmin@invalid.test", "true");
-
-        // create another user
-        privateTokenUser =
-                container.createUserToken(userName, "passwordforsimpleuser12#", "testsimple@invalid.test", "false");
+        long startTime = System.currentTimeMillis();
+
+        // Keep WebDriver session alive during GitLab startup to prevent timeout
```
apuig: This was crucial for the CI runs, as the GitLab image takes nearly 10 minutes to start and, without it, a Selenium session exception occurs

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -21,12 +18,13 @@
         id = "gitlab-plugin",
         ports = {80, 443, 22})
 public class GitLabContainer extends DockerContainer {
-    protected static final String REPO_DIR = "/home/gitlab/gitlabRepo";
+    public static final int GITLAB_API_CONNECT_TIMEOUT_MS = 30_000;
+    public static final int GITLAB_API_READ_TIMEOUT_MS = 120_000;
```
jglick: FWIW I like to manipulate `java.time` types such as `Duration` internally, and only convert to/from `int` / `long` at the last moment to interoperate with older APIs. Thus for example

```java
private static final Duration READINESS_TIMEOUT = Duration.ofMinutes(10);
```

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -46,67 +44,54 @@ public String httpHost() {
...
                                 .build();
-                        HttpResponse<String> response = client.send(request, HttpResponse.BodyHandlers.ofString());
+                        var response = client.send(request, HttpResponse.BodyHandlers.ofString());
                         return response.body().contains("GitLab Community Edition");
-                    } catch (IOException ignored) {
-                        // we can not use .ignoring as this is a checked exception (even though a callable can throw
-                        // this!)
-                        return Boolean.FALSE;
+                    } catch (IOException | InterruptedException ignored) {
+                        return false;
```
jglick: IIRC there is a proper method to wait ignoring exceptions.

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -119,4 +104,41 @@ public String createUserToken(String userName, String password, String email, St
...
+                        .findFirst()
+                        .ifPresent(group -> {
+                            try {
+                                gitlabapi.getGroupApi().deleteGroup(group.getId());
+                            } catch (GitLabApiException e) {
+                                throw new RuntimeException("Failed to delete group: " + groupName, e);
+                            }
+                        });
+            } catch (Exception e) {
+                System.err.println("Failed to delete group '" + groupName + "': " + e.getMessage());
```
jglick: Rethrow as `AssertionErrror`?

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +58,560 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
-    }
-
-    public String getPrivateTokenUser() {
-        return privateTokenUser;
-    }
+    private static final int BRANCH_INDEXING_TIMEOUT_SECONDS = 120;
+    private static final int BUILD_COMPLETION_TIMEOUT_SECONDS = 120;
+    private static final int JOB_CREATION_TIMEOUT_SECONDS = 30;
+    private static final int GITLAB_API_RETRY_TIMEOUT_SECONDS = 30;
+    private static final int POLLING_INTERVAL_SECONDS = 10;
```
jglick: ditto

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +58,560 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
-        container.waitForReady(this);
-
-        // create an admin user
-        privateTokenAdmin =
-                container.createUserToken(adminUserName, "arandompassword12#", "testadmin@invalid.test", "true");
-
-        // create another user
-        privateTokenUser =
-                container.createUserToken(userName, "passwordforsimpleuser12#", "testsimple@invalid.test", "false");
+        long startTime = System.currentTimeMillis();
```
jglick: `Instant.now`, then there is a way to subtract two to get a `Duration`

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +58,560 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
+        String fullPath = owner + "/" + project;
+        branchSource
+                .waitFor()
+                .withMessage("Waiting for GitLab project '%s' to appear in dropdown", fullPath)
+                .withTimeout(Duration.ofSeconds(time.seconds(GITLAB_API_RETRY_TIMEOUT_SECONDS)))
+                .until(() -> {
+                    try {
+                        return branchSource.find(branchSource.by.option(fullPath)) != null;
+                    } catch (NoSuchElementException e) {
+                        return false;
```
jglick: again, IIRC there is some method to ignore a specific exception and treat as false

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +58,560 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
+                });
     }
 
-    private void assertExistAndResult(final WorkflowJob job, final boolean withSuccess) {
-        final Build.Result expectedResult = (withSuccess) ? Build.Result.SUCCESS : Build.Result.FAILURE;
-        waitFor(job, Matchers.pageObjectExists(), (int) time.seconds(3));
-        waitFor(job.getLastBuild().getResult(), Matchers.containsString(expectedResult.name()), (int) time.seconds(3));
+    // ==================== Git operations ====================
+
+    private void setupInitialBranchViaGit(String repoUrl, String branchName, String content) throws IOException {
```
jglick: to be pedantic,
```suggestion
    private void setUpInitialBranchViaGit(String repoUrl, String branchName, String content) throws IOException {
```

since _setup_ is a noun whereas _set up_ is the (separable) verb

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +58,560 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
+                .withTimeout(Duration.ofSeconds(time.seconds(BRANCH_INDEXING_TIMEOUT_SECONDS)))
+                .pollingEvery(Duration.ofSeconds(POLLING_INTERVAL_SECONDS))
+                .until(() -> {
+                    try {
+                        try {
+                            WorkflowJob existingJob = multibranchJob.getJob(branchName);
+                            existingJob.url("").openStream().close();
+                            return existingJob;
+                        } catch (Exception e) {
+                        }
```
jglick: Probably another way to write this for more debuggable failures.

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +58,560 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
+    private void awaitTagsAvailabilityViaApi(GitLabApi gitlabapi, String projectPath, String... tagNames) {
+        long startTime = System.currentTimeMillis();
+        for (String tagName : tagNames) {
+            waitFor()
+                    .withMessage("Waiting for tag '%s' in GitLab API", tagName)
+                    .withTimeout(Duration.ofSeconds(time.seconds(GITLAB_API_RETRY_TIMEOUT_SECONDS)))
+                    .until(() -> {
+                        try {
+                            var tags = gitlabapi.getTagsApi().getTags(projectPath);
+                            return tags != null && tags.stream().anyMatch(tag -> tagName.equals(tag.getName()));
```
jglick: Not sure if the Selenium version supports it, but for Awaitility it is better to

```java
await().until(() -> someThing(), matchesSomeMatcher());
```

jglick (approved):
Not familiar with the original, but looks reasonable.

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
apuig: works with `UncheckedIOException`

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
apuig: This is involved in  `@Before`, its intended as a best-effort attempt.
Would it be better to ignore this step entirely and leave the container’s state as is, since it will NOT be reused anyway?

src/test/java/plugins/GitLabPluginTest.java
apuig: Other tests already make use of `.ignoring(AssertionError.class)`, now this is 
```
assertThat(tags.stream().map(Tag::getName).toList(), hasItems(tagNames));
```

src/test/java/plugins/GitLabPluginTest.java
apuig: somehow improved

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -46,67 +49,55 @@ public String httpHost() {
...
+                .connectTimeout(time(READINESS_CONNECTION_TIMEOUT))
+                .build();
+
         p.waitFor()
                 .withMessage("Waiting for GitLab to come up")
-                .withTimeout(Duration.ofSeconds(200)) // GitLab starts in about 2 minutes add some headway
-                .pollingEvery(Duration.ofSeconds(2))
+                .withTimeout(READINESS_TIMEOUT)
+                .pollingEvery(READINESS_POLL_INTERVAL)
+                .ignoring(UncheckedIOException.class)
```
jtnord:
```suggestion
                .ignoring(IOException.class)
```

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -46,67 +49,55 @@ public String httpHost() {
...
                 .until(() -> {
                     try {
-                        HttpRequest request = HttpRequest.newBuilder()
-                                .uri(getHttpUrl().toURI())
+                        var request = HttpRequest.newBuilder()
+                                .uri(new URL(getHttpUrl()).toURI())
                                 .GET()
-                                .timeout(Duration.ofSeconds(1))
+                                .timeout(time(READINESS_REQUEST_TIMEOUT))
```
jtnord: 5 seconds (and more when elastic) to get the default page seems a little excessive?  why is the timeout that large, If a simple request takes that amount of time, then a complex one used in the actual tests would seem to want to take a lot longer?

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -46,67 +49,55 @@ public String httpHost() {
...
                                 .build();
-                        HttpResponse<String> response = client.send(request, HttpResponse.BodyHandlers.ofString());
+                        var response = client.send(request, HttpResponse.BodyHandlers.ofString());
                         return response.body().contains("GitLab Community Edition");
-                    } catch (IOException ignored) {
-                        // we can not use .ignoring as this is a checked exception (even though a callable can throw
```
jtnord, line 77: is incorrect and you can.

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -119,4 +110,40 @@ public String createUserToken(String userName, String password, String email, St
...
+                        .findFirst()
+                        .ifPresent(project -> {
+                            try {
+                                gitlabapi.getProjectApi().deleteProject(project);
+                            } catch (GitLabApiException e) {
+                                System.err.println("Failed to delete project '" + repoName + "': " + e.getMessage());
+                            }
+                        });
+            } catch (Exception e) {
+                System.err.println("Failed to list projects for '" + repoName + "': " + e.getMessage());
```
jtnord: Use a logger please so the full stack and all causes are logged.

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -119,4 +110,40 @@ public String createUserToken(String userName, String password, String email, St
...
+                .withRequestTimeout(GITLAB_API_CONNECT_TIMEOUT_MS, GITLAB_API_READ_TIMEOUT_MS)) {
+            try {
+                gitlabapi.getProjectApi().getProjects(repoName).stream()
+                        .filter(proj -> repoName.equals(proj.getName()))
+                        .findFirst()
+                        .ifPresent(project -> {
+                            try {
+                                gitlabapi.getProjectApi().deleteProject(project);
+                            } catch (GitLabApiException e) {
+                                System.err.println("Failed to delete project '" + repoName + "': " + e.getMessage());
```
jtnord: Use a logger please so the full stack and all causes are logged.

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -119,4 +110,40 @@ public String createUserToken(String userName, String password, String email, St
...
+
+            try {
+                gitlabapi.getGroupApi().getGroups(groupName).stream()
+                        .filter(g -> groupName.equals(g.getName()))
+                        .findFirst()
+                        .ifPresent(group -> {
+                            try {
+                                gitlabapi.getGroupApi().deleteGroup(group.getId());
+                            } catch (GitLabApiException e) {
+                                System.err.println("Failed to delete group '" + groupName + "': " + e.getMessage());
```
jtnord: Use a logger please so the full stack and all causes are logged.

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -119,4 +110,40 @@ public String createUserToken(String userName, String password, String email, St
...
+                            }
+                        });
+            } catch (Exception e) {
+                System.err.println("Failed to list groups for '" + groupName + "': " + e.getMessage());
```
jtnord: Use a logger please so the full stack and all causes are logged.

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
```diff
@@ -46,67 +49,55 @@ public String httpHost() {
...
-                        HttpResponse<String> response = client.send(request, HttpResponse.BodyHandlers.ofString());
+                        var response = client.send(request, HttpResponse.BodyHandlers.ofString());
                         return response.body().contains("GitLab Community Edition");
-                    } catch (IOException ignored) {
-                        // we can not use .ignoring as this is a checked exception (even though a callable can throw
-                        // this!)
-                        return Boolean.FALSE;
+                    } catch (IOException e) {
+                        throw new UncheckedIOException(e);
                     }
```
jtnord: remove this and reformat the preceeding code so that the `IOException` is handled in the wait

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +64,559 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
+                    .add("--shm-size", "1g")
+                    .add("--memory", "4g")
+                    .add("--memory-swap", "5g"));
+            container = starter.start();
+            container.waitForReady(this);
+            adminToken =
+                    container.createUserToken(ADMIN_USERNAME, "arandompassword12#", "testadmin@invalid.test", "true");
+            userToken =
+                    container.createUserToken(USERNAME, "passwordforsimpleuser12#", "testsimple@invalid.test", "false");
+            System.out.println("GitLab container init: " + elapsed(startTime));
```
jtnord: use a Logger.

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +64,559 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
+        } finally {
+            keepaliveThread.interrupt();
+        }
     }
 
+    @After
+    public void cleanup() throws IOException {
+        if (container != null) {
+            container.cleanup(adminToken, REPO_NAME, GROUP_NAME);
```
jtnord: given the container is being closed, does it need to be cleanedup?

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +64,559 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
+            jenkins.open();
+
+            var starter = gitLabServer.starter();
+            // https://docs.gitlab.com/ee/install/requirements.html#memory
+            starter.withOptions(new CommandBuilder()
+                    .add("--shm-size", "1g")
+                    .add("--memory", "4g")
+                    .add("--memory-swap", "5g"));
+            container = starter.start();
+            container.waitForReady(this);
```
jtnord: FYI, you could simplify all of the above and the thread if you used webdriver to wait for the ready status instead of an native http client.

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +64,559 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
+        multibranchJob.waitForBranchIndexingFinished((int) BRANCH_INDEXING_TIMEOUT.toSeconds());
+
+        // Then all branches and MR discovered, valid builds succeed, broken build fails
+        var log = multibranchJob.getBranchIndexingLogText();
+        assertThat(log, containsString("Finished: SUCCESS"));
+        assertThat(log, containsString("Scheduled build for branch: main"));
+        assertThat(log, containsString("Scheduled build for branch: firstbranch"));
+        assertThat(log, containsString("Scheduled build for branch: failedjob"));
+        assertThat(log, containsString("Scheduled build for branch: MR-1"));
+        assertThat(log, containsString("1 merge requests were processed"));
```
jtnord: NIT use `allOf` if you want to assert multiple matchers on the same object

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +64,559 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
+        final List<WebElement> scanGitLabNow = driver.findElements(
+                by.xpath("//a[span[@class='task-link-text' and text()='Scan GitLab Project Now']]"));
 
-        // create Jenkinsfile on the main branch
-        gitlabapi.getRepositoryFileApi().createFile(project.getId(), file, branchName, "Add Jenkinsfile");
-    }
+        if (!scanGitLabNow.isEmpty()) {
+            scanGitLabNow.get(0).click();
+            return;
+        }
```
jtnord: what happens if the list is not empty (or the list has multiple matches?)

if you are only calling this with something from GitLab than use `findElement` and fail if the element is not found, otherwise if the element lookup fails to find a match the scan will not be triggered which could lead to either failing tests without an obvious reason, or worse flaky tests that are only working due to a periodic scan kicking in.

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +64,559 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
-    private void addFile(GitLabApi gitlabapi, Project project, String branchName, String JenkinsfileContent)
-            throws GitLabApiException {
-        RepositoryFile file = new RepositoryFile();
-        file.setFileName("Jenkinsfile");
-        file.setFilePath("Jenkinsfile");
-        file.setContent(JenkinsfileContent);
+    /**
+     * GitLab uses "Scan GitLab Project Now" instead of generic "Scan Repository Now"
+     */
+    private void reIndex(WorkflowMultiBranchJob job) {
```
jtnord, line 372: NIT: it shouldn't matter - scanning a repo now should for all IIUC is `$jobUrl/build` and so the WorkflowMultiBranchJob PO should be able to handle this in a generic way.

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +64,559 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
+                .ignoring(NoSuchElementException.class)
+                .until(() -> {
+                    Build lastBuild = job.getLastBuild();
+                    if (lastBuild == null) {
+                        return false;
+                    }
+                    String result = lastBuild.getResult();
+                    return result != null && result.contains(expectedResult.name());
+                });
+    }
```
jtnord: you should be able to do this this more simply with

```suggestion
        Build b = job.getLastBuild();
        b.waitUntilFinished(BUILD_COMPLETION_TIMEOUT);
        b.shouldBe(expectedResult);
    }
```

getLastBuild uses the permalink (and so should never return null - so I am not sure why lines 396-398 where added).  Build pages are also lazy - you can create one for any permalink or build number even if they do not exist.

`waitUntilFinished` calls `waitUntilStarted` that will wait until the build at the expected URL exists.

And this uses the Page Objects to show what is happening - so if for example the build is running but taking to long - you would see this in the screen recording as the console is opened in the call.

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +64,559 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
-        assertThat(branchIndexingLog, containsString("Scheduled build for branch: MR-1"));
-        assertThat(branchIndexingLog, containsString("1 merge requests were processed"));
+    private void assertJobExists(final WorkflowJob job) {
+        waitFor()
+                .withMessage("Waiting for job '%s' to be created", job)
+                .withTimeout(JOB_CREATION_TIMEOUT)
+                .pollingEvery(POLLING_INTERVAL)
+                .ignoring(UncheckedIOException.class)
+                .until(() -> {
+                    checkCanOpenJobUrl(job);
```
jtnord: please do not use IOStreams etc.  in the case of failure the screen recording will not show anything useful to help understand the state of the system, nor would a har recording help (should that be enabled).

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +64,559 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
-    private void configureJobWithGitLabBranchSource(
-            final WorkflowMultiBranchJob job, String adminUserName, String project) {
-        final GitLabBranchSource ghBranchSource = job.addBranchSource(GitLabBranchSource.class);
-        ghBranchSource.setOwner(adminUserName);
-        ghBranchSource.setProject(adminUserName, project);
+    private static void checkCanOpenJobUrl(final WorkflowJob job) throws UncheckedIOException {
+        try (var unused = job.url("").openStream()) {
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
```
jtnord: open the jobs parent using webdriver and assert that something has a link to the job url instead for better debugability.

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +64,559 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
+                .withTimeout(GITLAB_API_RETRY_TIMEOUT)
+                .until(() -> {
+                    try {
+                        return projApi.createProject(projectSpec);
+                    } catch (Exception e) {
+                        try {
+                            return projApi.getProject(projectSpec.getPath());
+                        } catch (Exception ex) {
+                            return null;
+                        }
```
jtnord: add `.ignoring(`Exception`)` to the wait (or better still be specific about the exception?) and reduce this block indentation.

otherwise this swallows exceptions in the case of a timeout.

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +64,559 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
+                .until(() -> {
+                    try {
+                        return groupApi.createGroup(groupParams).withVisibility(Visibility.PRIVATE);
+                    } catch (Exception e) {
+                        try {
+                            return groupApi.getGroup(groupPath);
+                        } catch (Exception ex) {
+                            return null;
+                        }
+                    }
```
jtnord: add .ignoring(Exception) to the wait (or better still be specific about the exception?) and reduce this block indentation.

otherwise this swallows exceptions in the case of a timeout.

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +64,559 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
+                .withTimeout(GITLAB_API_RETRY_TIMEOUT)
+                .until(() -> {
+                    try {
+                        return groupApi.addMember(groupId, userId, accessLevel);
+                    } catch (Exception e) {
+                        try {
+                            return groupApi.getMember(groupId, userId);
+                        } catch (Exception ex) {
+                            return null;
+                        }
```
jtnord: add .ignoring(Exception) to the wait (or better still be specific about the exception?) and reduce this block indentation.

otherwise this swallows exceptions in the case of a timeout.

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +64,559 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
+                    } catch (Exception e) {
+                        try {
+                            var filter = new MergeRequestFilter().withProjectId(project.getId());
+                            return gitlabapi.getMergeRequestApi().getMergeRequests(filter).stream()
+                                    .filter(mr -> mrTitle.equals(mr.getTitle()))
+                                    .findFirst()
+                                    .orElse(null);
+                        } catch (Exception ex) {
+                            return null;
+                        }
```
jtnord: add .ignoring(Exception) to the wait (or better still be specific about the exception?) and reduce this block indentation.

otherwise this swallows exceptions in the case of a timeout.

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +64,559 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
+                .withMessage("Waiting for GitLab branch '%s' to be available", branchName)
+                .withTimeout(GITLAB_API_RETRY_TIMEOUT)
+                .ignoring(GitLabApiRuntimeException.class)
+                .until(() -> {
+                    try {
+                        var branch = gitlabapi.getRepositoryApi().getBranch(projectId, branchName);
+                        return branch != null && branch.getCommit() != null;
+                    } catch (GitLabApiException e) {
+                        throw new GitLabApiRuntimeException(e);
+                    }
```
jtnord: why is this caught and something else thrown?  just add the correct exception to the `ignoring`?

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +64,559 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
+                });
+        System.out.println("GitLab branch '" + branchName + "' wait: " + elapsed(startTime));
```
jtnord: logging?

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +64,559 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
+                        try {
+                            var tags = gitlabapi.getTagsApi().getTags(projectPath);
+                            assertThat(tags, is(notNullValue()));
+                            assertThat(tags.stream().map(Tag::getName).toList(), hasItems(tagNames));
+                            return true;
+                        } catch (GitLabApiException e) {
+                            throw new GitLabApiRuntimeException(e);
+                        }
+                    });
+        }
```
jtnord: why is this in a loop - waiting for `tagName` when the body of the wait is using `hasItems(tagNames)`?

Also just throw the `GitLabApiException` and use it in the ignore?

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +64,559 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
+        System.out.println("GitLab tags wait: " + elapsed(startTime));
```
jtnord: logging?

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +64,559 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
+        waitFor()
+                .withMessage("Waiting for GitLab MR %d merge ref to be available", mrIid)
+                .withTimeout(GITLAB_API_RETRY_TIMEOUT)
+                .ignoring(GitLabApiRuntimeException.class)
+                .until(() -> {
+                    try {
+                        return gitlabapi.getRepositoryFileApi().getFile(projectId, "Jenkinsfile", mergeRef);
+                    } catch (GitLabApiException e) {
+                        throw new GitLabApiRuntimeException(e);
+                    }
```
jtnord: just add this exception to the ignoring

src/test/java/plugins/GitLabPluginTest.java
```diff
@@ -49,307 +64,559 @@ public class GitLabPluginTest extends AbstractJUnitTest {
...
+                });
+        System.out.println("GitLab MR " + mrIid + " merge ref wait: " + elapsed(startTime));
```
jtnord: logging

src/test/java/plugins/GitLabPluginTest.java
jtnord: alternatively - should the container be only started and stopped once for the test suite rather than per test (if the startup time is so large?)

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
apuig: container cleanup removed

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
apuig: changed to 1s

src/test/java/plugins/GitLabPluginTest.java
apuig: container cleanup removed.

I considered reusing the container, but since I didn't see this approach used in other tests, I decided to keep the scope of this PR focused on GitLab. 

I'm happy to explore container reuse in a follow-up PR. However, this would likely lead to a broader discussion around Docker lifecycle management, the use of docker-fixtures, and potentially adopting testcontainers (with the known issues!!).

src/test/java/plugins/GitLabPluginTest.java
apuig: Yes, this is much simpler!

Thanks for sharing additional insights!

src/test/java/plugins/GitLabPluginTest.java
apuig: OUCH

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
apuig: Perhaps I'm missing some context here, and I have to admit, with all these try/catch blocks, it really feels like I’m fighting the framework, and it’s definitely polluting the readability of the tests.

But when I try something like:
```java
waitFor()
    .withTimeout(Duration.ofSeconds(10))
    .ignoring(IOException.class)
    .until(() -> {
        if (null == null) {
            System.err.println("about to throw IOException");
            throw new IOException("");
        }
        return true;
    });
 ```
 It fails immediately on the first iteration, wrapped as `AssertionError`:
 ```
 java.lang.AssertionError: java.io.IOException: 
at org.jenkinsci.test.acceptance.junit.Wait$1.apply(Wait.java:128)
at org.openqa.selenium.support.ui.FluentWait.until(FluentWait.java:203)
at org.jenkinsci.test.acceptance.junit.Wait.until(Wait.java:120)
```
https://github.com/jenkinsci/acceptance-test-harness/blob/a73c4ed524d9e0564b4f97b8ff9aeb36ce8d567e/src/main/java/org/jenkinsci/test/acceptance/junit/Wait.java#L119-L130

---
Honestly, I HATE the current implementation with `UncheckFoo`, but I haven’t found a cleaner way to use `Wait`.

If you think it makes sense to include, I could use something along these lines:
```java
public <Return, E extends Exception> Return untilIgnoring(Class<E> exceptionType, ThrowingCallable<Return, E> isTrue) {
    return this.ignoring(exceptionType).until(new Callable<Return>() {
        @Override
        public Return call() throws Exception {
            try {
                return isTrue.call();
            } catch (RuntimeException ex) {
                throw ex;
            } catch (Exception ex) {
                if (exceptionType.isInstance(ex)) {
                    throw ex;
                }
                throw new AssertionError(ex);
            }
        }
    });
}
```
But this would lead to a bigger discussion about the framework’s infrastructure, and I’d rather not derail the GitLab tests scope.

src/test/java/plugins/GitLabPluginTest.java
apuig: now visiting job page

src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer/Dockerfile
```diff
@@ -3,5 +3,118 @@ FROM gitlab/gitlab-ce:18.6.1-ce.0
...
+
+ENV GITLAB_SKIP_PG_UPGRADE=true \
+    GITLAB_SKIP_TAIL_LOGS=true \
+    GITLAB_POST_RECONFIGURE_SCRIPT="" \
+    GITLAB_ROOT_PASSWORD=testpassword123 \
+    GITLAB_SHARED_RUNNERS_REGISTRATION_TOKEN="" \
+    GITLAB_SKIP_UNMIGRATED_DATA_CHECK=true \
+    GITLAB_SKIP_RECONFIGURE=false
+
+ENV GITLAB_OMNIBUS_CONFIG="prometheus_monitoring['enable'] = false; \
```
lemeurherve: Suggestion: put all this configuration in a `gitlab.rb` file then `COPY gitlab.rb /etc/gitlab/` like in https://github.com/jenkinsci/acceptance-test-harness/pull/1425/files#diff-77f64d4d071f5e6f86c44fd13abc2a951c1663da19f7999e78a7bfa03b151ad3

EDIT: there might be equivalents of the env vars defined above, in which case  you could also put them in that config file too instead.

src/test/java/plugins/GitLabPluginTest.java
apuig: You genius! I tried so many things but totally missed the simplest solution.
Thanks, this gets rid of a ton of code! :heart:

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
jtnord: > Perhaps I'm missing some context here

Our overloads are apparently broken and I shall fix them unless someone gets there sooner - in the mean time use `until(Function)` that does not swallow and replace the exceptions with something else.

https://github.com/SeleniumHQ/selenium/blob/a3aa3264f9ed5cf246fa2adac90441cefabe8b24/java/src/org/openqa/selenium/support/ui/FluentWait.java#L197-L238

jtnord: https://github.com/jenkinsci/acceptance-test-harness/issues/2559 FTR

src/test/java/plugins/GitLabPluginTest.java
jtnord: > I didn't see this approach used in other tests,

possibly because those containers start in a reasonable amount of time?

> I decided to keep the scope of this PR focused on GitLab

I was thinking of only doing it for this specific test class as the startup time for the container seems to be on the high side.

more in relation to https://github.com/jenkinsci/acceptance-test-harness/pull/2285#issuecomment-3648413146 and how we would want to handle these tests. 

Anyway - this is not a blocking comment when `JobDSL` is taking so much more time, and something that could be improved afterwards if needed.

src/test/java/plugins/GitLabPluginTest.java
jtnord: Or there is a `Matcher` for [PageObjectExists](https://github.com/jenkinsci/acceptance-test-harness/blob/6c00bd2204988966b5ece06f5f8bccddb1992035/src/main/java/org/jenkinsci/test/acceptance/Matchers.java#L158-L178) but that suffers the same lack of details on a failure to aid test failure debugging.

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
apuig: YAY! This allows a simpler error handling.

I’ve put all the heavy lifting into a `toFunction` method with comments referencing this issue
https://github.com/jenkinsci/acceptance-test-harness/pull/2285/commits/33c26a8c9221368896ab795a3fbac8faeb3583f6

Thanks!

src/test/java/plugins/GitLabPluginTest.java
apuig: created `createOrGetWithRetry` method to reuse the logic on all those _idempotent_ helper methods.
Now it will retry, but always report the first exception (presumably the one containing relevant info)

src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer/Dockerfile
apuig: https://github.com/jenkinsci/acceptance-test-harness/pull/2285/commits/b851d4d1b4606d7307da0f7b0ad0f52f7a9e64a3

Great suggestion! I think it even sped things up a bit in the process :racing_car: (let's wait for this CI run :roll_eyes:)
> FTR I've also updated the link in this question: https://gitlab.com/gitlab-org/omnibus-gitlab/-/work_items/9503

src/test/java/plugins/GitLabPluginTest.java
apuig: basic implementation (just holding a static container), totally agree it’s worth avoiding that expensive startup time

https://github.com/jenkinsci/acceptance-test-harness/pull/2285/commits/3198be39182eee1e0f80090af63cdca34f4cb626

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
apuig: container cleanup included again, but not it doesn't swallow exception

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/GitLabContainer.java
jtnord: > YAY! This allows a simpler error handling.
> 
> I’ve put all the heavy lifting into a `toFunction` method with comments referencing this issue [33c26a8](https://github.com/jenkinsci/acceptance-test-harness/commit/33c26a8c9221368896ab795a3fbac8faeb3583f6)
> 
> Thanks!

you shouldn't need any function just change `until( () -> { /* the code */})` to `until(o -> {/* the code */})`

apuig: With this suggested idiom, the compiler complains about the body of the function throwing a checked exception. The `toFunction` method I added uses a sneaky throw to handle this :see_no_evil:

PR #2561, opened 2025-12-18
Fix experimental UI Build With Parameters ATHs

RanjitCS:
The new experimental build page changes the `Build With Parameters` link to a button. The ATHs need to be changed in accordance else it will just fail.
Before experimental changes
image
image

After experimental changes
image
image

This PR is to fix ATHs that break due to this change of the `Build With Paramters`
### Testing done
SImple test that clicks the link now passes

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/BuildWithParameters.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/BuildWithParameters.java
@@ -49,7 +49,7 @@ public class BuildWithParameters extends PageObject {
     }
 
     public void start() {
-        if (Boolean.getBoolean("new-build-page.flag.defaultValue")) {
+        if (Boolean.getBoolean("new-job-page.flag.defaultValue")) {
             find(by.xpath("//dialog[@open]//div[@id='bottom-sticker']//button[contains(., 'Build')]"))
                     .click();
         } else {
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Job.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Job.java
@@ -367,7 +367,7 @@ public class Job extends TopLevelItem {
             // so wait for it to be added and then disappear
             waitFor(waitFor(By.id("notification-bar"))).until(bar -> !bar.isDisplayed());
         } else {
-            if (Boolean.getBoolean("new-build-page.flag.defaultValue")) {
+            if (Boolean.getBoolean("new-job-page.flag.defaultValue")) {
                 clickButton("Build with Parameters");
             } else {
                 clickLink("Build with Parameters");
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Parameter.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Parameter.java
@@ -53,7 +53,7 @@ public abstract class Parameter extends PageAreaImpl {
             return super.path(rel);
         }
 
-        if (Boolean.getBoolean("new-build-page.flag.defaultValue")) {
+        if (Boolean.getBoolean("new-job-page.flag.defaultValue")) {
             return by.xpath(
                     "//input[@name='name' and @value='%s']/parent::div[@name='parameter']//*[@name='%s']", name, rel);
         }


src/main/java/org/jenkinsci/test/acceptance/po/BuildWithParameters.java
```diff
@@ -49,6 +49,11 @@ public BuildWithParameters enter(List<Parameter> definitions, Map<String, ?> val
     }
 
     public void start() {
-        clickButton("Build");
+        if (Boolean.getBoolean("new-job-page.flag.defaultValue")) {
```
alecharp, line 52: My fear is that this is ok in testing while the "original" value of the property is false (when the flag value is coming from the default flag value, coming from the BooleanExperimentalFlag class). But when we change a flag default value (from the flag class itself) from false to true, this won't report the correct status of the flag.

This should be good enough for now but in mid-long term, we would need to have some sort of API in ATH to get the flag value.

cc @jtnord @janfaracik

jtnord: > This should be good enough for now but in mid-long term, we would need to have some sort of API in ATH to get the flag value.

The value is per user so you would need to do this prior to *every* request prior to interacting. 
It would be better if you can detect this from the page by the page doing something.  or just assuming its false and click the button and then try the fallback (new way).

Anyway - this is another case of DO NOT DO THIS AT ALL...  use a `data-test-id` in the actual page for the button in both legacy and new cases - use the ID to get the button.  no funny code changes.  IE fix upstream.

jtnord (changes requested):
change the experimental pages (and current ones) to use `data-testid`s and do not rely on hard coded paths.

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
```diff
@@ -367,7 +367,11 @@ public Build scheduleBuild(Map<String, ?> params) {
             // so wait for it to be added and then disappear
             waitFor(waitFor(By.id("notification-bar"))).until(bar -> !bar.isDisplayed());
         } else {
-            clickLink("Build with Parameters");
+            if (Boolean.getBoolean("new-job-page.flag.defaultValue")) {
+                clickButton("Build with Parameters");
```
jtnord, line 371: this no longer uses a new page but a dialog - as such the result of this is asynchronous and as such this must be using something like `runThenHandleInputDialog(java.lang.Runnable,java.lang.String)`

I say something like as I do not beleive there is currently something for "do something that will cause a dialog, when the dialog appears to this, then click the default button").

src/main/java/org/jenkinsci/test/acceptance/po/Parameter.java
```diff
@@ -53,6 +53,11 @@ public By path(String rel) {
             return super.path(rel);
         }
 
+        if (Boolean.getBoolean("new-job-page.flag.defaultValue")) {
+            return by.xpath(
+                    "//input[@name='name' and @value='%s']/parent::div[@name='parameter']//*[@name='%s']", name, rel);
```
jtnord, line 58: time to add `data-testid=parameter-${name}` so things can be found.

However the location structure is identical between the existing and experimental page?dialog vs standard page. 
The only difference is that we are inside a dialog. In other words this should be the same code for both cases given the search is in the whole document `//...`

src/main/java/org/jenkinsci/test/acceptance/po/BuildWithParameters.java
```diff
@@ -49,6 +49,11 @@ public BuildWithParameters enter(List<Parameter> definitions, Map<String, ?> val
...
+            find(by.xpath("//dialog[@open]//div[@id='bottom-sticker']//button[contains(., 'Build')]"))
```
jtnord, line 53: `BuildWithParameters` is a dialog in the Job page now - so this whole class should probably be deprecated.  The Page `/build` is still present and directly accessible and uses the old format and anyone using this PO to access that page should still work.

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
jtnord: this also needs to wait until it is scheduled = see the code comment above about the notification bar

src/main/java/org/jenkinsci/test/acceptance/po/BuildWithParameters.java
jtnord: the above comment is mostly moot as the build with parameters is not a page in the new UI so this code should be deprecated and the code put into the Build pageObject.

PR #2563, opened 2025-12-20
Adapt credentials plugin for cards

timja:
See https://github.com/jenkinsci/credentials-plugin/pull/984

Running full set of tests then I'll revert the local override so it can be merged

### Testing done

Ran credentials plugin locally:

The two failures were weird frameworky issues, they passed on re-run.
image

### Submitter checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -492,10 +492,6 @@ and
           </systemPropertyVariables>
           <!-- SUREFIRE-1588 workaround; too late for systemProperties: -->
           <argLine>-Djdk.net.URLClassPath.disableClassPathURLCheck=true</argLine>
-          <environmentVariables>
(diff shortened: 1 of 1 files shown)


src/test/java/core/CredentialsTest.java
```diff
@@ -33,7 +33,7 @@ public class CredentialsTest extends AbstractJUnitTest {
     private static final String CRED_DOMAIN = "mydomain";
     private static final String CRED_USER = "user";
     private static final String CRED_PWD = "password";
-    private static final String CRED_DESCR = "descr";
+    private static final String CRED_DESCR = "My-super-unique-description";
```
timja, line 36: It would be better if the tests used ID for locating the credentials but since its not I was a bit worried that this could end up not being unique enough so I just made it more specific, can drop this though.

PR #2572, opened 2026-01-03
Test 'Refine the 'New Item' screen'

janfaracik:
Test https://github.com/jenkinsci/jenkins/pull/26031

### Submitter checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -41,7 +41,7 @@ if (needSplittingFromWorkspace) {
 def axes = [
   jenkinsVersions: ['lts', 'latest'],
   platforms: ['linux'],
-  jdks: [17, 21, 25],
+  jdks: [21, 25],
   browsers: ['firefox'],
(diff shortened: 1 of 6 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/JobsMixIn.java
```diff
@@ -97,6 +97,10 @@ public void copy(Job from, String to) {
 
     public void copy(String from, String to) {
         visit("newJob");
+
+        // Newer versions of Jenkins have an additional radio before the 'from' input is available, so click it
+        all(by.radioButton("Duplicate an existing item")).forEach(WebElement::click);
```
jtnord, line 102: so this should actually wait until that JavaScript has kicked in and the "from" is visible?
currently it assumes the from is immediately visible which would be a mistake and lead to flakes in this code.

PR #2581, opened 2026-01-13
(Placeholder) Use Java 21 container in SshAgentContainer

gbhat618:
The tests in `SshSlavesPluginTest` are failing when running with controller compiled for java 21 due agent being still Java 17.
Will be using the image from https://github.com/jenkinsci/docker-fixtures/pull/138

PR should be in draft until the `docker-fixtures` has a release and need to bump https://github.com/jenkinsci/acceptance-test-harness/blob/7b8c43086a6f28acb2b83f57abae9410b5b57eab/pom.xml#L357-L361 

### Testing done

TBU

### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

Issue #857, opened 2022-07-08
Replace `docker-fixtures` with `testcontainers`


timja:
[`docker-fixtures`](https://github.com/jenkinsci/docker-fixtures/) has been deprecated for a long time, https://www.testcontainers.org is the recommended replacement.

We should switch to it.

jtnord:
there are issues with testcontainers - I would not recommend switching right away, or this [PR ](https://github.com/testcontainers/testcontainers-java/pull/6445) should be inlined.

olamy:
or upgrade to junit5 and simply use `@Testcontainers(disabledWithoutDocker = true)` ?

jtnord:
noting https://github.com/jenkinsci/acceptance-test-harness/pull/1701#issuecomment-2324335811

jglick:
> it will skip the entire test class/suite and not just the test that requires the docker container

FWIW I find JUnit rules (4 or 5) to just be conceptual baggage in most cases. You can use Testcontainers directly without any JUnit-specific APIs, using a `try`-with-resources block, and use JUnit’s normal systems for assuming some prerequisites from the environment.

PR #2582, opened 2026-01-13
Converting `SshAgentContainer` & `XvncSlaveContainer` to Testcontainers

jglick:
Beginning of #857 (not converting other fixtures yet). Needed to make agents run on Java 21 (https://github.com/jenkinsci/acceptance-test-harness/pull/2576#issuecomment-3745160757). Avoids the need for a new `docker-fixtures` release (https://github.com/jenkinsci/docker-fixtures/pull/138#issuecomment-3744780104 https://github.com/jenkinsci/docker-fixtures/pull/122#issuecomment-3046351724) and so supersedes #2581. Cannot just use `jenkins/ssh-agent` like in https://github.com/jenkinsci/ssh-agents-plugin/pull/984 since some of these tests rely on password authentication, whereas the official image only supports SSH private key authentication. For now I just inlined the original `Dockerfile` contents; could be cleaned up in various ways later.

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer.java
@@ -43,27 +43,21 @@ public class SshAgentContainer extends GenericContainer<SshAgentContainer> {
     }
 
     public String getEncryptedEd25519PrivateKey() throws IOException {
-        return new String(
-                SshAgentContainer.class
-                        .getResourceAsStream("SshAgentContainer/ed25519.priv")
-                        .readAllBytes(),
-                StandardCharsets.US_ASCII);
+        return load("ed25519.priv");
     }
 
     public String getEncryptedEd25519PrivateKeyPassphrase() throws IOException {
-        return new String(
-                SshAgentContainer.class
-                        .getResourceAsStream("SshAgentContainer/ed25519.pass")
-                        .readAllBytes(),
-                StandardCharsets.US_ASCII);
+        return load("ed25519.pass");
     }
 
     public String getPrivateKeyString() throws IOException {
-        return new String(
-                SshAgentContainer.class
-                        .getResourceAsStream("SshAgentContainer/unsafe")
-                        .readAllBytes(),
-                StandardCharsets.US_ASCII);
+        return load("unsafe");
+    }
+
+    private String load(String resourceFile) throws IOException {
+        try (var is = SshAgentContainer.class.getResourceAsStream("SshAgentContainer/" + resourceFile)) {
+            return new String(is.readAllBytes(), StandardCharsets.US_ASCII);
+        }
     }
 
     public SshSlaveLauncher configureSSHSlaveLauncher(DumbSlave agent) {


pom.xml
```diff
@@ -414,6 +414,11 @@ and
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-jdk14</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.testcontainers</groupId>
+      <artifactId>testcontainers</artifactId>
+      <version>2.0.3</version>
+    </dependency>
```
jtnord, line 421: I would be surprised if this did not break the keycloak tests and need an update in the `oid-auth` tests.

Any tests using testcontainers (including any in CloudBees?) may need to be updated (both dependencies and tests?).
For this ATH you can upgrade `com.github.dasniko:testcontainers-keycloak` to `4.1.0` (any 4.x) and hope for the best but I guess there are annotation or code changes needed at the same time.

CloudBees also uses some testcontainers that would likely be broken by this.

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer.java
```diff
@@ -24,28 +24,52 @@
...
     }
 
-    public String getEncryptedEd25519PrivateKeyPassphrase() {
-        return resource("ed25519.pass").asText();
+    public String getEncryptedEd25519PrivateKey() throws IOException {
+        return new String(
+                SshAgentContainer.class
+                        .getResourceAsStream("SshAgentContainer/ed25519.priv")
+                        .readAllBytes(),
+                StandardCharsets.US_ASCII);
```
jtnord: please ensure streams are closed even in test code.

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer.java
```diff
@@ -24,28 +24,52 @@
...
  */
-@DockerFixture(id = "ssh-agent", ports = 22)
-public class SshAgentContainer extends JavaContainer {
+public class SshAgentContainer extends GenericContainer<SshAgentContainer> {
 
-    public String getEncryptedEd25519PrivateKey() {
-        return resource("ed25519.priv").asText();
+    public SshAgentContainer() {
+        super(new ImageFromDockerfile("localhost/testcontainers/ath-ssh-agent", false)
+                .withFileFromClasspath(".", SshAgentContainer.class.getName().replace('.', '/')));
```
jtnord, line 41: NIT: generally you should only add the things that are needed - otherwise all of them can be sent to the testcontainer server for building (and it takes longer). (the build is currently only a local(ish) docker compliant server so currently this does not add too much overhead)

pom.xml
jglick: #2262 perhaps

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer.java
```diff
@@ -24,28 +24,52 @@
...
  * Jenkins agent with various login methods.
  */
-@DockerFixture(id = "ssh-agent", ports = 22)
-public class SshAgentContainer extends JavaContainer {
+public class SshAgentContainer extends GenericContainer<SshAgentContainer> {
 
-    public String getEncryptedEd25519PrivateKey() {
-        return resource("ed25519.priv").asText();
+    public SshAgentContainer() {
+        super(new ImageFromDockerfile("localhost/testcontainers/ath-ssh-agent", false)
```
jtnord, line 40: I thought not deleting the image would causes issues with stale images when the Dockerfile or other input was changed, but apparently there is something in testcontainers here that handles this.  (could not find docs on it though)

src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer/Dockerfile
```diff
@@ -1,4 +1,47 @@
...
+  apt-get install -y \
+  openssh-server \
+  locales
+RUN mkdir -p /var/run/sshd
+
+# create a test user
+RUN useradd test -d /home/test && \
+  mkdir -p /home/test/.ssh && \
+  echo 'ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDzpxmTW9mH87DMkMSqBrSecoSHVCkKbW5IOO+4unak8M8cyn+b0iX07xkBn4hUJRfKA7ezUG8EX9ru5VinteqMOJOPknCuzmUS2Xj/WJdcq3BukBxuyiIRoUOXsCZzilR/DOyNqpjjI3iNb4los5//4aoKPCmLInFnQ3Y42VaimH1298ckEr4tRxsoipsEAANPXZ3p48gGwOf1hp56bTFImvATNwxMViPpqyKcyVaA7tXCBnEk/GEwb6MiroyHbS0VvBz9cZOpJv+8yQnyLndGdibk+hPbGp5iVAIsm28FEF+4FvlYlpBwq9OYuhOCREJvH9CxDMhbOXgwKPno9GyN kohsuke@atlas' > /home/test/.ssh/authorized_keys && \
+  echo 'ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDagNSCDst/8z5oH9S5QWr+QNdx+haImY0FD3IQvKdD+eWI9zUbBgtoo/yYEbLvpTWiKsgT3Hw1F8mZ+/bd2Uv3lPyoG+TSzrHL4gSal6d1RWVjCOzSosciXVm4gRUvJjKXzaz8dOg+ii9yIrbeONNK0nlDUCAKy5YXSEl0avcPdUDyR3cStL6870SyanxAzktDw0n8xMq4F/alF3PZ002bcZJrmDeNVAwkP+uO2Tf8pN37SU+nApotZmlmZR32xYHnx+/OiQ7gOAVYmgNRMg0Kwh6Q73FcY3ZWCeNHwLnr95LoEAdj3On8Qr62VhGThuQNVCqBc6SeYjArfjijpcW9 jenkins-ci@localhost' >> /home/test/.ssh/authorized_keys && \
```
jtnord, line 20: Why do this as well as the following below (those public keys are identical are they not, and if not would it not be better to just manage them all in the same way to avoid confusion)?

```
COPY *.pub /tmp/
RUN cat /tmp/*.pub >> /home/test/.ssh/authorized_keys
```

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer.java
jglick: > generally you should only add the things that are needed

WDYM? This is sending https://github.com/jenkinsci/acceptance-test-harness/tree/5f032d5666d5e8e8fe0ccdf635d6a0866654c851/src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer which is exactly what is needed by the Docker daemon.

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer.java
jglick: AFAIK specifying `true` (or not specifying it, e.g., just use the no-arg ctor) just means that it will do something like `docker rmi` after the test completes, throwing away your cache. Whereas passing `false` is the equivalent of just `docker build`.

src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer/Dockerfile
jglick: To try to keep this PR straightforward to review,

> For now I just inlined the original Dockerfile contents; could be cleaned up in various ways later.

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer.java
jtnord: > > generally you should only add the things that are needed
> 
> WDYM? This is sending https://github.com/jenkinsci/acceptance-test-harness/tree/5f032d5666d5e8e8fe0ccdf635d6a0866654c851/src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer which is exactly what is needed by the Docker daemon.

it only needs **some** of the files in that directory (the Dockerfile and the `.pub`s and the .pub is [questionable](https://github.com/jenkinsci/acceptance-test-harness/pull/2582#discussion_r2694143569)) .

e.g.

```.java
   // ...
   .withFileFromClasspath("Dockerfile", SshAgentContainer.class.getName().replace('.', '/'))
   .withFileFromClasspath("ed25519.pub", SshAgentContainer.class.getName().replace('.', '/') + "/ed25519.pub")
   .withFileFromClasspath("unsafe_enc_key.pub", SshAgentContainer.class.getName().replace('.', '/') + "/unsafe_enc_key.pub")
   //...
```

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer.java
jglick: Confirmed: edits to `Dockerfile` are honored when you next `mvn test`.

jtnord: > Confirmed: edits to `Dockerfile` are honored when you next `mvn test`.

sorry I was unclear - I had already tested this happened did not mean for you to test it, and this was a comment for anyone else who may see this in the future.

src/main/resources/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer/Dockerfile
jglick: IOW, this is making the minimum changes:
- changing Java 17 to 21
- switch from `docker-fixtures` to exact equivalent `Dockerfile` layers

src/main/java/org/jenkinsci/test/acceptance/docker/fixtures/SshAgentContainer.java
jglick: OK we could `.dockerignore` the private keys for example, but these are tiny files. We just do not want to be streaming megabytes of unused binaries, and we are not.

jtnord: hence was a NIT, not blocking anything on this

PR #2589, opened 2026-01-20
Wait for 'Build Now' link visibility before clicking

LakshyaBagani:
The Jenkins core race condition fix hides the 'Build Now' button (display: none) until JavaScript loads. This caused ATH tests to time out because clickLink('Build Now') couldn't find the hidden element.

This fix adds `waitFor(by.link('Build Now'))` before clicking, which polls until the element is visible (with ATH's default 120s timeout).

Fixes timeouts in:
- core.FreestyleJobTest.doNotDiscardSuccessfulBuilds
- core.FreestyleJobTest.archiveArtifacts
- core.TriggerRemoteBuildsTest.triggerBuildRemotely

### Testing done

The fix was tested by reviewing the CI failure logs which showed all 3 tests failing at `Job.scheduleBuild(Job.java:368)` when calling `clickLink("Build Now")`. The tests timed out after 120 seconds waiting for an element that was hidden.

The `waitFor(by.link("Build Now"))` method polls until the element becomes visible, using ATH's existing wait infrastructure. This is the same pattern already used elsewhere in the codebase (e.g., the notification-bar wait on the following line).

Related upstream change: https://github.com/jenkinsci/jenkins/pull/26124 (Jenkins core race condition fix)

### Submitter checklist

Desired revierwe
@timja @mawinter69 @MarkEWaite

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/Job.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/Job.java
@@ -362,6 +362,8 @@ public class Job extends TopLevelItem {
         open();
         int nb = getJson().get("nextBuildNumber").intValue();
         if (parameters.isEmpty()) {
+            // Wait for "Build Now" link to become visible (JS shows it after page load)
+            waitFor(by.link("Build Now"));
             clickLink("Build Now");
             // the notification bar can place itslef over other elements
             // so wait for it to be added and then disappear


src/main/java/org/jenkinsci/test/acceptance/po/Job.java
```diff
@@ -362,6 +362,8 @@ public Build scheduleBuild(Map<String, ?> params) {
         open();
         int nb = getJson().get("nextBuildNumber").intValue();
         if (parameters.isEmpty()) {
+            // Wait for "Build Now" link to become visible (JS shows it after page load)
+            waitFor(by.link("Build Now"));
```
jtnord, line 366: Are you sure this fixes the issue,  as this is looking for a `a` element with the text `Build Now` and IIUC the upstream change does not change that.  
In essence I think this is just wasting some CPU such that the timing changes enough that clicking the link works.

Additionally if you do need to wait and then interact with the thing you waited for then use the WebElement returned by `waitFor`

timja, line 366: I'm not sure that this would work, please test this in your PR in the jenkins repository using an incremental build when the tests here pass

krisstern:
Then we should close this PR

PR #2592, opened 2026-01-22
Adapt Script Console for experimental Manage Jenkins UI

janfaracik:
Testing https://github.com/jenkinsci/jenkins/pull/11333

### Submitter checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -30,7 +30,7 @@
   <properties>
     <changelist>999999-SNAPSHOT</changelist>
     <spotbugs.skip>true</spotbugs.skip>
-    <jenkins.version>2.548-rc37927.a_a_13b_d4e0224</jenkins.version>
+    <jenkins.version>2.547</jenkins.version>
     <selenium.version>4.40.0</selenium.version>
     <!-- aligned with selenium -->
     <guava.version>33.5.0-jre</guava.version>


src/main/java/org/jenkinsci/test/acceptance/po/Node.java
```diff
@@ -82,7 +82,7 @@ public String runScript(String script, int timeoutSeconds, Object... args) {
         visit("script");
         CodeMirror cm = new CodeMirror(this, "/script");
         cm.set(String.format(script, args));
-        clickButton("Run");
+        find(by.css("button[value='Run']")).click();
```
jtnord, line 85: doesn't this imply that `clickButton` should be fixed after all this is a button and it has the text `Run`?
If we have multiple buttons all with the same value - then this would appear to be a UX issue as well. If we need to disambiguate the main "run" from an example script "run" then why are we not using IDs or data-test-ids

timja: The text is different, its just the button has the text `Run example` but because it does a contains it breaks it.

It uses a contains because some buttons have SVGs in them so its a workaround for that.

IDs or data-test-ids would work but this also works fine too.

jtnord: so to cope with svg we should be using `innerText` and if that SVG has text then it should be included in the match (the string supplied to `clickButton(String)`) as it is part of the button text?

PR #2621, opened 2026-02-21
Fix #2594: Make ByFactory.button use exact text match

### LINKED ISSUE #2594: clickButton(String) is overly permissive in finding buttons
https://github.com/jenkinsci/acceptance-test-harness/pull/2592 was prompted as the text matching on clickButton was overly permissive, looking for contains rather than being equal.

If a button contains an SVG with text then that text should be included in the match, ie if the button is defined as follows:

```html
<button>this is a button
  <svg viewBox="0 0 240 80" xmlns="http://www.w3.org/2000/svg">
    <text x="20" y="35" class="small">My</text>
    <text x="40" y="35" class="heavy">cat</text>
    <text x="55" y="55" class="small">is</text>
    <text x="65" y="55" class="Rrrrr">Grumpy!</text>
  </svg>  
</button>
```
then [`clickButton()`](https://github.com/jenkinsci/acceptance-test-harness/blob/45631331a9687b5c7e417b2e5947ee36f0845f3e/src/main/java/org/jenkinsci/test/acceptance/po/CapybaraPortingLayerImpl.java#L98-L102) should be called as `clickButton("this is a button\nMy\ncat\nis\nGrumpy"); not `clickButton("this is a button");

This causes issues when in the example of the referenced PR there are other buttons that contain "Run" that are matched.

TL;DR `ByFactory.byButton` with String should take exact text to match, or an ID.  it should not use partial text.

Harmit07:
Fixes #2594

### Description
Updated the XPath selector in `ByFactory.button(String)` to use exact text matching instead of overly permissive partial text matching (`contains()`). 

Previously, `clickButton("Run")` would match buttons with nested SVGs or extra text (e.g., `<button>Run <svg>...</svg></button>`), leading to unintended elements being clicked when multiple buttons shared a partial word. The selector now strictly enforces exact equality (`=`) for the text content, `@id`, `@name`, `@value`, and `@title` attributes.

### Testing done

Ran the ATH test suite locally (`mvn clean test`). 

As anticipated in the original issue description, making this locator strictly evaluate exact text causes failures in existing tests that were relying on the overly permissive partial match (e.g., `ConfigFileProviderTest`, `CompressArtifactsPluginTest`, and `CredentialsTest` where timeouts and `NoSuchElementException`s occur because the tests are not passing the exact string). 

I am submitting this core fix for the locator strategy. I am happy to help update the specific failing test cases to use the exact button strings based on the CI results, or collaborate with maintainers to update the reliant tests!

### Submitter checklist


timja:
The failing tests need to be updated too otherwise this can’t be merged

src/main/java/org/jenkinsci/test/acceptance/po/Node.java
```diff
@@ -82,7 +82,9 @@ public String runScript(String script, int timeoutSeconds, Object... args) {
         visit("script");
         CodeMirror cm = new CodeMirror(this, "/script");
         cm.set(String.format(script, args));
-        find(by.css("button[value='Run']")).click();
+
+        // REVERTED WORKAROUND: Now using the standard clickButton helper
```
timja: don't include junk comments
```suggestion
```

src/main/java/org/jenkinsci/test/acceptance/plugins/ssh_slaves/SshSlaveLauncher.java
```diff
@@ -37,7 +37,9 @@ public SshSlaveLauncher(PageObject context, String path) {
     public <T extends Credential> T addCredential(Class<T> type) {
         find(by.button("Add")).click();
 
-        find(by.css(".jenkins-dropdown")).findElement(by.button("Global")).click();
+        find(by.css(".jenkins-dropdown"))
+                .findElement(by.css(".jenkins-dropdown__item"))
```
timja: why has this been changed? part of this issue is that we should be using the link text if there's additional text then that should be included here

Harmit07: Hi @timja, I used the CSS selector because the exact text match was failing with 'Global' and 'Jenkins'. I will revert this to use the exact button text. Based on the logs, I'll update it to the full visible string once I've confirmed the exact wording in the UI.

Harmit07:
Hi @timja, thanks for the pointer! I checked the DOM and realized XPath squashes the <span> and <div> text together since there's no space between them. I've updated the code to exactly match 'GlobalCredentials...'.

timja (approved):

timja:
Thanks!

src/main/java/org/jenkinsci/test/acceptance/ByFactory.java
```diff
@@ -173,7 +173,7 @@ private static String fieldXPath(String base, String locator) {
      */
     public By button(String locator) {
         return xpath(
-                ".//input[./@type = 'submit' or ./@type = 'reset' or ./@type = 'image' or ./@type = 'button'][((./@id = '%1$s' or ./@name = '%1$s' or contains(./@value, '%1$s')) or contains(./@title, '%1$s'))] | .//input[./@type = 'image'][contains(./@alt, '%1$s')] | .//button[(((./@id = '%1$s' or contains(./@value, '%1$s')) or contains(normalize-space(.), '%1$s')) or contains(./@title, '%1$s'))] | .//input[./@type = 'image'][contains(./@alt, '%1$s')]",
+                ".//input[./@type = 'submit' or ./@type = 'reset' or ./@type = 'image' or ./@type = 'button'][./@id = '%1$s' or ./@name = '%1$s' or ./@value = '%1$s' or ./@title = '%1$s'] | .//input[./@type = 'image'][./@alt = '%1$s'] | .//button[./@id = '%1$s' or ./@value = '%1$s' or normalize-space(.) = '%1$s' or ./@title = '%1$s'] | .//input[./@type = 'image'][./@alt = '%1$s']",
```
RanjitCS, line 176: FYI, as this change works for exact matches from now on
```
<button>
   <strong>Run</strong>
   <p>Something</p>
</button>
```
```
clickButton("Run") #would have worked
clickButton("RunSomething") # is the change that needs to be made from now on  
```
now the more problematic issue is when you have something like 
```
<button>
   <strong>Run</strong>
   <svg >
     <title>some random title</title>
    </svg>
</button>
clickButton("Runsome random title")
```
which isnt exactly user-friendly imo, as the title will be some random thing. Maybe we could think about accepting the check against the child elements as well? something like
```
or text()[normalize-space(.)='%1$s'] or *[normalize-space(.) = '%1$s']]
```

jtnord: > now the more problematic issue is when you have something like...
> which isnt exactly user-friendly imo


correct it is not user friendly - but probably not in the way you are thinking.

Think of accessibility options (screen readers or the like?) where you may ponder if adding titles in svgs are a good idea in buttons when the information they convey is already conveyed in the text :)


```xml
<button>
   <strong>Delete</strong>
   <svg >
     <title>A rubbish bin or trashcan</title>

     <!-- paths etc that draw a rubbish bin -->
    </svg>
</button>
```

Would reading out "A rubbish bin or trashcan" do anything that "Delete" would not - would it make it easier to understand the flow of a page or not?

PR #2648, opened 2026-03-20
Fix Wait.until overloads replacing checked exceptions with AssertionError

### LINKED ISSUE #2559: Wait.Until overloads replace checked exceptions
### Jenkins and plugins versions report

As noted by @apuig  in https://github.com/jenkinsci/acceptance-test-harness/pull/2285#discussion_r2625101216 _[permalink edited by jglick]_ ignoring checked exceptions does not work correctly with our overloads of `until` in `src/main/java/org/jenkinsci/test/acceptance/junit/Wait.java`

This is because we catch checked exceptions and throw an `AssertionException`.  we should just propagate the underlying exception so that it can be correctly ignored.

### What Operating System are you using (both controller, and any agents involved in the problem)?

N/A

### Reproduction steps

with the following code

```
waitFor()
    .withTimeout(Duration.ofSeconds(10))
    .ignoring(IOException.class)
    .until(() -> {
        if (null == null) {
            System.err.println("about to throw IOException");
            throw new IOException("");
        }
        return true;
    });
```

### Expected Results

the wait fails only after the timeout and the timeout contains details of the last exception

### Actual Results

The wait fails immediately on the first iteration, wrapped as `AssertionError`:

```
java.lang.AssertionError: java.io.IOException: 
at org.jenkinsci.test.acceptance.junit.Wait$1.apply(Wait.java:128)
at org.openqa.selenium.support.ui.FluentWait.until(FluentWait.java:203)
at org.jenkinsci.test.acceptance.junit.Wait.until(Wait.java:120)
```

### Anything else?

Any tests that have assumed that a checked exception has been wrapped in an `AssertionError` will need to be adjusted

### Are you interested in contributing a fix?

_No response_

AMITYADAV-16:
Checked exceptions thrown inside Wait.until(Callable) and Wait.until(Predicate) 
were being wrapped in AssertionError, which prevented FluentWait.ignoring() from 
correctly matching and suppressing them. Fixed by using sneakyThrow to propagate 
the original exception type.

Also removed the toFunction workaround from GitLabPluginTest that was added to 
work around this bug, and removed AssertionError.class from the ignoring() call 
in awaitTagsAvailabilityViaApi.

Fixes #2559

### DIFF
--- a/src/test/java/plugins/GitLabPluginTest.java
+++ b/src/test/java/plugins/GitLabPluginTest.java
@@ -15,7 +15,6 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.Callable;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.function.Function;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.gitlab4j.api.GitLabApi;
@@ -161,7 +160,7 @@ public class GitLabPluginTest extends AbstractJUnitTest {
                     .withMessage("Deleting all GitLab projects")
                     .withTimeout(GITLAB_API_RETRY_TIMEOUT.multipliedBy(2))
                     .ignoring(GitLabApiException.class)
-                    .until(toFunction(() -> {
+                    .until(() -> {
                         for (var project : gitlabapi.getProjectApi().getProjects()) {
                             if (project.getMarkedForDeletionOn() != null) {
                                 continue;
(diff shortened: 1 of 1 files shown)


MarkEWaite:
@AMITYADAV-16 this pull request appears to be stalled.  The ci.jenkins.io results indicate that the GitLab API tests have never passed with this change.  Do you intend to continue refining this pull request or should it be closed?

PR #2685, opened 2026-04-19
Test 'Refine System Log' pages

janfaracik:
Tests https://github.com/jenkinsci/jenkins/pull/26614

### Submitter checklist


src/main/java/org/jenkinsci/test/acceptance/po/JenkinsLogger.java
```diff
@@ -22,7 +22,13 @@ public JenkinsLogger(Jenkins jenkins, String name) {
      * @see Jenkins#createLogger(String, Map)
      */
     public static JenkinsLogger create(Jenkins j, String name, Map<String, Level> levels) {
-        j.visit("log/new");
+        j.visit("log");
+
+        // This can be simplified to clicking the Add recorder button once the baseline is high enough
```
jtnord, line 27: isn't this always a button?  why not use the button finder?

src/main/java/org/jenkinsci/test/acceptance/po/JenkinsLogger.java
```diff
@@ -22,7 +22,13 @@ public JenkinsLogger(Jenkins jenkins, String name) {
...
+        j.find(by.xpath(
+            "//button[normalize-space(.)='Add recorder'] | //a[normalize-space(.)='Add recorder']"
+        )).click();
+        j.waitFor(by.xpath("//form[contains(@name, 'configSubmit')]"), 10);
```
jtnord: if this is using a diaglog (which it looks like from the upstream) then this method of obtaining the button is incorrect.
Check the helpers for `runThenHandleDialog` (or somethign similar)  (TL;DR this can match whilst the dialog is animating into view, and cause the click to miss) causing flaky tests

jtnord (changes requested):

src/main/java/org/jenkinsci/test/acceptance/po/JenkinsLogger.java
janfaracik: Nope, before this PR the Add Recorder is a link to `/jenkins/manage/log/new`

src/main/java/org/jenkinsci/test/acceptance/po/JenkinsLogger.java
janfaracik: It's trying to handle both scenarios in one, that being there's a page with the form or there's the new dialog with the form.

jtnord: Either way because the button could be animated using the dialog, you will need to ensure the button is not moving before the click is attempted

janfaracik: Grand. Updated it to handle both scenarios.

jtnord (approved):

src/main/java/org/jenkinsci/test/acceptance/po/JenkinsLogger.java
```diff
@@ -22,9 +22,18 @@ public JenkinsLogger(Jenkins jenkins, String name) {
      * @see Jenkins#createLogger(String, Map)
      */
     public static JenkinsLogger create(Jenkins j, String name, Map<String, Level> levels) {
-        j.visit("log/new");
-        j.find(by.path("/name")).sendKeys(name);
-        j.clickButton("Create");
+        j.visit("log");
+
+        // This can be simplified to clicking the Add recorder button once the baseline is high enough
```
jtnord:
```suggestion
        // TODO: This can be simplified to clicking the Add recorder button once the baseline is higher than 2.571
```

PR #2692, opened 2026-04-30
Adapt to oic-auth changes from https://github.com/jenkinsci/oic-auth-plugin/pull/727

olamy:
Need some adaptation due to this change https://github.com/jenkinsci/oic-auth-plugin/pull/727/changes#diff-9fb278fbfab3237a9630bf5099c3101577357743bc30fbfe163dffab9feb12ccL6

Signed-off-by: Olivier Lamy <olamy@apache.org>

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/OicAuthSecurityRealm.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/OicAuthSecurityRealm.java
@@ -5,7 +5,7 @@ import org.openqa.selenium.WebElement;
 /**
  * Security Realm provided by oic-auth plugin
  */
-@Describable("Login with OpenID Connect")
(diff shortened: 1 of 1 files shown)


jglick:
OK but IIRC you can ask `@Describable` to take a class name (`Descriptor.id`), or give multiple names so the test will pass on both plugin versions at least temporarily.

olamy:
> OK but IIRC you can ask `@Describable` to take a class name (`Descriptor.id`), or give multiple names so the test will pass on both plugin versions at least temporarily.

Thanks. should be fixed with 0121f0cc3c68ec43e83746177b7eed9e79507838

jglick (approved):

jglick:
(The id approach is preferred when it works. Not sure if it does here.)

PR #2702, opened 2026-05-15
Correctly setup I/O for subprocesses

### LINKED ISSUE #2701: each git command in GitRepo take ~ 10s on windows
### Reproduction steps

enable the logging in https://github.com/jenkinsci/acceptance-test-harness/blob/c90abc11d5a43a5d14bf9b43c641e32184d53ac6/src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java#L167-L168 and then run a test that excersizes the code (e.g. GitLabTest)

### Expected Results

each git command completes in sub-second time

### Actual Results

commands take approx 10 seconds

### Are you interested in contributing a fix?

yes

jtnord:
As we are handling the streams we must use PIPE (the default) not INHERIT.

On windows this caused delays of approx 10s, presumable waiting for surefire to read the error stream.  But we want the error stream anyway as it will contain errors, and the message is used when the process exits with non zero status.

Rather than handle the error stream specially, just redirect it to the output.

fixes: #2701

### Testing done

ran the new test before the fix and observed it failed.  
ran the test with the fixed code 10 times and did not observe any failures

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java
@@ -13,6 +13,7 @@ import java.io.FileWriter;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.io.UncheckedIOException;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.Charset;
@@ -22,6 +23,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardOpenOption;
 import java.nio.file.attribute.PosixFilePermission;
+import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.HashSet;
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java
```diff
@@ -149,12 +149,11 @@ public String gitDir(File dir, Object... args) {
         try {
             var log = Logger.getLogger("GitRepo");
             var start = Instant.now();
-            Process p = pb.directory(dir)
-                    .redirectInput(ProcessBuilder.Redirect.INHERIT)
-                    .redirectError(ProcessBuilder.Redirect.INHERIT)
```
jtnord, line 154: the default is `Redirect.PIPE` which is what we actually require

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java
```diff
@@ -149,12 +149,11 @@ public String gitDir(File dir, Object... args) {
...
-                    .start();
+            Process p = pb.directory(dir).redirectErrorStream(true).start();
```
jtnord: no point putting any error messages in the surefire log, put it in the assert message (using the SF log is what causes the delay)

jglick: What about when exit code is zero? I would still want to see output in the Surefire log.

BTW using `INHERIT` can cause Surefire to crash under some conditions, so it is very good to get rid of it.

jglick: When the exit code is _nonzero_ I meant. Now I see that this is returned from the method, which is…OK, if the caller is either using that, or printing it.

And `redirectErrorStream(true)` is wrong for that use case. You want stderr to go to SF logs, while stdout is returned.

jtnord: when the exit code is non zero the output (both std.out and std.err) is put in the message of the [AssertionError](https://github.com/jtnord/acceptance-test-harness/blob/e6407582268290b3483741b1f9faa755144f5751/src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java#L168-L170). Sure the mix is perhaps not the best but this would be exactly the same as if you ran a git command in a local terminal with no redirection.

If you exit with zero and `git` prints something on std.out and std.err the caller would see something perhaps unexpected, but YAGNI?

jglick: > if you exit with zero and `git` prints something on std.out and std.err the caller would see something perhaps unexpected

Breaking the calling code, yes.

jtnord: but presumably this could be a good thing too, if the warning is unexpected, causing a failing build, dumping it in a log that would never be inspected (the tests pass so shy look at a log that would be truncated anyway in Jenkins) would not help anyone?

jtnord: there appears to only be a single caller that cares about the returned String (`getLastSha1` which is not used in this test suite).
the only reference to that is in [warnings-ng](https://github.com/jenkinsci/warnings-ng-plugin/blob/277a958abc09da9d9835d11792ea9b88cbb2d22a/ui-tests/src/test/java/io/jenkins/plugins/analysis/warnings/GitForensicsUiTest.java#L80-L108)

jglick: If there is a warning printed by `git` but the test passes, why do we care? Warnings get printed by all sorts of things. If a dev happens to be editing the test they might pay attention.

jtnord: 2f7eb743648d3fa1e613f78e9e4c6095b1cc8b49

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java
```diff
@@ -444,4 +458,29 @@ public Path mkdir(String path) {
...
+        private final StringBuilder output;
+
+        StreamReader(InputStream is, StringBuilder output) {
+            this.is = is;
+            this.output = output;
+        }
+
+        @Override
+        public void run() {
+            try (BufferedReader reader = new BufferedReader(new InputStreamReader(is))) {
```
jglick, line 478: Some static analysis tools will warn that this should be split into two statements.

jglick (approved):

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java
```diff
@@ -444,4 +458,29 @@ public Path mkdir(String path) {
...
+                String line;
+                while ((line = reader.readLine()) != null) {
+                    output.append(line);
+                    output.append(System.lineSeparator());
+                }
+                // EOS close the stream from our side
+                is.close();
```
jtnord: why did I add this... 🤦 the bufferedreader is closed, which should close the InputStreamReader which should close the InputStream...


```suggestion
```

src/main/java/org/jenkinsci/test/acceptance/plugins/git/GitRepo.java
```diff
@@ -444,4 +458,27 @@ public Path mkdir(String path) {
...
+        @Override
+        public void run() {
+            try (BufferedReader reader = new BufferedReader(new InputStreamReader(is))) {
+                String line;
+                while ((line = reader.readLine()) != null) {
+                    output.append(line);
+                    output.append(System.lineSeparator());
+                }
+            } catch (IOException e) {
+                throw new UncheckedIOException("failure reading git output/error", e);
```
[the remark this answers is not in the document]
jtnord: ed9a718f2fde1c7d0f76e7122fa73e6efcb3ddf2

PR #2704, opened 2026-05-15
Noticed several runs of GitLabPlug failing due to the sticky elements

jtnord:
The code as was does not make sense to me, we would not want to disable the sticky elements before navigation, only after (for the case that no navigation occured).

However I do not fully understand the comment either, it implies we have not asked for the configure page but ended up here by some redirection? this happens when creating a new job but that was already [explicitly handled](https://github.com/jenkinsci/acceptance-test-harness/blob/677593f07fa67ea804379079be286c995449eb06/src/main/java/org/jenkinsci/test/acceptance/po/JobsMixIn.java#L39-L41) 

~In order to understand that, remove the sticky elements and add a check to ensure we are on the expected page.~

Now that it is understood, the issue is that navigation that ends up being performed via a `click` or other interaction on a `WebElement` does not cause an `afterGet` or `afterTo`. 
Because we want the framework to do these kind of things for us, now handle navigation via a click in `Scroller` (by capturing `afterClick` and cheking for a URL change.

In the future I intend to further clean this up so that it would also work for new elements injected into the dom (e.g. dialogs), but that would be a follow up.

### Testing done

ensure the tests in CI pass.

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/config_file_provider/ConfigFileProvider.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/config_file_provider/ConfigFileProvider.java
@@ -56,9 +56,7 @@ public class ConfigFileProvider extends PageObject {
 
         clickButton("Next");
         String id = getElement(by.name("config.id")).getAttribute("value");
-        // clicking the button leads to the load of a new page but it is not a navigation event from the listeners PoV
-        // so we need to disable the sticky elmements again
-        new Scroller(driver).disableStickyElements();
+
         return newInstance(type, this, id);
     }
 }
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/config_file_provider/ConfigFileProvider.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/config_file_provider/ConfigFileProvider.java
@@ -26,7 +26,6 @@ package org.jenkinsci.test.acceptance.plugins.config_file_provider;
 import org.jenkinsci.test.acceptance.po.Folder;
 import org.jenkinsci.test.acceptance.po.Jenkins;
 import org.jenkinsci.test.acceptance.po.PageObject;
-import org.jenkinsci.test.acceptance.selenium.Scroller;
 import org.openqa.selenium.WebElement;
 
 public class ConfigFileProvider extends PageObject {
--- a/src/main/java/org/jenkinsci/test/acceptance/po/ConfigurablePageObject.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/ConfigurablePageObject.java
@@ -31,7 +31,6 @@ import com.google.inject.Injector;
 import groovy.lang.Closure;
 import java.net.URL;
 import java.util.concurrent.Callable;
-import org.jenkinsci.test.acceptance.selenium.Scroller;
 import org.openqa.selenium.By;
 import org.openqa.selenium.Keys;
 import org.openqa.selenium.WebElement;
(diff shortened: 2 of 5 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/ConfigurablePageObject.java
```diff
@@ -106,13 +105,12 @@ public <T> T configure(final Callable<T> body) {
      * @see #getConfigUrl()
      */
     public void configure() {
-        // Automatic disabling of sticky elements doesn't seem to occur after a redirect,
-        // so force it after the configuration page has loaded
-        new Scroller(driver).disableStickyElements();
```
jtnord, line 111: was introduced in https://github.com/jenkinsci/acceptance-test-harness/pull/762 but that also handles the specific case for JobMixIn where me may redirect.

src/main/java/org/jenkinsci/test/acceptance/plugins/config_file_provider/ConfigFileProvider.java
```diff
@@ -56,9 +55,7 @@ protected WebElement find(String caption) {
 
         clickButton("Next");
         String id = getElement(by.name("config.id")).getAttribute("value");
-        // clicking the button leads to the load of a new page but it is not a navigation event from the listeners PoV
-        // so we need to disable the sticky elmements again
```
jtnord, line 60: now handled by the framework, the only special case should now be dialogs.

src/main/java/org/jenkinsci/test/acceptance/selenium/Scroller.java
```diff
@@ -97,6 +99,23 @@ public void afterTo(Navigation navigation, URL url) {
...
+    }
+
+    @Override
+    public void afterClick(WebElement element) {
+        if (!Objects.equals(preClickURL, driver.getCurrentUrl())) {
+            // we have performed navigation via a click
+            disableStickyElements();
+        }
+        preClickURL = null;
+    }
```
jtnord, line 117: this handles the case where navigation has occurred due to clicking on a button.

src/main/java/org/jenkinsci/test/acceptance/selenium/Scroller.java
```diff
@@ -97,6 +99,23 @@ public void afterTo(Navigation navigation, URL url) {
...
+        WebDriverListener.super.beforeClick(element);
+    }
+
+    @Override
+    public void afterClick(WebElement element) {
+        if (!Objects.equals(preClickURL, driver.getCurrentUrl())) {
+            // we have performed navigation via a click
+            disableStickyElements();
+        }
+        preClickURL = null;
```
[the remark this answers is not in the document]
jtnord: if this throws an exception a stale preClickURL does not matter.  it is always set in the `beforeClick` method.

src/main/java/org/jenkinsci/test/acceptance/po/ConfigurablePageObject.java
```diff
@@ -106,13 +105,12 @@ public <T> T configure(final Callable<T> body) {
...
-
         if (!driver.getCurrentUrl().equals(getConfigUrl().toExternalForm())) {
             visit(getConfigUrl());
         }
+        // ensure that code is using the correct URL so that navigation hooks work correctly.
```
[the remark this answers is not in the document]
jtnord: by code I am refering to the calling code here. The change makes this appear as though it is not user error.

PR #2713, opened 2026-05-19
stop relying on javascript to run at the right time to disable sticky features

jtnord:
Changes how we disable sticky elements so that it is reliable and not subject to timing issues.
If any downstream test/PO previously used `new Scroller(driver).disableStickyElements();` this should now be removed (no replacement is needed).
This does affect the timing of tests, such that the ATH is now quicker once loading.  The speed increase may intern reveal more flaky code where a test /`PageObject`/`PageArea` did not correctly wait for an asynchronous event to occur that manipulated the dom.

Rather than injecting Javascript on every page load, we manipulate the CSS returned by Jenkins to strip the sticky positioning at source.

`Safari` is no longer a supported browser due to the lack of support for webdriver-bidi

### Submitter checklist


src/main/java/org/jenkinsci/test/acceptance/po/JobsMixIn.java
jtnord, line 1: split to #2719 whilst it is needed and discovered here, its purely because we are reacting much quicker on a page load.

src/test/java/plugins/TimestamperPluginTest.java
jtnord, line 1: split to #2720

src/test/java/plugins/NodeLabelParameterPluginTest.java
jtnord, line 1: extracted to #2721

src/test/java/org/jenkinsci/test/acceptance/po/SnippetGeneratorTest.java
jtnord, line 1: extracted to https://github.com/jenkinsci/acceptance-test-harness/pull/2731

src/main/java/org/jenkinsci/test/acceptance/po/SnippetGenerator.java
jtnord, line 1: extracted to https://github.com/jenkinsci/acceptance-test-harness/pull/2731

src/main/java/org/jenkinsci/test/acceptance/po/ArtifactArchiver.java
jtnord, line 1: extracted to https://github.com/jenkinsci/acceptance-test-harness/pull/2731

src/main/java/org/jenkinsci/test/acceptance/selenium/Scroller.java
```diff
@@ -82,20 +97,118 @@ public Scroller(WebDriver driver) {
         } catch (IOException e) {
             throw new Error("Failed to load the JavaScript file", e);
         }
+
+        Script script = new Script(driver);
+        script.addPreloadScript("""
```
jtnord: I don't think this is actually needed..

src/main/java/org/jenkinsci/test/acceptance/po/Job.java
jtnord, line 1: extracted to https://github.com/jenkinsci/acceptance-test-harness/pull/2732

src/test/java/plugins/NodeLabelParameterPluginTest.java
jtnord, line 1: extracted to https://github.com/jenkinsci/acceptance-test-harness/pull/2732

src/main/java/org/jenkinsci/test/acceptance/plugins/junit/TestReport.java
jtnord, line 1: split to #2739

src/main/java/org/jenkinsci/test/acceptance/po/Slave.java
jtnord, line 1: split to #2738

src/main/java/org/jenkinsci/test/acceptance/selenium/StickyElementIntercepter.java
```diff
@@ -0,0 +1,198 @@
...
+                                    .orElse(null);
+                            expires = clientResponse
+                                    .headers()
+                                    .firstValue("Last-Modified")
+                                    .orElse(null);
+                        }
+
+                        String patchedCSS = removeStickyPositioning(css);
+                        // LOGGER.warning(() -> "replaced sticky ***before:***\n" + css + "\n***after:***\n" +
+                        // patchedCSS);
```
jtnord:
```suggestion
```

src/main/java/org/jenkinsci/test/acceptance/selenium/StickyElementIntercepter.java
```diff
@@ -0,0 +1,198 @@
...
+                        if (lastModified != null) {
+                            headers.add(
+                                    new Header("Last-Modified", new BytesValue(BytesValue.Type.STRING, lastModified)));
+                        }
+                        if (expires != null) {
+                            headers.add(new Header("Expires", new BytesValue(BytesValue.Type.STRING, expires)));
+                        }
+                        responseParams.headers(headers);
+                        network.provideResponse(responseParams);
+                        // network.continueResponse(new ContinueResponseParameters(responseId).);
```
jtnord:
```suggestion
```

src/main/java/org/jenkinsci/test/acceptance/selenium/StickyElementIntercepter.java
```diff
@@ -0,0 +1,198 @@
...
+                    } catch (IOException e) {
+                        throw new UncheckedIOException("Could not recieve CSS from Jenkins at URL:" + requestUrl, e);
```
jtnord:
```suggestion
                        throw new UncheckedIOException("Could not retrieve CSS from Jenkins at URL:" + requestUrl, e);
```

src/main/java/org/jenkinsci/test/acceptance/selenium/StickyElementIntercepter.java
```diff
@@ -0,0 +1,198 @@
...
+                    } catch (InterruptedException e) {
+                        throw new UncheckedIOException(
+                                "Could not recieve CSS from Jenkins at URL: " + requestUrl,
```
jtnord:
```suggestion
                                "Could not retrieve CSS from Jenkins at URL: " + requestUrl,
```

src/main/java/org/jenkinsci/test/acceptance/selenium/StickyElementIntercepter.java
```diff
@@ -0,0 +1,198 @@
...
+    public StickyElementIntercepter(WebDriver driver) {
+
+        @SuppressWarnings("resource")
+        Network network = new Network(driver);
+        AddInterceptParameters p = new AddInterceptParameters(InterceptPhase.BEFORE_REQUEST_SENT);
+        String id = network.addIntercept(p);
+        network.onBeforeRequestSent(request -> {
+            final String responseId = request.getRequest().getRequestId();
+            final RequestData requestData = request.getRequest();
+            // if (responseDetails.getIntercepts().contains(id)) {
```
jtnord:
```suggestion
```

src/main/java/org/jenkinsci/test/acceptance/selenium/StickyElementIntercepter.java
```diff
@@ -0,0 +1,198 @@
...
+                        return;
+                    }
+                    try {
+                        network.continueRequest(new ContinueRequestParameters(responseId));
+                    } catch (Exception ignored) {
+                        if (requestUrl.endsWith("/apple-touch-icon.png") || requestUrl.endsWith("/favicon.svg")) {
+                            // these often cause errors so ignore them, it does not impact Jenkins or the browser based
+                            // testing.
+                            return;
+                        }
```
jtnord:
```suggestion
                        if (requestUrl.endsWith("/apple-touch-icon.png")
                                || requestUrl.endsWith("/favicon.svg")
                                || requestUrl.endsWith("/favicon-dark.ico")) {
                            // these often cause errors so ignore them.
                            // it does not impact Jenkins or the browser based testing.
                            return;
                        }
```

src/main/java/org/jenkinsci/test/acceptance/selenium/StickyElementInterceptor.java
```diff
@@ -0,0 +1,196 @@
...
+                            css = clientResponse.body();
+                            lastModified = clientResponse
+                                    .headers()
+                                    .firstValue("Expires")
+                                    .orElse(null);
+                            expires = clientResponse
+                                    .headers()
+                                    .firstValue("Last-Modified")
+                                    .orElse(null);
+                        }
```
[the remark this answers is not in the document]
jtnord: >HttpClient is not AutoCloseable

go back to school we are using Java21 and this does compile.

src/main/java/org/jenkinsci/test/acceptance/selenium/StickyElementIntercepter.java
```diff
@@ -0,0 +1,196 @@
...
+                                    client.send(clientRequest, BodyHandlers.ofString(StandardCharsets.UTF_8));
+                            css = clientResponse.body();
+                            lastModified = clientResponse
+                                    .headers()
+                                    .firstValue("Expires")
+                                    .orElse(null);
+                            expires = clientResponse
+                                    .headers()
+                                    .firstValue("Last-Modified")
+                                    .orElse(null);
```
jtnord:
```suggestion
                            expires = clientResponse
                                    .headers()
                                    .firstValue("Expires")
                                    .orElse(null);
                            lastModified = clientResponse
                                    .headers()
                                    .firstValue("Last-Modified")
                                    .orElse(null);
```

src/main/java/org/jenkinsci/test/acceptance/selenium/StickyElementInterceptor.java
```diff
@@ -0,0 +1,196 @@
...
+                    } catch (InterruptedException e) {
+                        throw new UncheckedIOException(
+                                "Could not retrieve CSS from Jenkins at URL: " + requestUrl,
+                                new IOException("cause", e));
+                    }
+                } else {
+                    if (requestUrl.startsWith("data:")) {
+                        // XXX should not be intercepted
+                        return;
+                    }
```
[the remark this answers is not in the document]
jtnord: no... data URLs are not network requests and should never trigger the intercept, they can not be continued and attempting to do so caused issues.

src/main/java/org/jenkinsci/test/acceptance/selenium/StickyElementInterceptor.java
```diff
@@ -0,0 +1,196 @@
...
+                        }
+                        if (expires != null) {
+                            headers.add(new Header("Expires", new BytesValue(BytesValue.Type.STRING, expires)));
+                        }
+                        responseParams.headers(headers);
+                        network.provideResponse(responseParams);
+                    } catch (IOException e) {
+                        throw new UncheckedIOException("Could not retrieve CSS from Jenkins at URL:" + requestUrl, e);
+                    } catch (InterruptedException e) {
+                        throw new UncheckedIOException(
```
jtnord:
```suggestion
                        Thread.currentThread().interrupt();
                        throw new UncheckedIOException(
```

src/main/java/org/jenkinsci/test/acceptance/selenium/StickyElementInterceptor.java
```diff
@@ -0,0 +1,196 @@
...
+        for (Header h : requestData.getHeaders()) {
+            if ("Cookie".equalsIgnoreCase(h.getName())) {
+                BytesValue v = h.getValue();
+
+                switch (v.getType()) {
+                    case BytesValue.Type.STRING:
+                        return v.getValue();
+                    default:
+                        throw new IllegalStateException("expected a String header for cookie, but got: " + v.getType());
+                }
```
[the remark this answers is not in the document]
jtnord: > will fail to compile if BytesValue.Type is an enum (the usual pattern for Selenium’s BytesValue.Type).

`BytesValue.Type` is an enum this compiles perfectly - go away.

src/main/java/org/jenkinsci/test/acceptance/selenium/StickyElementIntercepter.java
```diff
@@ -0,0 +1,196 @@
...
+ * 	at org.jenkinsci.test.acceptance.po.CapybaraPortingLayer.clickButton(CapybaraPortingLayer.java:66)
+ * 	at org.jenkinsci.test.acceptance.po.JenkinsConfig.addTool(JenkinsConfig.java:35)
+ * 	at org.jenkinsci.test.acceptance.plugins.ant.AntInstallation.install(AntInstallation.java:37)
+ * 	at plugins.AntPluginTest.autoInstallAnt(AntPluginTest.java:76)
+ * 	}
+ * </pre>
+ *
+ * <p>
+ * This work around simply replaces any occurence of {@code position:\\s*sticky} with {@code position: relative}
+ */
```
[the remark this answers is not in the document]
jtnord: this fix is broken!!

src/main/java/org/jenkinsci/test/acceptance/Config.java
```diff
@@ -124,14 +121,13 @@ private WebDriver createWebDriver(TestCleaner cleaner, TestName testName) throws
                         cleaner, "selenium/standalone-chrome:4.35.0", buildChromeOptions(testName));
             case "chrome":
                 return new ChromeDriver(buildChromeOptions(testName));
-            case "safari":
-                return new SafariDriver(buildSafariOptions(testName));
```
jtnord, line 128: safari has no bidi support (according to the interweb its support is the actual browser is sketchy, and in selenium java bindings it is non existing)

src/main/java/org/jenkinsci/test/acceptance/Config.java
```diff
@@ -216,22 +212,6 @@ private ChromeOptions buildChromeOptions(TestName testName) throws IOException {
         return chromeOptions;
     }
 
-    private SafariOptions buildSafariOptions(TestName testName) throws IOException {
-        SafariOptions safariOptions = new SafariOptions();
-        safariOptions.setImplicitWaitTimeout(Duration.ofSeconds(IMPLICIT_WAIT_TIMEOUT));
-        safariOptions.setPageLoadTimeout(Duration.ofSeconds(PAGE_LOAD_TIMEOUT));
-
-        // no bidi support :(
-        // safariOptions.enableBiDi();
```
jtnord, line 225: bidi is now required.

src/main/java/org/jenkinsci/test/acceptance/selenium/StickyElementInterceptor.java
```diff
@@ -0,0 +1,197 @@
...
+                        responseParams.headers(headers);
+                        network.provideResponse(responseParams);
+                    } catch (IOException e) {
+                        throw new UncheckedIOException("Could not retrieve CSS from Jenkins at URL:" + requestUrl, e);
+                    } catch (InterruptedException e) {
+                        Thread.currentThread().interrupt();
+                        throw new UncheckedIOException(
+                                "Could not retrieve CSS from Jenkins at URL: " + requestUrl,
+                                new IOException("cause", e));
+                    }
```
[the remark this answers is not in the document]
jtnord: no we want to fail the test, find the issue and fix it

src/main/java/org/jenkinsci/test/acceptance/selenium/StickyElementInterceptor.java
```diff
@@ -0,0 +1,197 @@
...
+ * </pre>
+ *
+ * <p>
+ * This work around simply replaces any occurence of {@code position:\\s*sticky} with {@code position: relative}
+ */
+public class StickyElementInterceptor {
+
+    private final Logger LOGGER = Logger.getLogger(StickyElementInterceptor.class.getName());
+
+    public StickyElementInterceptor(WebDriver driver) {
```
timja: Did you consider just adding a system property to Jenkins and then modifying the CSS based on a value added to e.g. head to remove the stickiness of these elements without all this hackery?

jtnord: > Did you consider just adding a system property to Jenkins and then modifying the CSS based on a value added to e.g. head to remove the stickiness of these elements without all this hackery?

Yes, but I could not think of a sustainable way that would reliably work, to the point I feel it is a lot more hackery than this solution.

We would need to inject the system property into a running server like we do the update center (as the ATH can be used to point to an existing Jenkins install) so would have to be something set like the mock update center at the start of every test run. But that itself can have some issues as there are sticky elements on the script page that we need to interact with, so they could interfere with the interaction to disable them!

We would also still need to build 2 sets of CSS files in Jenkins, and direct the jelly to load the correct one depending on the flags.
Additionally it would impact anyone using adjuncts in all plugins (they would all need to be traced down and have 2 versions of any CSS, one with sticky and one without).

Not providing 2 CSS files but performing the manipulation in Javascript is just flaky as when the CSS is loaded the JS would run **sometime** afterwards. The dom may have already been complete as depending on where the CSS is loaded it comes from can be loaded async at arbitrary time after the main page.
There is already production issues in JS async loading due to the CSP work (e.g. the Behavior not being applied in time for some elements leading to "this requires a POST" type issues), so manipulating in JS would seem to be a path towards instability, the exact thing I am trying to get rid of.

My first instinct was actually to make the CSS something like

```css
myclass.position: ${stickyEnabled()?"sticky":relative"}
```
or
```css
myclass.position: ${stickyClass}
```

But claude and internet crawling to me you can not do that, and there is still the issue with any plugin using sticky positioing that would need to be adapted. (we would still have need to have a function to set that flag too but I guess we could have introduced a new API `${JENKINS_URL}/disableStickyClasses` )!

src/main/java/org/jenkinsci/test/acceptance/Config.java
```diff
@@ -371,20 +370,18 @@ public WebDriver createWebDriver(TestCleaner cleaner, TestName testName, Elastic
...
-        Dimension oldSize = base.manage().window().getSize();
+        Dimension oldSize = driver.manage().window().getSize();
         if (oldSize.height < 1090 || oldSize.width < 1680) {
-            base.manage().window().setSize(new Dimension(1680, 1090));
+            driver.manage().window().setSize(new Dimension(1680, 1090));
         }
-        Scroller scroller = new Scroller(base);
-        final EventFiringDecorator<WebDriver> decorator = new EventFiringDecorator<>(scroller);
-        WebDriver d = decorator.decorate(base);
+        StickyElementIntercepter scroller = new StickyElementIntercepter(driver);
```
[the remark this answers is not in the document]
jtnord: all the browsers we now support have bidi enabled.

src/main/java/org/jenkinsci/test/acceptance/selenium/StickyElementInterceptor.java
```diff
@@ -0,0 +1,212 @@
...
+                        .thenAccept(clientResponse -> providePatchedCss(requestId, requestUrl, clientResponse))
+                        .exceptionally(t -> {
+                            // The request stays blocked and will fail with a page-load timeout. A failure here
+                            // likely indicates a bug either in this interceptor or in Jenkins, so surface it.
+                            LOGGER.log(
+                                    Level.WARNING,
+                                    t,
+                                    () -> "Could not retrieve CSS from Jenkins at URL: " + requestUrl);
+                            return null;
+                        });
```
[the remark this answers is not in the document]
jtnord: this is fine.

src/main/java/org/jenkinsci/test/acceptance/selenium/StickyElementInterceptor.java
```diff
@@ -0,0 +1,212 @@
...
+                if (requestUrl.startsWith("data:")) {
+                    // XXX should not be intercepted
+                    return;
+                }
+                try {
+                    network.continueRequest(new ContinueRequestParameters(requestId));
+                } catch (Exception ignored) {
+                    if (requestUrl.endsWith("/apple-touch-icon.png")
+                            || requestUrl.endsWith("/favicon.svg")
+                            || requestUrl.endsWith("/favicon-dark.ico")) {
```
jtnord, line 127: these are known to have no side effects.

src/main/java/org/jenkinsci/test/acceptance/selenium/StickyElementInterceptor.java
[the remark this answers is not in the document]
jtnord: in fact this is better than fine.  by blocking the request this found some issues in proprietary plugins as the page load timed out due to referencing a non existing stylesheet.

src/main/java/org/jenkinsci/test/acceptance/selenium/StickyElementInterceptor.java
timja: We don't need two sets of CSS files.

I pointed claude at my comment and it implemented this which I've tested and it works fine:
https://github.com/jenkinsci/jenkins/pull/26876

If script console turns out to be an issue we can just add an API to toggle this or something, pretty straightforward and a lot simpler IMO.

jtnord: Bloody AI.  I asked claude before I started and it assured me you could not modify CSS rules like this...

PR #2716, opened 2026-05-20
Add acceptance test coverage for parameterized pipelines

### LINKED ISSUE #2690: Add test coverage for parameterised pipelines
### What feature do you want to see added?

See issue: https://github.com/jenkinsci/jenkins/issues/26683

Would've expected this to have been caught

### Upstream changes

_No response_

### Are you interested in contributing this feature?

_No response_

Dipanshusinghh:
Hi,

I've added the test coverage for parameterized pipelines to cover the gap mentioned in #2690.

Fixes #2690

Earlier in Jenkins 2.560, triggering a pipeline with parameters threw a ClassCastException (JENKINS-26683), but we didn't had any test in ATH to catch it.

I added a new testcase `testParameterizedPipeline` in `PipelineTest.java`. It defines a string parameter on job, saves the configuration and then runs build with custom parameter value to check if it succeeds and prints correct value in console output.

Tested it locally and it works fine. Let me know if any change is needed.

Thanks!

### DIFF
--- a/src/test/java/plugins/PipelineTest.java
+++ b/src/test/java/plugins/PipelineTest.java
@@ -66,9 +66,8 @@ public class PipelineTest extends AbstractPipelineTest {
 
     @Test
     public void testParameterizedPipeline() {
-        final WorkflowJob job = createPipelineJobWithScript("node {\n"
-                + "  echo \"param is ${params.MY_PARAM}\"\n"
-                + "}");
+        final WorkflowJob job =
+                createPipelineJobWithScript("node {\n" + "  echo \"param is ${params.MY_PARAM}\"\n" + "}");
 
         job.configure();
         job.addParameter(StringParameter.class).setName("MY_PARAM").setDefault("default-val");
--- a/src/test/java/plugins/PipelineTest.java
+++ b/src/test/java/plugins/PipelineTest.java
@@ -47,7 +47,6 @@ import org.junit.Test;
  * 4- Assert results have been collected
  *
  */
-@WithPlugins({"git", "junit", "javadoc"})
 public class PipelineTest extends AbstractPipelineTest {
 
     @Before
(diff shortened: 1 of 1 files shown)


jtnord (changes requested):
formatting errors -> https://github.com/jenkinsci/acceptance-test-harness/pull/2716/checks?check_run_id=76942555165

please take a look

jglick:
https://github.com/jenkinsci/plugin-pom/blob/master/README.md#format-as-you-build

Dipanshusinghh:
Thnx ,
for the guidance, @jtnord and @jglick. I really appreciate you pointing out the formatting issues & sharing the reference link. I have cleaned up the imports, formatted the code properly, &pushed the updates to the branch  .It should be good to go now!

Dipanshusinghh:
Hi @jtnord, I have fixed the formatting issues using Spotless. all CI tests & build checks are passing now. please take another look & review

src/test/java/plugins/PipelineTest.java
```diff
@@ -60,6 +64,19 @@ public void testCompletePipeline() {
         this.assertJavadoc(job);
     }
 
+    @Test
+    public void testParameterizedPipeline() {
+        final WorkflowJob job =
+                createPipelineJobWithScript("node {\n" + "  echo \"param is ${params.MY_PARAM}\"\n" + "}");
```
jtnord: 1) its preferable to use a multiline string these days.
2) you don't need a `node` to run `echo`, this can be a single `echo` statement that will run much quicker.

This test does not need `git` `junit` or `javadoc`, only the standard workflow plugins.  as such its best to not install them.  Change the default installation of plugins for this test to just be the what is needed for the basis of pipeline and add specific `WithPlugins` for tests that need specific test.  The base probably already has what is needed -> https://github.com/Dipanshusinghh/acceptance-test-harness/blob/9bba9612db20f8a68c521621225010a9b226d27b/src/main/java/org/jenkinsci/test/acceptance/AbstractPipelineTest.java#L38C1-L39C1

Dipanshusinghh: Makes sense, thanks @jtnord. I've updated the script to a text block, removed the node wrapper, and moved @WithPlugins to the method level. Pushed!

jtnord (approved):

jtnord:
Thankyou @Dipanshusinghh

PR #2731, opened 2026-05-28
Deflake SnippetGeneratorTest

jtnord:
loading of the describables HTML is asynchronous to the page load, so sometimes clicking on "generate snippet" would occur before the describables FORM had been injected into the page, which would mean the form submitted would be empty of all fields, and then the test would fail because

```
Expected: is "archiveArtifacts artifacts: '', followSymlinks: false"
     but: was "java.lang.NullPointerException: Cannot invoke \"String.trim()\" because \"artifacts\" is null
        at hudson.tasks.ArtifactArchiver.<init>(ArtifactArchiver.java:122)
        at java.base/jdk.internal.reflect.DirectConstructorHandleAccessor.newInstance(DirectConstructorHandleAccessor.java:62)
        at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
        at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:483)
        at org.kohsuke.stapler.RequestImpl.invokeConstructor(RequestImpl.java:686)
        at org.kohsuke.stapler.RequestImpl.instantiate(RequestImpl.java:1015)
        at org.kohsuke.stapler.RequestImpl$TypePair.convertJSON(RequestImpl.java:866)
        at org.kohsuke.stapler.RequestImpl.bindJSON(RequestImpl.java:633)
        at org.kohsuke.stapler.RequestImpl.bindJSON(RequestImpl.java:628)
        at hudson.tasks.ArtifactArchiver$DescriptorImpl.newInstance(ArtifactArchiver.java:371)
        at hudson.tasks.ArtifactArchiver$DescriptorImpl.newInstance(ArtifactArchiver.java:340)
        at PluginClassLoader for workflow-cps//org.jenkinsci.plugins.workflow.cps.Snippetizer.doGenerateSnippet(Snippetizer.java:523)
        ..."
```

We now obtain the Step PageArea and forcibly interact with it.  This inturn calls `resolve()` on the control which waits until it is visible.

### Testing done

extracted from #2713 where the test was reliably observed to fail (both locally and In CI).
rann 10 times with the fix and observed no failures.

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/po/ArtifactArchiver.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/ArtifactArchiver.java
@@ -3,7 +3,7 @@ package org.jenkinsci.test.acceptance.po;
 /**
  * @author Kohsuke Kawaguchi
  */
-@Describable("Archive the artifacts")
+@Describable({"Archive the artifacts", "archiveArtifacts: Archive the artifacts"})
 public class ArtifactArchiver extends AbstractStep implements PostBuildStep {
     public ArtifactArchiver(Job parent, String path) {
         super(parent, path);
--- a/src/main/java/org/jenkinsci/test/acceptance/po/SnippetGenerator.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/po/SnippetGenerator.java
@@ -25,6 +25,11 @@ public class SnippetGenerator extends PageObject {
         return (WorkflowJob) super.getContext();
     }
 
+    public <T extends Step> T selectStep(Class<T> stepDescriptor) {
+        control("/").select(stepDescriptor);
+        return newInstance(stepDescriptor, getContext(), "/prototype");
+    }
+
     /**
      * Generates the sample pipeline script.
      *
--- a/src/test/java/org/jenkinsci/test/acceptance/po/SnippetGeneratorTest.java
+++ b/src/test/java/org/jenkinsci/test/acceptance/po/SnippetGeneratorTest.java
@@ -17,7 +17,7 @@ public class SnippetGeneratorTest extends AbstractJUnitTest {
         SnippetGenerator snippetGenerator = new SnippetGenerator(job);
         snippetGenerator.open();
         ArtifactArchiver aa = snippetGenerator.selectStep(ArtifactArchiver.class);
-        aa.includes(null);
-        assertThat(snippetGenerator.generateScript(), is("archiveArtifacts artifacts: '', followSymlinks: false"));
+        aa.includes("*.bin");
+assertThat(snippetGenerator.generateScript(), is("archiveArtifacts artifacts: '*.bin', followSymlinks: false"));
     }
 }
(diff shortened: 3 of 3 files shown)


src/main/java/org/jenkinsci/test/acceptance/po/SnippetGenerator.java
```diff
@@ -25,6 +25,11 @@ protected WorkflowJob getContext() {
         return (WorkflowJob) super.getContext();
     }
 
+    public <T extends Step> T selectStep(Class<T> stepDescriptor) {
+        control("/").select(stepDescriptor);
+        return newInstance(stepDescriptor, getContext(), "/prototype");
+    }
```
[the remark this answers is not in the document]
jtnord: this is fine as all the options are added statically in the original HTML

src/main/java/org/jenkinsci/test/acceptance/po/ArtifactArchiver.java
```diff
@@ -3,7 +3,7 @@
 /**
  * @author Kohsuke Kawaguchi
  */
-@Describable("Archive the artifacts")
+@Describable({"Archive the artifacts", "archiveArtifacts: Archive the artifacts"})
```
jglick, line 6: Should we not be specifying the impl class FQN for maintainability?

src/test/java/org/jenkinsci/test/acceptance/po/SnippetGeneratorTest.java
```diff
@@ -16,7 +16,8 @@ public void createPipelineSnippetToArchiveArtifacts() {
 
         SnippetGenerator snippetGenerator = new SnippetGenerator(job);
         snippetGenerator.open();
-
+        ArtifactArchiver aa = snippetGenerator.selectStep(ArtifactArchiver.class);
+        aa.includes(null);
         assertThat(snippetGenerator.generateScript(), is("archiveArtifacts artifacts: '', followSymlinks: false"));
```
jglick:
```suggestion
        aa.includes("*.bin");
assertThat(snippetGenerator.generateScript(), is("archiveArtifacts artifacts: '*.bin', followSymlinks: false"));
```
would be more natural

jglick (approved):

src/test/java/org/jenkinsci/test/acceptance/po/SnippetGeneratorTest.java
jtnord: 👍 I was not sure if testing the empty string was there for a reason.

src/main/java/org/jenkinsci/test/acceptance/po/ArtifactArchiver.java
jtnord: I was initially surprised, as It's not available in the select options in the snippet generator if that's what you mean (the option name is the same as it's text which is this human readable string).

At the same time we are supposed to be driving the same way a user would drive the browser, so using internal FQNs over names seems to go against that, and the names do not change that regularly and are easy to adapt if they do.

The difference here is that you get a different display in Freestyle than in the snippet generator (which prepends the step name followed by colon), I could change selectStep to do some ends with matching but I figured it's not worth the effort given nothing else appears to use the snippet generator.

src/test/java/org/jenkinsci/test/acceptance/po/SnippetGeneratorTest.java
jglick: I think it was an artifact of the fact that the test was implicitly assuming this entry was the first in the list, and not actually configuring it directly.

jglick (approved):

PR #2739, opened 2026-06-02
better diagnostics for TestRport.assertFailureContent

jtnord:
Observed in https://github.com/jenkinsci/acceptance-test-harness/runs/79077204013 that the expected tests output was not there, but the screenshot looked right, so add some better diagnostics.

Rather than saying we did not find X, say also what we did find.

### Testing done

ran the sporadically failing test locally, it passed before the change and after the change.  Does not proove this code is good but its not breaking the happy case.

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/test/acceptance/plugins/junit/TestReport.java
+++ b/src/main/java/org/jenkinsci/test/acceptance/plugins/junit/TestReport.java
@@ -1,9 +1,12 @@
 package org.jenkinsci.test.acceptance.plugins.junit;
 
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.empty;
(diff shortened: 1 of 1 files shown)


MarkEWaite (approved):

src/main/java/org/jenkinsci/test/acceptance/plugins/junit/TestReport.java
```diff
@@ -48,23 +51,28 @@ public void assertFailureContent(String test, String content) {
...
+                return;
             }
+            // not our needle, save for later diagnostics
+            contents.add(c);
         }
-        assertThat("No test found with name " + test, rows, not(empty()));
-        assertThat("No test found with given content", found);
+        // if we got here we know that we have failed.
+        // but still assert so that we have a better idea of what we did find
+        assertThat("No test found with given content", contents, contains(containsString(content)));
```
jtnord:
```suggestion
        assertThat("No test found with given content", contents, hasItem(containsString(content)));
```

PR #2752, opened 2026-06-10
Save job before copying

daniel-beck:
The behavior change in https://github.com/jenkinsci/jenkins/commit/037c2c30cd26b926ec9df3d1b60e16b80608edb4 seems just enough to make this test fall over (`<actions/>` is [basically initialized randomly](https://github.com/jenkinsci/jenkins/blob/ad6da6a27530aae33b48369fb559d38d4b8a86bc/core/src/main/java/hudson/model/Actionable.java#L68-L98)), but in my testing adding this explicit call makes it pass again.

### Testing done

Locally executed the test

### Submitter checklist


timja (approved):

daniel-beck:
FTR another option may be to normalize the XML for comparison (i.e., strip an empty `<actions/>`), which would not introduce a behavior change during the test, but rather relax the comparison.

PR #2759, opened 2026-06-15
[JENKINS-76483] Config File Provider plugin UI changes

alecharp:
This is due to https://github.com/jenkinsci/config-file-provider-plugin/pull/436.

The table listing the existing configuration files has been re-worked and the UI test for it was not able to "survive" the changes without changing itself.

### Testing done

Running `BROWSER=firefox JENKINS_VERSION=2.555.1 mvn test -Dtest=ConfigFileProviderTest` locally to validate the changes.

### Submitter checklist

### DIFF
--- a/src/test/java/plugins/ConfigFileProviderTest.java
+++ b/src/test/java/plugins/ConfigFileProviderTest.java
@@ -174,7 +174,7 @@ public class ConfigFileProviderTest extends AbstractJUnitTest {
         // We want to delete the config file and re-run the job to see it fail
         jenkins.visit("configfiles");
         // this won't age well
-        runThenHandleDialog(() -> driver.findElement(by.xpath("//td[.='%s']/parent::tr/td[2]/a[1]", mvnConfig.id()))
+        runThenHandleDialog(() -> driver.findElement(by.xpath("//td[.='%s']/parent::tr/td[6]/a[1]", mvnConfig.id()))
                 .click());
 
         jobLog = this.buildJobAndGetConsole(job, false);


alecharp:
<s>At this point, I still have a problem with `MavenSettingConfig#replaceAll`.</s>

Re-running the tests locally and now they pass, without any code change.

src/test/java/plugins/ConfigFileProviderTest.java
```diff
@@ -174,7 +174,7 @@ public void testConfigFileProviderWithFolders() {
         // We want to delete the config file and re-run the job to see it fail
         jenkins.visit("configfiles");
         // this won't age well
-        runThenHandleDialog(() -> driver.findElement(by.xpath("//td[.='%s']/parent::tr/td[2]/a[1]", mvnConfig.id()))
+        runThenHandleDialog(() -> driver.findElement(by.xpath("//td[.='%s']/parent::tr/td[6]/a[1]", mvnConfig.id()))
```
jglick, line 177: Seems fragile. Could the selector somehow find the content regardless of which column it happens to be in?

> this won't age well

Indeed.

alecharp: I agree, fix is far from great, but it's easy. We could introduce testid in the plugin so that we could find them easily here, but that means another release of the plugin (not a problem exactly, but it's more time that's all).

timja: release should be straightforward?

timja (approved):
Shipping to unblock this repo, would be good to remove the fragility in the test

alecharp:
>would be good to remove the fragility in the test

https://issues.jenkins.io/browse/JENKINS-76484

src/test/java/plugins/ConfigFileProviderTest.java
jglick: > We could introduce testid in the plugin

That would be good, though I meant something in the selector that would work without any further plugin changes.

jglick: Like #2760 in fact.

PR #2760, opened 2026-06-16
[JENKINS-76483] Fix delete button selector in ConfigFileProviderTest

aneveux:
:wave: Hello team!

This addresses [JENKINS-76483](https://issues.jenkins.io/browse/JENKINS-76483), caused by https://github.com/jenkinsci/config-file-provider-plugin/pull/436, very sorry about it :bow: 

## Summary
- The `testConfigFileProviderWithFolders` test was failing to locate the delete button on the config files page
- Switched the XPath selector from whatever attribute was previously used to `data-url` containing `removeConfig`
- This aligns with a UI change in the Config File Provider plugin that was already addressed in the page object layer

Thanks for your review :heart:

### DIFF
--- a/src/test/java/plugins/ConfigFileProviderTest.java
+++ b/src/test/java/plugins/ConfigFileProviderTest.java
@@ -173,8 +173,8 @@ public class ConfigFileProviderTest extends AbstractJUnitTest {
 
         // We want to delete the config file and re-run the job to see it fail
         jenkins.visit("configfiles");
-        // this won't age well
-        runThenHandleDialog(() -> driver.findElement(by.xpath("//td[.='%s']/parent::tr/td[6]/a[1]", mvnConfig.id()))
+        runThenHandleDialog(() -> driver.findElement(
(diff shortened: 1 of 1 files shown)


aneveux:
Sorry I just realized @alecharp was quicker and fixed in https://github.com/jenkinsci/acceptance-test-harness/pull/2759.
Still keeping this PR open as it might be another alternative for a less fragile approach?

LMK what you think.

jglick (approved):
🤷 if it works

alecharp:
I'm looking into introducing `data-testid` in the plugin to have a more permanent solution to the problem.

timja (approved):

alecharp:
If that works for everyone, let's just use https://issues.jenkins.io/browse/JENKINS-76484 then and merge this.

jtnord:
> I'm looking into introducing `data-testid` in the plugin to have a more permanent solution to the problem.

or check the title which is human readable.

jtnord (approved):
better than before

